PR #153, opened 2026-02-01
Fix terminated signal including truncated (#150)

### LINKED ISSUE #150: BUG: terminated signal includes truncated, breaking value bootstrapping in all sequential envs
## Summary

In `SequentialTradingEnv._step`, the `terminated` signal is set to `self.truncated or done`, which means episodes that end because the data ran out (truncation) are incorrectly marked as terminated. This breaks value bootstrapping in every RL algorithm (PPO, SAC, DQN, IQL, etc.) trained on these environments.

## Root Cause

**File**: `torchtrade/envs/offline/sequential.py:494-497`

```python
next_tensordict.set("done", self.truncated or done)
next_tensordict.set("truncated", self.truncated)
next_tensordict.set("terminated", self.truncated or done)
```

Per TorchRL conventions:
- `done = terminated OR truncated`
- `terminated` = episode ended due to **environment dynamics** (e.g., bankruptcy/liquidation)
- `truncated` = episode ended due to **time limit** (e.g., ran out of data)

When `self.truncated=True` and `done=False` (data ran out but agent is not bankrupt):
- **Current behavior**: `terminated=True` ❌
- **Correct behavior**: `terminated=False` ✅

## Impact on RL Training

When an episode is **truncated** (not truly terminal), RL algorithms should bootstrap the value of the final state:

```
V(s_T) ≈ r_T + γ * V(s_{T+1})   # truncated: bootstrap future value
V(s_T) ≈ r_T                      # terminated: no future value
```

By marking truncated episodes as terminated, the agent learns that the end-of-data state has **zero future value**. This causes:

1. **Systematic undervaluation** of states near the end of episodes — the agent thinks the world ends, so it discounts future rewards
2. **Horizon-dependent policies** — the agent behaves differently as it approaches `max_steps`, optimizing for "end of episode" rather than continuous trading
3. **Biased value function** — the error propagates backward through TD learning, affecting value estimates for all states

This affects **every sequential env** (`SeqLongOnlyEnv`, `SeqFuturesEnv`, `SeqLongOnlySLTPEnv`, `SeqFuturesSLTPEnv`) and **every RL algorithm** trained on them.

## Suggested Fix

```python
# BEFORE (wrong)
next_tensordict.set("done", self.truncated or done)
next_tensordict.set("truncated", self.truncated)
next_tensordict.set("terminated", self.truncated or done)

# AFTER (correct)
next_tensordict.set("terminated", done)
next_tensordict.set("truncated", self.truncated)
next_tensordict.set("done", done or self.truncated)
```

Where `done` is the result of `_check_termination()` (bankruptcy/liquidation only).

## Affected Files

- `torchtrade/envs/offline/sequential.py` — `_step` method (~line 494)
- Check if `sequential_sltp.py` has the same pattern
- Check if `onestep.py` has the same pattern

## Verification

After fix, confirm:
- Episode ending from data exhaustion: `terminated=False, truncated=True, done=True`
- Episode ending from bankruptcy/liquidation: `terminated=True, truncated=False, done=True`
- Normal step: `terminated=False, truncated=False, done=False`

BY571:
## Summary

- Separate `terminated` (bankruptcy/liquidation) from `truncated` (data exhaustion/max steps) in all offline environments
- Extract `_check_truncation()` method from `_check_termination()` so max_steps is correctly treated as truncation, not termination
- Fix `OneStepTradingEnv` to set `terminated = not self.truncated` instead of always `True`
- Add regression tests across all three env types, parametrized over spot/futures modes

Fixes #150

## Changes

**`sequential.py`**: Split `_check_termination` into `_check_termination` (bankruptcy only) and `_check_truncation` (time/data limit). Update `_step` to use both.

**`sequential_sltp.py`**: Same `_step` fix (inherits the new methods from parent).

**`onestep.py`**: Set `terminated = not self.truncated` — when the internal rollout ends due to data exhaustion, the position didn't reach a natural conclusion.

## Signal semantics after fix

| Scenario | terminated | truncated | done |
|----------|-----------|-----------|------|
| Normal step | False | False | False |
| Bankruptcy/liquidation | True | False | True |
| Data exhaustion | False | True | True |
| Max steps reached | False | True | True |

## Test plan

- [x] `test_truncation_does_not_set_terminated` — parametrized spot/futures for `SequentialTradingEnv` and `SequentialTradingEnvSLTP`
- [x] `test_bankruptcy_sets_terminated_not_truncated` — verifies inverse case (bankruptcy = terminated, not truncated)
- [x] `test_normal_step_signals` — verifies non-terminal steps have all flags False
- [x] `test_non_truncated_step_sets_terminated` — verifies OneStep env sets terminated correctly
- [x] Full offline test suite: 428 passed
- [x] Full test suite: 938 passed (5 pre-existing example failures unrelated)

### DIFF
--- a/tests/envs/offline/test_onestep.py
+++ b/tests/envs/offline/test_onestep.py
@@ -421,6 +421,21 @@ class TestOneStepRegression:
             next_td = onestep_env.step(action_td)
             assert next_td["next"]["done"].item()
 
+    def test_non_truncated_step_sets_terminated(self, onestep_env):
+        """Non-truncated one-step episode should set terminated=True, truncated=False.
+
+        Regression test for #150: one-step envs must distinguish terminated
+        (SL/TP trigger, hold completed) from truncated (data exhaustion).
+        """
+        td = onestep_env.reset()
+        action_td = td.clone()
+        action_td["action"] = torch.tensor(0)  # Hold - no rollout, no truncation
+        next_td = onestep_env.step(action_td)
+
+        assert next_td["next"]["done"].item() is True
+        assert next_td["next"]["terminated"].item() is True
+        assert next_td["next"]["truncated"].item() is False
+
     def test_action_spec_immutable(self, onestep_env):
         """Action spec should not change during episode."""
         initial_spec = onestep_env.action_spec
--- a/tests/envs/offline/test_sequential.py
+++ b/tests/envs/offline/test_sequential.py
@@ -570,3 +570,91 @@ class TestSequentialEnvRegression:
                 assert value.shape == expected_shape or value.numel() == expected_shape[0]
             if expected_dtype:
                 assert value.dtype == expected_dtype
+
+    @pytest.mark.parametrize("leverage,action_levels,hold_action", [
+        (1, [0, 1], 0),           # Spot
+        (10, [-1, 0, 1], 1),      # Futures
+    ], ids=["spot", "futures"])
+    def test_truncation_does_not_set_terminated(self, sample_ohlcv_df, leverage, action_levels, hold_action):
+        """Truncated episodes (data exhaustion/max steps) must NOT set terminated=True.
+
+        Regression test for #150: terminated included truncated, breaking
+        value bootstrapping in all sequential envs.
+        """
+        config = SequentialTradingEnvConfig(
+            leverage=leverage,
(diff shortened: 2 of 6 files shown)


BY571:
@claude review

PR #186, opened 2026-02-07
Optimize SequentialTradingEnv _step() for ~2x FPS improvement

BY571:
## Summary

- Pre-allocate reusable tensor buffers in `__init__` and write in-place with `.clone()` to eliminate ~40k `torch.tensor()` calls per 5k steps
- Replace `get_base_features()` (uses `searchsorted`) with `get_sequential_observation_with_ohlcv()` (direct index lookup) in `_get_observation()`
- Cache `allows_short` property to avoid repeated linear scan (~4x per step)
- Use `math.log` instead of `np.log` for single-float reward computation (~3-5x faster for scalars)

All changes are internal — no API changes, specs, or outputs affected. Benchmark: ~9,500 FPS vs ~4,000 FPS baseline (~2.4x improvement).

## Test plan

- [x] All 1008 existing tests pass (0 failures)
- [x] `check_env_specs` passes for both spot and futures modes
- [x] Benchmark confirms ~2x+ FPS improvement
- [x] Review agents confirmed buffer aliasing safety via `.clone()`

### DIFF
--- a/torchtrade/envs/offline/sequential.py
+++ b/torchtrade/envs/offline/sequential.py
@@ -226,7 +226,7 @@ class SequentialTradingEnv(TorchTradeOfflineEnv):
         self._allows_short = any(a < 0 for a in self.action_levels) and self.leverage > 1
 
         self._account_state_buf = torch.zeros(6, dtype=torch.float)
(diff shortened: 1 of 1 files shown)


BY571:
@claude review

PR #222, opened 2026-04-27
Add Polymarket prediction market env (PolymarketBetEnv)

BY571:
## Summary

- Add **`PolymarketBetEnv`** — a rolling one-shot betting env for short-cadence Polymarket binary markets (BTC/ETH/SOL up-or-down at 5m / 15m / 1h / 4h / daily cadences). Pattern B / contextual-bandit shape: each step is an independent bet that resolves before the next decision, so there is no carried position and the observation deliberately omits any \`account_state\`. Specs follow TorchRL conventions (\`Binary\` for boolean done flags, reward inside a \`Composite\`, \`done\` auto-derived from \`terminated|truncated\`); \`check_env_specs\` passes.
- Extend **\`MarketScanner\`** with \`slug_prefix\` (case-sensitive structural matching — the env's stable series identifier), \`max_time_to_resolution_minutes\` upper bound, and a new \`next_active_market(prefix)\` helper that uses Gamma's \`end_date_min\` filter so short-cadence markets surface (they have ~$0 24h volume and never make the volume-sorted top page). \`scan()\` automatically switches to chronological sort when targeting upcoming markets.
- Add **broker examples and docs**: \`examples/broker/polymarket/\` with \`scan_markets.py\` (CLI: \`--keyword\` fuzzy / \`--slug-prefix\` exact / \`--max-resolution-minutes\` / \`--max\`), \`run_dry_run.py\` end-to-end demo (uses public \`env.step()\`), and a README walkthrough that follows the **discover → identify → configure → run** flow. \`docs/environments/online.md\` Polymarket section rewritten; root README updated.
- **Workflow**: run \`scan_markets.py\` once to identify the slug prefix you want (\`btc-updown-5m-\`, \`bitcoin-up-or-down-\`, etc.), paste it into \`PolymarketBetEnvConfig\`, swap markets by changing one string. \`dry_run=True\` works without \`py-clob-client\` installed.
- Reviewed across 4 rounds with \`torchrl-engineer\` + \`code-simplifier\` + \`pr-test-analyzer\` per CLAUDE.md; final round caught and fixed a runtime blocker in the example (\`env._step()\` + \`td[\"done\"]\` after spec change).

## Test plan

- [x] Polymarket suite: \`pytest tests/envs/polymarket/ -v\` — 99 passing
- [x] Full repo suite: \`pytest tests/ --ignore=tests/actor\` — 1647 passing, 15 skipped
- [x] Spec round-trip: \`check_env_specs(PolymarketBetEnv(...))\` passes
- [x] Live discovery: \`scan_markets.py --slug-prefix btc-updown-5m-\` returns active 5m markets from Gamma
- [x] Live dry-run example: \`run_dry_run.py\` constructs env, scans for next market, completes one bet end-to-end (without \`py-clob-client\` installed)
- [x] No stale references to old \`PolyTimeBarEnv\` / \`PolymarketObservationClass\` in source, tests, docs, or examples
- [x] No temp planning/review markdown files staged


BY571:
@CharlieHelps review

BY571:
@charliecreates both items addressed in `f1c9f85`:

**1) Dry-run no longer initializes CLOB client / API creds.** `order_executor.py` now short-circuits to `self.client = None` for any `dry_run=True` regardless of `py-clob-client` availability — paper trading is fully offline. Two new tests pin this: `test_dry_run_does_not_construct_clob_client` (asserts the `ClobClient` constructor is never called) and `test_dry_run_does_not_derive_api_creds` (asserts `create_or_derive_api_creds` + `set_api_creds` are never invoked).

**2) Order failures no longer book phantom P&L.** `env._step` now captures the `trader.buy()` result; on `success=False` it logs the failure and sets effective `stake = 0` so `_compute_payoff` returns `0.0` (no win, no loss, no cash mutation). New test `test_failed_order_in_live_mode_books_zero_payoff` forces `trader.buy` to return `{"success": False, "error": "insufficient USDC"}` and asserts reward is 0 and cash is unchanged.

Tests: 102 polymarket / 1650 full repo, `check_env_specs` passes.

BY571:
@charliecreates thanks for verifying. FYI there's one follow-up commit since (`781a8ae`) — strengthens `test_failed_order_in_live_mode_books_zero_payoff` to also pin `_step_count`, `terminated`, `truncated`, the next-market fetch, and that the next observation is the real next-market state rather than the terminal zero-tensor. Parametrized over both possible underlying outcomes too. Plus a small consistency nit: `cancel_all` now uses `self._dry_run` (matching `buy()`) instead of `self.client is None`.

Test count is 104 now (was 102 when you re-ran). No behavior change vs `f1c9f85` for either of the original findings.

BY571:
@CharlieHelps review recent changes

BY571:
@charliecreates fixed in `fc94abd`.

`BinanceOHLCVTransform._attach_observations` no longer uses `continue` on missing keys — every declared `(timeframe, window)` is unconditionally written. When the observer omits a key, the transform now fills with `torch.zeros((window, n_features))` of the declared shape and logs a warning. Spec contract is honored at runtime; downstream collectors that trust the spec won't hit missing-key failures.

The test that previously pinned the wrong behavior (`test_missing_source_key_is_skipped_silently`) was renamed to `test_missing_source_key_filled_with_zeros_to_honor_spec` and now asserts:
- both declared keys are present in the runtime TensorDict,
- the provided key carries the real observer value,
- the missing key is zero-filled at the declared shape,
- the warning surfaces (silent fallback would be the wrong choice).

`test_observer_exception_propagates` is unchanged — explicit observer exceptions still raise loudly. Only the partial-payload case falls back to zeros.

Tests: 13 transform / 197 polymarket+transform / 1667 full repo.

BY571:
@CharlieHelps review the changes

BY571:
@charliecreates both items addressed in `e766c63`:

**1) Scanner default-vs-docs conflict.** `MarketScannerConfig._filter_markets` now auto-relaxes `min_time_to_resolution_hours` to 0 whenever the user has explicitly opted into upcoming-market targeting (i.e. `slug_prefix` or `max_time_to_resolution_minutes` is set). Both signals already mean "I want short-cadence/upcoming markets"; the 24h floor was internally inconsistent with that intent. Copy-paste `MarketScanner(MarketScannerConfig(slug_prefix="btc-updown-5m-"))` now returns the expected upcoming series with no extra config required.

New `TestAutoRelaxMinTimeWhenTargetingUpcoming` parametrizes over (no targeting / slug_prefix only / max_minutes only / both) and asserts the 24h floor only fires in the no-targeting case.

Live verified: `scan_markets.py --slug-prefix btc-updown-5m-` returns 3 upcoming markets resolving in 5/10/15 min.

**2) Class docstring drift.** `PolymarketBetEnv` step 5 changed from "fetched from Gamma" to "polled from the Polymarket CLOB (`/midpoint` per outcome token)". The runtime path has been on CLOB midpoint since `6f857d1`; this brings the lifecycle description back in line.

Status: 218 polymarket+transform+actor tests, full repo green.

PR #226, opened 2026-04-29
fix(polymarket): retry transient Gamma API failures (#225)

### LINKED ISSUE #225: PolymarketBetEnv: single Gamma API timeout terminates the entire run
## Summary

A single transient `requests.ReadTimeout` from the Polymarket Gamma API causes `PolymarketBetEnv` to silently terminate a long-running session. There is no retry mechanism in `MarketScanner.next_active_market`, so any 15-second network blip during a 24-hour unattended run ends the experiment with hundreds of bets unplaced.

## Repro context

- Long-running dry-run loop: `while not done: step → wait for resolution → fetch next market → repeat`
- After ~25 successful 5-minute markets, Gamma API became briefly unresponsive (`https://gamma-api.polymarket.com/markets`, `read timeout=15`)
- The single failed call terminated the run; 263 of 288 planned markets were never attempted

## What's happening (code references against `8c49e8c`)

`torchtrade/envs/live/polymarket/market_scanner.py`, `next_active_market` (~line 237):

```python
try:
    resp = requests.get(
        f\"{GAMMA_API_BASE}/markets\",
        params={...},
        timeout=15,
    )
    resp.raise_for_status()
    raw_markets = resp.json()
except Exception:
    logger.exception(\"Failed to fetch upcoming markets from Gamma API\")
    return None
```

The `except` block does the right thing locally — log + return `None`. But `PolymarketBetEnv._fetch_next_market` propagates the `None` into the env's step path, where it ends up flipping `done=True` (no further market available) and breaks the consumer's loop. From the consumer's perspective, an unattended 24h experiment quietly stops after one bad packet.

Same pattern exists in `_fetch_markets` (~line 202) for the broader scan path.

## Observed stack trace (snippet)

```
[2026-04-29 12:15:45,286] Failed to fetch upcoming markets from Gamma API
...
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='gamma-api.polymarket.com', port=443): Read timed out. (read timeout=15)
...
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='gamma-api.polymarket.com', port=443): Read timed out. (read timeout=15)

[polymarket_only] final cash: \$945.76  (start \$1,000.00)  wins=10 losses=15 unresolved=0 parse_failures=0
```

## Why the current behaviour is wrong for long-running use

Polymarket markets resolve every 5 minutes; a 24-hour run is 288 sequential round-trips to the Gamma API. At a non-trivial probability of any single 15-second timeout, the *expected* uptime under the current code is much less than 24 hours. The retry cost (a few seconds) is negligible compared to the 5-minute market cadence.

## Proposed fix

Wrap the `requests.get` call in `MarketScanner._fetch_markets` and `MarketScanner.next_active_market` with a retry-with-exponential-backoff loop. Suggested defaults:

- Max attempts: 3
- Backoff: 1 s, 2 s, 4 s
- Worst-case wall clock: 3 × 15 s timeout + 7 s sleep = ~52 s (well under the 5-minute market cadence)
- Only retry on transient classes: `requests.Timeout`, `requests.ConnectionError`, and 5xx responses
- Do not retry on 4xx (these indicate a real client bug, not a transient blip)

A small helper would be reusable across both call sites:

```python
def _fetch_with_retry(url, params, timeout=15, attempts=3, backoff=1.0):
    last_exc = None
    for i in range(attempts):
        try:
            resp = requests.get(url, params=params, timeout=timeout)
            resp.raise_for_status()
            return resp.json()
        except (requests.Timeout, requests.ConnectionError) as exc:
            last_exc = exc
            if i + 1 < attempts:
                time.sleep(backoff * (2 ** i))
        except requests.HTTPError as exc:
            if 500 <= exc.response.status_code < 600 and i + 1 < attempts:
                last_exc = exc
                time.sleep(backoff * (2 ** i))
            else:
                raise
    raise last_exc
```

Plus a complementary fix at the env level: when `next_active_market` returns `None` *despite* the retries, distinguish \"transient API failure\" from \"genuinely no upcoming market in lookahead window\" — only the latter should set `done=True`. The former is recoverable on the next step.

## Why I'm not opening this as a PR yet

Filing as an issue first because (a) the right policy choice for the env-level recovery is a design decision (sleep-and-retry vs propagate vs skip), and (b) the same retry pattern likely belongs around the CLOB calls used during order placement / resolution polling, which I haven't yet exercised. Happy to do the implementation once the design preference is clear.

## Minimal repro

A 24-hour `PolymarketBetEnv` loop in dry-run mode with `--slug-prefix btc-updown-5m-` will reliably hit this within a day or two on most networks; it does not require any unusual configuration. The full producer-side context is in [torchtrade-llm-experiments / experiments/polymarket-llm](https://github.com/TorchTrade/torchtrade-llm-experiments/tree/main/experiments/polymarket-llm).

BY571:
## Summary

Closes #225.

A single 15-second `requests.ReadTimeout` from the Gamma API used to terminate a 24-hour `PolymarketBetEnv` run. After ~25 successful 5-minute markets, a brief network blip would silently end the experiment with hundreds of bets unplaced.

The scanner's `_fetch_json_with_retry` helper now retries transient failures up to 3 times with exponential backoff (1s, 2s) before giving up. Both `MarketScanner.scan` and `MarketScanner.next_active_market` go through the helper, preserving their existing `None`/`[]`-on-failure contract for the caller.

## Behavior

| Failure class | New behavior |
|---|---|
| `requests.Timeout` | Retry up to 3 times |
| `requests.ConnectionError` | Retry up to 3 times |
| 5xx HTTP response | Retry up to 3 times |
| 4xx HTTP response | Fail-fast (real client bug, not transient) |
| Other exceptions | Propagate to outer `except Exception` (unchanged) |

Worst-case wall clock under defaults: 3 attempts × 15s timeout + 1s + 2s ≈ 48s, well under the 5-minute market cadence so retries never push us past the next resolution.

## Why no env-level change

The issue suggested a complementary env-level recovery to distinguish "transient API failure" from "genuinely no upcoming market". After review, this isn't needed:

- After 3 failed retries (multi-minute outage), the scanner returns `None` and `PolymarketBetEnv._step` correctly sets `terminated = True`. Continuing to bet on markets the env can't verify exists would be unsafe in a live trading system.
- The retry policy alone closes the bug: a 24h run would need three independent 15s blips back-to-back to terminate, not one.

## Test plan

- [x] `pytest tests/envs/polymarket/ -v` (119 passed)
- [x] `pytest tests/` full suite (1716 passed, 15 skipped)
- [x] New `TestFetchJsonWithRetry` class covers timeout/conn-error recovery, 5xx retry, 4xx fast-fail, and exhausted-attempts propagation
- [x] New `TestPublicMethodsRecoverFromTransientFailure` parametrizes both call sites (`scan`, `next_active_market`) so a future refactor that bypasses the helper at one site is caught
- [x] Existing scanner tests (`test_scan_returns_empty_on_api_error`, `test_returns_none_on_api_error`) still pass — generic exceptions don't trigger retries and the outer wrapper's graceful-degradation contract is preserved

### DIFF
--- a/tests/envs/polymarket/test_env.py
+++ b/tests/envs/polymarket/test_env.py
@@ -2,7 +2,7 @@
 
 import itertools
 from datetime import datetime, timedelta, timezone
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, call, patch
 
 import pytest
 import torch
@@ -61,6 +61,9 @@ def _make_env(
         "initial_cash": 1_000.0,
         "bet_fraction": 0.1,
         "dry_run": True,
+        # Skip the 30s real-time wait between env-level next-market retries so
+        # tests that exercise the None path don't burn 60+ seconds each.
+        "next_market_retry_sleep_seconds": 0,
     }
     cfg_kwargs.update(config_overrides or {})
     config = PolymarketBetEnvConfig(**cfg_kwargs)
@@ -159,10 +162,12 @@ class TestReset:
         assert env._step_count == 0
 
     def test_raises_when_no_markets(self):
+        """Reset still raises after the env-level retry budget is exhausted."""
         env, scanner, _ = _make_env()
-        scanner.next_active_market.side_effect = [None]
+        scanner.next_active_market.side_effect = [None] * env.config.next_market_max_attempts
         with pytest.raises(RuntimeError, match="No active markets"):
             env.reset()
+        assert scanner.next_active_market.call_count == env.config.next_market_max_attempts
 
 
 # --- Step ------------------------------------------------------------------- #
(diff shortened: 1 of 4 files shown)


BY571:
@CharlieHelps review

PR #245, opened 2026-07-13
fix(live): a liquidation could freeze the trade guard for the rest of the episode

BY571:
## The bug (live money)

`_execute_trade_if_needed` short-circuits when `desired_action == current_action_level`. But **both cached fields are written only by trades the env itself makes** — nothing reconciled them with the exchange mid-episode.

So when a position closes behind the env's back — a **liquidation**, or a manual close in the exchange UI — the cached level stays at its pre-close value. An agent that re-requests the level it used to hold (*the common case for a converged policy repeating its last action*) matches the stale level, the guard calls it redundant, and **the re-entry is silently dropped**. It stays dropped until the policy happens to emit a different level. Live episodes are indefinite real-time loops, so this can persist for hours.

The nastiest part: **it's invisible from the observation.** `_get_observation()` re-reads the exchange every step and is always correct. Only the trade *guard* trusts the stale cache — so the agent can see it's flat while the env quietly refuses to act on it.

Affects **alpaca, binance, bitget** (they have the guard). bybit/okx have no guard and were immune, but get the same call so a future guard can't silently reintroduce it.

## The fix

`TorchTradeLiveEnv._sync_position_from_exchange()`, called at the top of every non-SLTP `_step()` **before** the guard. It reuses the `position_status` `_step` already fetches, so it costs **no extra API call**. On a mismatch (the change wasn't ours) the level is NaN'd — the level behind a position we did not open is unknowable, and NaN never compares equal, so the next command always executes. On a match the level is left alone, which is what still lets the guard suppress genuinely redundant trades.

This is the mid-episode counterpart of #243, which fixed only the reset-time case.

## What the review found (and it was brutal)

Ten rounds of `torchrl-engineer` + `code-simplifier` + `pr-test-analyzer` + `test-justification`. **The first version of this fix did not work**, and the reviewers caught every subsequent problem. Recording it honestly:

| round | finding |
|---|---|
| 1 | **The fix didn't work.** Exact `qty == 0` missed dust residuals — a `qty=1e-12` liquidation still froze the guard. The `SLTPMixin` I copied the shape from already had `POSITION_DUST_EPS`; I took the method but not the hard-won constant. |
| 2 | "One dust rule" was **false** (5 resets still hand-rolled exact-zero). My rename **silently deleted a guard** — a reviewer re-forked binance's sync with the old broken rule and the suite stayed **green**. |
| 3 | I'd fixed the env's *internal state*, **not `account_state`** — the agent still saw a phantom position. Two of my docstrings claimed otherwise. |
| 4 | The agent sees **six** numbers; I'd fixed one. `holding_time` grew on a flat book; the agent saw a position **4% from a liquidation that didn't exist**. My own fixtures hid it (`liquidation_price=0.0` short-circuits that branch). |
| 5 | **The entire Alpaca fix was untested** — delete it, suite green. My test zeroed the fields it was checking. |
| 6–8 | Three surviving mutants, each hiding behind a **dead assertion** of mine (`assert hold_counter == 0` on a *fresh* env passes no matter what the code does). |
| 9–10 | **The inverse hole:** I spent the whole PR proving a *flat* account reads flat, and never checked that an **open** position reads open. A mutation handing the policy its 20×-levered position as *flat, unlevered, and infinitely far from liquidation* passed all 1977 tests. |

Every one of those is now mutation-verified. Corrupting any element of `account_state` — flat **or** open — dies in every exchange that has that element.

## Not fixed (deliberate, tracked separately)

- **`holding_time` doesn't reset on a direct position flip** (short→long as one order). Pre-existing, byte-identical to `main`.
- For alpaca/binance, `hold_counter` logic is **duplicated** in `env_sltp.py` and unpinned there.

## Suite

**1981 passed**, 25 skipped, 1 xfailed.

> ⚠️ Unrelated: `tests/actor/test_frontier_llm_actor.py::test_generate_batch_preserves_order` only passes because an untracked `.env` supplies `OPENAI_API_KEY`. It fails on a clean checkout **of `main` too**. CI on a fresh clone is not green.

### DIFF
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -186,8 +186,8 @@ class TestAlpacaTorchTradingEnvReset:
         """A dust residual on reset is flat -- internally AND in what the agent sees.
 
         A close can leave a float residue (1e-12). Read as an open position it puts a phantom
-        direction in account_state with zero exposure -- an observation the policy never saw
-        in training -- and freezes the trade guard.
+        direction in account_state at zero exposure -- a combination the policy never saw in
+        training.
 
         Behavioural on purpose: the structural guard that every _reset uses the shared rule
         can be dodged by moving the derivation into a helper. This cannot.
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -185,20 +185,27 @@ class TestAlpacaTorchTradingEnvReset:
     def test_reset_reads_dust_as_flat(self, env):
         """A dust residual on reset is flat -- internally AND in what the agent sees.
 
-        A close can leave a float residue (1e-12). Read as an open position it puts a phantom
-        direction in account_state at zero exposure -- a combination the policy never saw in
-        training.
+        A close can leave a float residue (1e-12) behind, with a STALE entry price and market
+        value still attached. Read as an open position, those stale fields become the agent's
+        exposure and unrealized PnL for a position that does not exist.
 
-        Behavioural on purpose: the structural guard that every _reset uses the shared rule
-        can be dodged by moving the derivation into a helper. This cannot.
+        The stale values are the whole point: an earlier version of this test zeroed them, so
+        every element it checked was already 0 whatever the code did, and the fix it was
+        guarding could be deleted with the suite still green.
         """
         env.trader.position_qty = 1e-12
-        env.trader.position_value = 0.0
+        env.trader.position_value = 41.82        # stale market value left on the residual
+        env.trader.avg_entry_price = 45000.0     # stale entry -> a fake +22% unrealized PnL
 
         td = env.reset()
 
         assert env.position.current_position == 0
-        assert td["account_state"][1].item() == 0.0   # the direction the AGENT sees
+
+        exposure, direction, unrealized_pnl, holding_time, _lev, _dist = td["account_state"].tolist()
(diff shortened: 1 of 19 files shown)


BY571:
Fixed in `76e2d74`. You were right, and the bug was in the code this PR adds.

Reproduced first (alpaca): hold 5 bars → liquidate between steps → the agent re-enters at the same level in *that same* `_step`:

```
after 5 bars long                : hold_counter = 5
liquidation + same-step re-entry : hold_counter = 6, account_state[3] = 6.0
```

A brand-new position handed to the policy as six bars old.

**Fix:** a mismatch in `_sync_position_from_exchange` now discards `hold_counter` — in **both** syncs (`TorchTradeLiveEnv` and `SLTPMixin`, which had the same gap you noted). That sync is the only code that knows the position it was counting is gone.

**Regression:** `test_reentry_after_external_close_starts_a_fresh_holding_time` on all five exchanges — external close → same-step re-entry → asserts `account_state[3]` starts from the new position. Reverting the fix fails it. You called out exactly why the old tests couldn't see this: they only assert `trade()` was called.

One test change, not a code change: binance's bare fixture reported "no position" while the env believed it was long. With the sync correctly resetting on that disagreement, the counter could never age — the mock was unrealistic. It now reports what the env actually holds.

Suite: 1986 passed.

---

Worth saying: ten rounds of my own mutation-testing review missed this. Mutation kills *lines* — this bug is in a *sequence* (external close **and** re-entry in the same step), which no single-line mutant can produce. You read what the fix makes possible and asked what happens next. That's the check I didn't have.

PR #246, opened 2026-07-13
fix(alpaca): ask Alpaca for the qty precision instead of round(amount, 1)

BY571:
## The bug

`AlpacaOrderClass` quantity mode did `round(amount, 1)`. **Any quantity below 0.05 became `0.0`** — an order for 0.01 BTC submitted `qty=0.0`. BTC/USD's real increment is `0.0001`, so the old rule was wrong by three orders of magnitude.

It was pinned by a strict `xfail`; that `xfail` now XPASSes and is removed.

## The fix: ask the exchange, don't guess

The precision is **not guessed**. `alpaca-py`'s `Asset` model already carries it:

```python
class Asset:
    fractionable: bool
    min_order_size: Optional[float]       # crypto only
    min_trade_increment: Optional[float]  # crypto only
```

This matters: `min_order_size` is **dynamic** (for USD pairs it tracks ~`10/price`), so hardcoding would have been wrong even with the right number of digits today.

So Alpaca now follows the same "ask the exchange" pattern the other exchanges already use via CCXT (`bitget/order_executor.py::get_lot_size`): cached `get_asset()` lookup, then **floor** onto the grid — floor, never round, because rounding up can exceed the size (and the buying power) the caller asked for. `Decimal`, because `0.0001` isn't exactly representable in binary float.

Fallbacks: the asset's increment (crypto) → whole shares if not fractionable → **9 dp**, the documented API maximum. That 9 dp ceiling is the only hardcoded number left, it's the *loosest legal bound*, and it can only floor away digits Alpaca would have rejected anyway — it cannot reproduce the old silent `qty=0.0`. If the lookup fails, the failure mode is now a **loud exchange rejection** rather than a silent zero-quantity order.

## Two adjacent bugs found in the same path

1. **The `amount = -1` "sell everything" sentinel leaked into quantity mode.** Notional mode intercepts sells and full-closes; quantity mode had no such intercept, so it would have submitted a literal **`qty = -1`**. Now routes to `close_position()`.
2. **A sub-increment quantity was floored to `0.0` and submitted anyway.** Now refused with a warning.

## Mock change was load-bearing, not cosmetic

`MockTradingClient` gained `get_asset()`/`MockAsset`. The mock is duck-typed with no `spec=`, so calling `get_asset()` without it would raise `AttributeError` → swallowed by `trade()`'s broad `except Exception` → **every existing Alpaca order test would have started passing for the wrong reason** ("order not submitted").

## Verification

Mutation-verified: reverting to `round(amount, 1)` fails **8 tests**; changing the floor to a round fails 2.

**Full suite: 1919 passed.**

## Review status — stated honestly

**Zero review rounds.** This branch is implemented and mutation-verified but has **not** been through the mandated 3-round agent review. I'm not going to claim otherwise. Review before merge.

### DIFF
--- a/tests/envs/alpaca/test_order_executor.py
+++ b/tests/envs/alpaca/test_order_executor.py
@@ -13,7 +13,7 @@ from torchtrade.envs.live.alpaca.order_executor import (
     OrderStatus,
     PositionStatus,
 )
-from tests.mocks.alpaca import MockTradingClient, MockOrder, MockOrderStatus
+from tests.mocks.alpaca import MockTradingClient, MockOrder, MockOrderStatus, MockAsset
 
 
 # Standalone by design: the shared exchange test base assumes an enum TradeMode and a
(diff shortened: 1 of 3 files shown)


BY571:
All three fixed in `1c9da26`. Every one was mine, and all three were the same mistake — I got Alpaca's rule *nearly* right.

**1. Stale cache (`order_executor.py:109`)** — you're right, and my own PR body says `min_order_size` is dynamic (~`10/price`). Caching it was indefensible. Now **not cached**: one `get_asset()` per order. Orders here are at most one per bar; the call is cheap and a stale minimum isn't.

**2. Unsafe fallback (`:117`)** — agreed. 9 dp is the max precision for a *fractionable equity*, not a safe default for an unknown asset, and caching the exception as valid metadata poisoned every later order. Now **fails closed**: if the asset's rules can't be read, the order is refused rather than guessed. The failure mode is now a refusal, not a rejected order.

**3. Wrong lattice (`:134`)** — correct, and your example reproduces exactly. Valid quantities are `min + N*step`, not `N*step`. The floor is now anchored at the minimum:

```python
steps = ((qty - min_d) / step_d).to_integral_value(rounding=ROUND_DOWN)
return float(min_d + steps * step_d)
```

Your case (min `0.000166`, step `0.0001`, amount `0.00027`) now yields `0.000266` instead of the off-lattice `0.0002`.

Each has its own test, and each is mutation-verified — reverting the cache, the fail-closed, or the lattice anchor fails exactly one test:

- `test_min_order_size_is_not_cached_across_price_moves`
- `test_asset_lookup_failure_fails_closed`
- `test_quantity_lands_on_the_min_plus_n_steps_lattice`

Suite: 1921 passed.

PR #248, opened 2026-07-14
fix(polymarket): the bankruptcy docstring was lying about the wallet + pin two untested branches

BY571:
## What this actually is

It started as *"deduplicate the bankruptcy rule"* — PR #244 hoisted `_check_termination` into `TorchTradeLiveEnv`, but `PolymarketBetEnv` extends `EnvBase` directly, so the hoist couldn't reach it and it kept a hand-rolled copy.

**The dedup turned out to be the least valuable part.** What's worth shipping:

### 1. A false docstring on a real-money env

The config said:

> `initial_cash`: Starting USDC balance used for dry-run accounting; **live mode reads the real wallet via the trader.**

**It does not.** `self.cash` is simulated bookkeeping in *both* `dry_run` and live mode. Nothing anywhere reads the USDC wallet — `PolymarketOrderExecutor` has no balance method at all. `_reset()` restores the constant every episode, so real losses between episodes are erased from the baseline.

**So on a live-money env, the bankruptcy check is checking a fiction.** The docstring now says so. (The actual wallet-sync fix needs a balance-fetch capability that doesn't exist yet — separate, larger.)

The same lie also lived in the **class** docstring, 55 lines below. Caught in review. Fixed.

### 2. Two branches that had no coverage

Polymarket's bankruptcy test drove cash to `0` against a threshold of `5` — far past the boundary. So:
- flipping the strict `<` to `<=` — **passed**
- deleting the `done_on_bankruptcy` gate — **passed**

Both now pinned, through `env._is_bankrupt()` rather than the helper, so a **broken wiring** is caught: passing `bet_fraction` instead of `bankrupt_threshold`, hardcoding the gate, or swapping `cash`/`initial_cash` each now fail a test.

### 3. The dedup itself (a wash)

`is_bankrupt(*, current, initial, threshold, enabled)` in `torchtrade/envs/utils/termination.py`. Both `TorchTradeLiveEnv._check_termination` and `PolymarketBetEnv._is_bankrupt` call it with their own domain-correct inputs (portfolio value vs cash — Polymarket has no carried position to mark; the bet resolves into cash before the check).

**Polymarket deliberately stays on `EnvBase`.** It's a one-shot bandit: no carried position, no duplicate-action guard, event-driven waits rather than a bar clock. Forcing it under `TorchTradeLiveEnv` would drag in `_sync_position_from_exchange` and a constructor it can't satisfy. Only the arithmetic is shared, so only the arithmetic moved.

## What review removed

I added an `initial > 0` guard. **It was defensive code against an unreachable state** — it fires only when `portfolio < 0` *and* `initial <= 0` simultaneously, and an account that started at zero cannot trade into a negative balance.

Measured: deleting the guard killed exactly 4 tests — **all 4 written by me to justify the guard.** Nothing else depended on it. I'd added a guard, written tests to pin it, introduced a live-env behaviour change, and filed a follow-up about that behaviour change — for a branch that cannot execute.

It's gone. **The rule is now behaviourally identical to `main`** — no behaviour change to explain.

## Verification

Every mutation dies: `<` → `<=` (2), gate dropped (3), wrong config field (1), hardcoded gate (1), `cash` → `initial_cash` (3).

**Full suite: 2002 passed.**

Reviewed by `torchrl-engineer`, `code-simplifier` (twice), `pr-test-analyzer`, and a `test-justification` pass.

### DIFF
--- a/tests/envs/polymarket/test_env.py
+++ b/tests/envs/polymarket/test_env.py
@@ -319,7 +319,7 @@ class TestStep:
         # The two rows that pin the WIRING. The boundary and the arithmetic belong to the shared
         # rule and are pinned in tests/envs/test_termination.py -- asserting them again through
         # the env is the layering this test exists to avoid.
-        (199.99, True, True),   # sole killer of every wrong-config-field mutant
+        (199.99, True, True),   # sole killer of the bet_fraction mutant (both are fractions)
         (0.0, False, False),    # sole killer of a hardcoded gate
     ], ids=["just-below", "gate-off-while-broke"])
     def test_is_bankrupt_wiring(self, cash, enabled, expected):
@@ -333,8 +333,6 @@ class TestStep:
         against a threshold of 5, far past the boundary.
         """
         env, _, _ = _make_env(
-            outcomes=[0],
-            markets=[_make_market(), _make_market()],
             config_overrides={
                 "initial_cash": 1000.0,
                 "done_on_bankruptcy": enabled,
--- a/tests/envs/polymarket/test_env.py
+++ b/tests/envs/polymarket/test_env.py
@@ -316,11 +316,12 @@ class TestStep:
         assert td["terminated"].item()
 
     @pytest.mark.parametrize("cash,enabled,expected", [
-        (200.0, True, False),   # exactly at the threshold -> NOT bankrupt (the check is a strict <)
-        (199.99, True, True),   # a hair below -> bankrupt
-        (0.0, False, False),    # wiped out, but the gate is off -> keep betting
-        (0.0, True, True),      # wiped out, gate on
-    ], ids=["at-threshold", "just-below", "gate-off-while-broke", "gate-on-while-broke"])
+        # The two rows that pin the WIRING. The boundary and the arithmetic belong to the shared
+        # rule and are pinned in tests/envs/test_termination.py -- asserting them again through
+        # the env is the layering this test exists to avoid.
+        (199.99, True, True),   # sole killer of every wrong-config-field mutant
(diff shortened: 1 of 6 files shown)


BY571:
Good catch — and my "the guard is unreachable" reasoning was wrong in a way I hadn't probed.

The state is unreachable **via trading** (an account that started at zero can't trade into a negative balance), but reachable **via configuration**, since `PolymarketBetEnvConfig` is an unvalidated dataclass. Reproduced exactly as you described:

```
initial_cash = -100, cash after reset = -100
  old _is_bankrupt(): False    <- the `initial > 0` guard swallowed it
  new _is_bankrupt(): True
```

(My own first check used `current=0.0` and showed no divergence — the real reset state is `cash = initial_cash`, which does.)

**Fixed by validating at the boundary rather than restoring the guard** (`de5eac5` → latest).

The guard is what made the nonsense config *silent*: an env with negative starting capital constructed fine and simply never terminated — a disabled safety stop, absorbed instead of reported. That's the same silent-default pattern this repo keeps hitting (bitget's hardcoded lot size, alpaca's `round(amount, 1)`, the live envs' `balance.get(key, 0)`). Restoring the guard would preserve the old behaviour by preserving the silence.

`initial_cash <= 0` now raises at construction, so the state the guard defended against is genuinely impossible rather than quietly tolerated.

Scoped to `initial_cash` only — my first pass also validated `bet_fraction` and `bankrupt_threshold` and immediately broke `test_zero_stake_in_live_mode_does_not_call_trader`, since `bet_fraction=0` is a supported config here.

Suite: 2005 passed.

Issue #45, opened 2026-01-04
Feature: Prioritized Reset Sampling for Coverage-Based Training


BY571:
### Description

Implement prioritized reset sampling to encourage exploration of under-visited regions of the dataset during training. This feature builds on the coverage tracking metric from issue #41 and enables curriculum-style learning.

### Motivation

During RL training with random episode resets, certain starting positions may be visited more frequently than others due to pure randomness. This can lead to:
- Overfitting on frequently-sampled market conditions
- Under-exploration of rare but important scenarios
- Uneven learning across the dataset

By tracking which starting positions have been visited and prioritizing under-explored regions, we can:
- Ensure more comprehensive coverage of training data
- Improve generalization to diverse market conditions
- Enable curriculum learning strategies

#### 1. Coverage Tracking Foundation

First, implement coverage tracking in `MarketDataObservationSampler` (from issue #41):

```python
class MarketDataObservationSampler:
    def __init__(self, ...):
        # Existing code...

        # Coverage tracking
        self.coverage_counts = np.zeros(len(self.exec_times), dtype=np.int32)
        self.total_resets = 0

    def reset(self, random_start: bool = False) -> int:
        # Track which start_idx was selected
        # ... existing reset logic ...

        self.coverage_counts[start_idx] += 1
        self.total_resets += 1
        return self._end_idx - self._sequential_idx
```

#### 2. Prioritized Sampling Strategy

Add prioritized reset mode that samples under-visited positions with higher probability:

```python
class MarketDataObservationSampler:
    def reset(
        self,
        random_start: bool = False,
        prioritized: bool = False,
        epsilon: float = 0.1
    ) -> int:
        """
        Reset the sampler.

        Args:
            random_start: If True, start from random position
            prioritized: If True, prioritize under-visited positions
            epsilon: Probability of uniform random sampling (exploration)
        """
        total_len = len(self._exec_times_arr)

        if random_start:
            if self.max_traj_length is None:
                max_start_index = total_len - 1
            else:
                max_start_index = max(0, total_len - self.max_traj_length)

            if prioritized and self.total_resets > 0:
                # Epsilon-greedy: sometimes uniform, sometimes prioritized
                if self.np_rng.random() < epsilon:
                    # Uniform random (exploration)
                    start_idx = int(self.np_rng.integers(0, max_start_index + 1))
                else:
                    # Prioritized sampling (exploitation of under-explored regions)
                    # Compute inverse frequency weights
[...]

PR #249, opened 2026-07-14
fix(polymarket): refuse live trading instead of silently mis-trading it

BY571:
## The bug, and why the obvious fix would have been worse

Follow-up #45: `PolymarketBetEnv.cash` was simulated bookkeeping in **both** `dry_run` and live mode — nothing ever read the real wallet, and `_reset()` restored a constant every episode. So the bankruptcy check on a live-money env was checking a fiction.

The obvious fix — read the real USDC balance — is **actively dangerous here**, and that took real research to establish:

**This env never sells.** It buys and holds every bet through resolution. Polymarket resolution does **not** release collateral: a winning share is worth $1 but stays an ERC-1155 token until an explicit on-chain `redeemPositions` call, which **no Polymarket client exposes** (V1, V2, or the unified py-sdk). A live bot that never redeems watches its spendable collateral **drain monotonically to zero while it is winning.**

So a wallet-backed `cash` would make a *winning* agent declare itself bankrupt and halt. Strictly worse than the bug it replaced.

**And the live path is dead anyway.** `py-clob-client`, which `order_executor.py` is built on, was archived in May 2026 — *"no longer functional and should not be used for new or existing integrations."* Polymarket shipped CLOB V2 with new contracts and pUSD replacing USDC.e. It was never even a declared dependency, just an optional `try/except` import — so the live path was almost certainly never exercised. Meanwhile `dry_run` **defaulted to `False`**: the default config was live mode.

## What this does

Refuses live at the config boundary rather than pretending. `dry_run=False` raises with both reasons and the sources; `dry_run` now defaults to `True`. In paper mode the simulated cash isn't a bug — it's the correct semantics of a paper env. The lie was only ever that live mode *pretended* to be real.

Reviving live needs the CLOB V2 port **and** a redemption workflow. Tracked separately.

## Bugs found and fixed along the way

- **`close()` broke `TransformedEnv`** (pre-existing). `def close(self)` didn't accept `EnvBase.close`'s keyword-only `raise_if_closed`, which `TransformedEnv` forwards — so `TransformedEnv(env, RewardSum()).close()` raised `TypeError`, on the composition the env's own spec comment advertises.
- **The policy's entire observation vector was unpinned** (pre-existing). Swapping `spread`↔`liquidity`, or emitting `no_price` instead of `yes_price` (feeding the policy the *inverted* market), passed the whole suite. `check_env_specs` can't catch a permutation — `Bounded(0, inf)` accepts any order. Same class as CLAUDE.md invariant #3.
- **`_step`'s observation was unpinned too** — returning the stale, already-resolved market passed everything. Live, the policy conditions on the old market's quotes while the bet's `fill_price` comes from the new one: *the agent bets at prices it never saw.*
- **The wait→poll order was observed by nothing** — deleting `_wait_for_resolution`, or polling first, both passed. Live that polls a market that hasn't ended, burns the budget, and books **reward=0 on every bet**.
- **`bankrupt_threshold` was unvalidated** — `< 0` silently disables the safety stop. Its domain also disagreed with `SequentialTradingEnvConfig`'s already-settled `[0, 1)`.
- Four config knobs (grace, poll interval, max wait, and both lower bounds) were pinned only at their **defaults**, so a hardcoded constant passed. `max_wait` didn't even fail — it *hung*.
- **Dead code**: `_compute_payoff`'s `stake <= 0` guard became unreachable once the boundary validation landed (proven: 200k simulated episodes, min cash `0.0`, zero negative-cash episodes). Deleted rather than tested — validate at the boundary, don't guard in the rule.

## Docs

The README section literally titled **"Running for real"** still told users to `pip install py-clob-client` and set `dry_run=False`; the official docs still documented a funded USDC.e wallet; the root README listed Polymarket under *"Start live trading."* All corrected. The payoff is now honestly described as **modelled** (it fills at the quoted mid, with no spread crossing and no fees) rather than "realized".

## Verification

- **2016 passing**, 25 skipped.
- **15/15 mutations killed**, each verified to actually apply — including all four config knobs, both observation paths, the compounding rule, the resolution threshold, and the lifecycle.
- Reviewed over **4 rounds × 4 agents** (test-justification, code-simplifier, pr-test-analyzer, torchrl-engineer). Rounds 2, 3 and 4 each found real bugs *in my own fixes* — including several dead assertions and a false claim that had migrated into the one string this PR designated canonical.
- TorchRL verified by execution, not inspection: `check_env_specs`, rollout, `TransformedEnv+RewardSum`, `SerialEnv`, and `ParallelEnv` under both fork and spawn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,8 @@ See **[Online Environments Documentation](https://torchtrade.github.io/torchtrad
 
 ## Trading Platforms
 
-Start live trading with these supported platforms:
+Live trading is supported on the platforms below. **Polymarket is the exception: it is
+paper-only** — see its entry.
 
 ### 🪙 Cryptocurrency Trading
 
@@ -182,7 +183,7 @@ Start live trading with these supported platforms:
 **[Polymarket](https://polymarket.com/)** - Decentralized prediction market on Polygon
 - **Supported by:** `PolymarketBetEnv`
 - **Features:** Rolling one-shot bets on short-cadence binary markets (BTC/ETH/SOL up-or-down at 5m / 15m / 1h / 4h / daily cadences), Gamma API market scanner
-- **⚠️ Paper trading only:** `dry_run=False` raises. `py-clob-client` is archived/non-functional (Polymarket moved to CLOB V2), and the env holds bets to resolution while Polymarket only releases collateral on an on-chain redeem no client exposes — so a live bot would drain to zero *while winning*. Reviving live needs the V2 port **and** a redemption workflow.
+- **⚠️ Paper trading only — not live:** `dry_run=False` raises. `py-clob-client` is archived/non-functional (Polymarket moved to CLOB V2), and the env holds bets to resolution while Polymarket only releases collateral on an on-chain redeem no client exposes — so a live bot would drain to zero *while winning*. Reviving live needs the V2 port **and** a redemption workflow.
 - **Get Started:** [Browse markets at Polymarket](https://polymarket.com/)
 
 ### 📈 Stock & Crypto API
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ TorchTrade provides modular environments for both live trading with major exchan
 - 🎯 **Multi-Timeframe Observations** - Train on 1m, 5m, 15m, 1h bars simultaneously
 - 🤖 **Multiple RL Algorithms** - PPO, DQN, IQL, GRPO, DSAC, CTRL implementations
 - 📊 **Feature Engineering** - Add technical indicators and custom features
-- 🔴 **Live Trading** - Direct Alpaca, Binance, Bitget, Bybit, OKX, and Polymarket integrations
+- 🔴 **Live Trading** - Direct Alpaca, Binance, Bitget, Bybit, and OKX integrations (Polymarket is paper-only)
 - 🧠 **LLM Integration** - Use GPT-4o-mini or local LLMs as trading agents
 - 🔧 **LLM Tool Use** - Let LLM agents call tools mid-reasoning (e.g. live Google News for sentiment) before choosing an action
 - 🎓 **LLM GRPO Fine-tuning** - Train/finetune a local LLM actor on your data with GRPO ([guide](docs/guides/llm-grpo-training.md))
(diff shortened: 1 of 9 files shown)


BY571:
Both findings were right, and both are fixed in c69cdd9. Thank you — the first one is a live-money hole I talked myself into.

## 1. Public executor permitted live trading

You're right, and my reasoning was bad. I put the guard on `PolymarketBetEnvConfig` and left `PolymarketOrderExecutor` constructible with `dry_run=False`, on the grounds that it was inert "scaffolding for the V2 port". But it's exported from both `torchtrade.envs.live` and `torchtrade.envs.live.polymarket` — so a caller using the exported API could build a real `ClobClient` from a real key and post an order, while the env believed itself paper-only. **Guarding the env that happens to construct the class, instead of the class that can actually move money, is the wrong place for the guard.**

`PolymarketOrderExecutor` now refuses `dry_run=False`.

That made every V1 path in it unreachable, so rather than park dead code behind the guard I deleted it — `ClobClient` construction, API-cred derivation, `MarketOrderArgs`/`OrderType.FOK` posting, `cancel_all`'s live branch, and the `py_clob_client` import shim. It wouldn't have survived the CLOB V2 port regardless (different package, different types, pUSD collateral, plus a redemption workflow that doesn't exist), and git history has it if the port wants a reference. **`py_clob_client` is now referenced nowhere in the tree.**

Regression tests added, including the refusal holding against a fully-funded-looking config (key + chain + proxy `signature_type` + funder), and pinning the message's *substance* rather than a contentless phrase.

## 2. `private_key` removal — worse than you flagged

You reported the keyword break. The positional case is nastier, and I'd missed it: `private_key` was the **second positional parameter**, so removing it didn't just `TypeError` —

```python
PolymarketBetEnv(config, "0xkey")   # old: private_key.  new: silently binds to scanner
```

— it **silently rebound the key to `scanner`**, which then fails much later with a confusing `AttributeError` on a `str`. A silent break, which is precisely the bug class this PR exists to fix, introduced by me while fixing it.

Fixed both ways:
- Everything after `config` is now **keyword-only**, so the positional form is a loud `TypeError` at the call site.
- `private_key` is still **accepted** as a keyword purely so it can be **refused with an explanation**, instead of a bare "unexpected keyword argument".

I went with refuse rather than accept-and-ignore (which is what you suggested): silently swallowing a real private key would tell the caller their key is configured when the env can never sign anything. The error names the change and says why. If you'd still prefer a deprecation-warning-and-ignore, say so and I'll switch — but on a paper-only env I'd rather a passed key be an error than a no-op.

Suite: 2012 passing. Both fixes mutation-verified. One equivalent mutant noted honestly in the commit (`self._dry_run = True` → `= dry_run` survives, correctly — past the guard `dry_run` can only be `True`).

BY571:
@CharlieHelps review

PR #251, opened 2026-07-14
fix(live): a direct long→short flip reported the dead position's age to the policy

BY571:
## The bug

Every live env reset `holding_time` **only when flat**:

```python
if position_direction == 0:
    hold_counter = 0
else:
    hold_counter += 1
```

A **direct flip never passes through flat**, so the reset never fires:

```
step 1-5   long     hold_counter → 1,2,3,4,5
step 6     SHORT    hold_counter → 6      ← a BRAND-NEW position, reported as 6 bars old
```

`holding_time` is **`account_state[3]`** — the policy conditions on it directly. A hold-time-aware policy reads a position it *just opened* as one it has been sitting in for six bars.

**Reachable with the default config.** Default action levels are `[-1, 0, 1]` (binance) and `[-1, -0.5, 0, 0.5, 1]` (bitget/bybit/okx), so `+1 → -1` in one step is an ordinary action, not an exotic path. Verified by running the counter, not by reading it.

Five exchanges, two different hand-rolled spellings, the same hole in all of them.

## The fix

One rule, in `core/state.py`:

| transition | holding_time |
|---|---|
| flat → long | 1 |
| long → long | 2, 3, 4, … |
| **long → SHORT** | **1** ← the case every env missed |
| → flat | 0 |

A fresh open still reports `1`, so the normal path is unchanged — the only behaviour that moves is the flip. `PositionState` gains `hold_direction` (the direction the counter is counting), which `_reset()` now clears too.

Alpaca is spot and **cannot** flip, so its rule was already correct — but it's converged onto the shared one anyway (proven behaviourally identical for long-only levels). An exception you have to remember is how drift starts; this repo has already paid for that with three divergent SLTP action maps and four copies of `_check_termination`. There's a structural guard banning `hold_counter += 1` outside the rule.

The **offline envs are unaffected**: they gate on `action_value == self._prev_action_value`, so a flip changes the action and falls through to the trade path, which resets the counter.

## The test I got wrong first, and why it matters

My first end-to-end test was **vacuous** — reverting the fix left it green.

`_step` calls `_sync_position_from_exchange` *before* the trade executes. Mocking the short for the whole step made the sync see `exchange(short) != cache(long)`, treat it as an **external** change, and reset the counter itself (PR #245's logic) — masking the bug entirely.

A real flip: at sync time the trade hasn't executed, so the exchange still shows the **old long**; only when `_get_observation` re-reads it does it show the short. Both tests now model that, and reverting the fix fails them:

- bybit — *"a one-step-old short is reported as 7.0 bars old"*
- binance — *"a one-step-old short is reported as 6.0 bars old"*

**Both** are tested because they use *different wirings* — binance advances from the env's **cached** direction, bybit/bitget/okx from the direction **observed** on the exchange. A unit test on the rule can't tell those apart; only an end-to-end assertion on `account_state[3]` can.

## Verification

- **2026 passing.**
- 5/5 unit mutations killed, plus both end-to-end wirings.

### DIFF
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -1,7 +1,8 @@
-"""Shared observation-class tests.
+"""Shared test bodies for the per-exchange live-env test files.
 
-Each exchange (alpaca, binance, bitget, bybit, okx) subclasses BaseObservationClassTests in
-its own test_obs_class.py and implements create_observer / get_expected_symbol_format.
+- BaseObservationClassTests: subclassed by each exchange's test_obs_class.py.
+- assert_a_direct_flip_does_not_age_the_new_position: called by each exchange's
+  test_torch_env_futures*.py, so the flip contract has one body and five kills.
 """
 
 import pytest
@@ -269,7 +270,6 @@ class BaseObservationClassTests(ABC):
         assert observations[key].shape[0] == 100
 
 
-
 def assert_a_direct_flip_does_not_age_the_new_position(
     env, trader, PositionStatus, long_action, short_action, set_side=None
 ):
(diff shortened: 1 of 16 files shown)


BY571:
@CharlieHelps review

PR #255, opened 2026-07-16
refactor(envs): unify futures _get_observation into TorchTradeFuturesLiveEnv

BY571:
## What

The 4 live **futures** envs (binance/bitget/bybit/okx) each hand-rolled a **byte-identical** `_get_observation` + `_get_portfolio_value` (~90–120 lines apiece). Every `account_state` fix therefore had to land in 4 copies — the "fix the instance, not the class" trap that made `holding_time` (#252) and the liq-distance guard (#253) 4-way sweeps.

This lifts one shared copy into a new intermediate base **`TorchTradeFuturesLiveEnv(TorchTradeLiveEnv)`**; the 4 futures base envs inherit it and delete their copies.

## Behavior-preserving

- The shared body is **byte-identical to the old bybit/okx** `_get_observation`.
- binance/bitget differed **only** by two dead locals (`cash`, `entry_price` — assigned, never read), which are removed.
- **Alpaca (spot) is untouched** — it keeps its own `_get_observation` (leverage=1, no liquidation, cash-based) and does not inherit the futures base.

Verified: shared method diffed byte-for-byte against the old per-exchange versions; the precise per-exchange `account_state` assertions (exposure, leverage, dist_to_liq, holding_time, dust) all still pass.

## Guardrails

- **New structural guard** `test_no_futures_env_reforks_the_shared_observation` — asserts all **12** futures classes (base + Futures + FuturesSLTP × 4) resolve `_get_observation`/`_get_portfolio_value` to the shared base **by identity**, so a future re-fork fails loudly instead of silently un-doing the dedup. Mutation-verified.
- The existing AST hold-counter guards already scan the new file (via the `live/**` rglob) — confirmed by mutation (hand-rolled aging in the shared file is flagged).
- Added a test for the now-consolidated `include_base_features=True` passthrough branch (previously dark; futures configs default it off).

## Two pre-existing bugs surfaced during review — logged, NOT fixed here

This PR is deliberately behavior-preserving; both get their own PR + sign-off:
1. **`exposure_pct` denominator** (`account_state[0]`): Binance's `total_wallet_balance` **excludes** unrealized PnL, while bitget/bybit/okx map equity (incl. uPnL) to the same key → cross-exchange skew. Fix: use `total_margin_balance` (uniformly equity).
2. **`base_features` spec mismatch**: binance/bitget/bybit emit a `base_features` observation key absent from their `observation_spec` (only okx declares it) → `check_env_specs` fails for those 3.

Reviewed over the mandated **3 rounds × 4 agents**; Round 3 clean on all four. Full suite **2156 passed, 0 failed**.

### DIFF
--- a/tests/envs/bybit/test_torch_env_futures.py
+++ b/tests/envs/bybit/test_torch_env_futures.py
@@ -85,6 +85,27 @@ class TestBybitFuturesTorchTradingEnv:
         assert env.action_spec.n == 5  # [-1.0, -0.5, 0.0, 0.5, 1.0]
         assert env.action_levels == [-1.0, -0.5, 0.0, 0.5, 1.0]
 
+    def test_include_base_features_flows_through_shared_get_observation(self, mock_observer, mock_trader):
+        """The shared TorchTradeFuturesLiveEnv._get_observation passes base_features through.
+
+        The futures envs default include_base_features=False, so this branch of the now-shared
+        method was untested; one case here covers it for all four exchanges via inheritance.
+        """
+        from torchtrade.envs.live.bybit.env import (
+            BybitFuturesTorchTradingEnv,
+            BybitFuturesTradingEnvConfig,
+        )
+
+        config = BybitFuturesTradingEnvConfig(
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps review

PR #263, opened 2026-07-29
fix(examples): make offline IQL example runnable + drop dead HF dataset refs

BY571:
## Summary

Four library issues surfaced while working through the examples. Verifying the third turned up four more, because the offline IQL example had **four independent blockers** and its smoke test had been commented out with `# TODO: Enable offline IQL test once encoder shape mismatch is fixed` rather than the bug fixed.

## Reported issues

1. **IQL reward shape.** `reward/done/terminated` were built as `(n,)`; value estimators need `(*batch, 1)` to match `state_value`, so `DiscreteIQLLoss` raised `RuntimeError: All input tensors ... must share a unique shape`. Normalised at the single chokepoint in `make_offline_replay_buffer` — `dataset_to_td` passes HF column shapes through verbatim, so the HF branch had the identical latent defect. `action` deliberately left at `(n,)`; unsqueezing it instead trips IQL's `log_prob`/`min_Q` guard.
2. **Five dead HF dataset paths** (all 401): `ethusdt_spot_1m`, `btcusdt_spot_1m_01_2020_to_12_2025`, `AlpacaLiveData_LongOnly-v0`, `BTCUSD_sport_1m_12_2024_to_09_2025` (the `sport` typo), `Sebasdi/TorchTrade_btcusd_spot_1m_12_2024_to_09_2025`. Repointed at live equivalents.
3. **Offline IQL example unrunnable.** Now runs end to end.
4. **`GoogleNewsTool.run(as_of=...)`** silently ignored, making news-in-backtest a look-ahead leak. The parameter is now removed: no caller passes it, and dropping it yields `TypeError`, which the actor's per-tool guard already surfaces to the model as a tool error — the same loud failure without keeping a parameter nothing uses.

## Found while verifying #3

- Synthetic buffer emitted a generic `observation` key the model never reads (it is built from `observation_spec`) → now rolls out the env.
- Model lazy-init hardcoded a 4-timeframe/14-feature tensordict **no shipped config produces** → now a real reset observation. `observation_spec.rand()` is unusable: the specs are `Bounded` with infinite bounds, so sampling yields NaN. Root cause filed separately as #264 (17 sites across 9 files, offline and live).
- Eval reporting read a non-existent `base_env.action_history`.
- `iql_offline` smoke test re-enabled.

## Why this rotted

Two silent-green mechanisms, both fixed:

- Every dead-dataset test sat behind `skipif(not ..._available())`, so a 401 read as **skip**, not failure.
- `run_command` shelled out to literal `"python"` with `shell=True`, smoke-testing the examples against whatever interpreter was on `PATH` rather than the pinned `torchrl>=0.13.2`. Now `sys.executable`, and guarded by a test.

## Review findings addressed

Round 1 of the mandated review caught two things this PR had got wrong:

- **The normalisation loop crashed the HF branch.** `td.get(key)` returns `None` for a missing key, so any dataset without `next.terminated` — an ordinary shape — died with `AttributeError`. `terminated` is now synthesised from `done`. The fixture in this PR builds exactly such a dataset but stopped one call short of feeding it through, so nothing caught it.
- **The headline fix had no test that failed when reverted.** The smoke test only drives the env-rollout branch, where rewards are already `(n,1)`. Both gaps now have mutation-verified regression tests.

Also from review: wired the dead `cfg.logger.eval_steps` (eval was hardcoded to 10000 steps, burning 126s of a 300s smoke-test budget — now instant); routed `data_path` by whether it is an `org/name` repo id, so a relative on-disk buffer path is no longer sent to the Hub; dead-code sweep in the rewritten function; HF availability checked via `HfApi().dataset_info` instead of downloading 1.4M rows at collection time.

## Test coverage

Replaced 5 permanently-skipped HF tests (which gave `dataset_to_td` **zero** real coverage) with in-memory ones, plus the re-enabled smoke test.

| | before | after |
|---|---|---|
| suite excl. examples | 2184 passed | 2200 passed |
| `test_examples.py` collected | 15 | 16 |
| …actually executed | 10 (5 always skipped) | 16 (0 skipped) |

## Verification

```
Full suite (CUDA_VISIBLE_DEVICES=""):  2200 passed, 18 skipped, 0 failed
Offline IQL example end-to-end:        EXIT=0
```

All run against `.venv/bin/python` (torchrl 0.13.2, matching the pin) — not the conda interpreter on `PATH`, which carries torchrl 0.10.1.

## Not included (deliberate)

- `config.yaml` declares `trading_mode: "spot"`, but `env_maker` never reads it, so the env defaults to `action_levels=[-1,0,1]` with `leverage=1` — the short action is clipped to flat, the policy carries a dead action, and `eval/sell_actions` can only ever be zero. One-line fix, but it changes the action space and therefore what the example trains.
- `cfg.logger.eval_envs: 5` is also dead (`eval_num_envs=1` is hardcoded). Wiring it would 5x eval cost, so it is a separate call.
- The `Bounded(-inf, inf)` spec bug is #264 — it touches every environment and does not belong in an examples fix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,6 +46,9 @@ jobs:
           WANDB_MODE: disabled
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           HUGGING_FACE_HUB_TOKEN: ${{ secrets.HF_TOKEN }}
+          # Fail, don't skip, if the datasets the example smoke tests need are
+          # unreachable -- a silent skip is how dead dataset paths stayed green.
+          TORCHTRADE_REQUIRE_HF: "1"
         run: |
           pytest tests/ -v --cov=torchtrade/envs --cov-report=term-missing --cov-report=xml
 
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ from torchtrade.envs.offline import SequentialTradingEnv, SequentialTradingEnvCo
 import datasets
 
 # Load historical data from HuggingFace
-df = datasets.load_dataset("Torch-Trade/btcusdt_spot_1m_01_2020_to_12_2025")
+df = datasets.load_dataset("Torch-Trade/btcusdt_spot_1m_03_2023_to_12_2025")
 df = df["train"].to_pandas()
 df['0'] = pd.to_datetime(df['0'])
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -126,7 +126,7 @@ See the **[Examples](examples.md)** page for all available algorithms (PPO, DQN,
 import datasets
 
 # Load pre-processed Bitcoin data
-ds = datasets.load_dataset("Torch-Trade/btcusdt_spot_1m_01_2020_to_12_2025")
+ds = datasets.load_dataset("Torch-Trade/btcusdt_spot_1m_03_2023_to_12_2025")
 df = ds["train"].to_pandas()
 df['0'] = pd.to_datetime(df['0'])  # First column is timestamp
 df.columns = ['timestamp', 'open', 'high', 'low', 'close', 'volume']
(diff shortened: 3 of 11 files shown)


BY571:
@CharlieHelps review

Issue #61, opened 2026-01-09
[LiveEnv] Polymarket Integration


BY571:
https://docs.polymarket.com/quickstart/fetching-data

PR #265, opened 2026-07-29
fix(envs): use Unbounded for unbounded specs instead of Bounded(-inf, inf)

### LINKED ISSUE #264: Observation specs use Bounded(low=-inf, high=inf); spec.rand() returns NaN across every environment
## Summary

Every environment declares its `account_state` and `market_data` observation specs as `Bounded(low=-torch.inf, high=torch.inf, ...)`. `Bounded.rand()` computes `uniform_() * (high - low) + low`, which with infinite bounds is `0 * inf` / `inf + -inf` — **every sample is NaN**. The spec then rejects its own output.

Reproduced on the pinned torchrl 0.13.2:

```python
>>> from torchrl.data import Bounded, Unbounded
>>> import torch
>>> s = Bounded(low=-torch.inf, high=torch.inf, shape=(2, 3), dtype=torch.float)
>>> torch.isnan(s.rand()).all()
tensor(True)
>>> s.is_in(s.rand())        # spec rejects its own sample
tensor(False)

>>> u = Unbounded(shape=(2, 3), dtype=torch.float)
>>> torch.isfinite(u.rand()).all(), u.is_in(u.rand())
(tensor(True), tensor(True))
```

`Unbounded` is the correct spec for an unconstrained range — it samples from a real finite-variance distribution.

## Why it hasn't been caught

`check_env_specs()` passes today: it builds its dummy batch via `env.fake_tensordict()`, which uses `observation_spec.zero()`, and otherwise does a real `env.rollout()`. It never calls `.rand()`. So the bug is invisible to the standard spec check.

It bites anything using the ordinary TorchRL idiom of drawing a dummy batch from the spec to initialise lazy modules:

```python
example_td = env.observation_spec.rand()   # -> all NaN, silently
```

Lazy shape inference then initialises against NaN without raising, and the first forward pass produces NaN logits. This is how it surfaced: in #263 the offline IQL example's model lazy-init could not use `observation_spec.rand()` and had to fall back to `env.reset()`. That PR works around it locally; this issue is the underlying cause.

## Scope

17 occurrences across 9 files — offline *and* live, so this is the shared spec builders rather than one env:

- `torchtrade/envs/core/offline_base.py:142` (account_state), `:173` (market_data)
- `torchtrade/envs/core/base.py:56`
- `torchtrade/envs/core/live.py:234`
- `torchtrade/envs/offline/vectorized_sequential.py:204`, `:217`, `:227`
- `torchtrade/envs/live/alpaca/base.py:161`, `:175`
- `torchtrade/envs/live/binance/base.py:171`, `:184`
- `torchtrade/envs/live/bitget/base.py:188`, `:201`
- `torchtrade/envs/live/bybit/base.py:152`, `:165`
- `torchtrade/envs/live/okx/base.py:154`, `:167`

`torchtrade/envs/core/README.md:201` documents the same `low=-np.inf` pattern and should be updated with the fix.

## Suggested fix

Replace `Bounded(low=-torch.inf, high=torch.inf, shape=S, dtype=D)` with `Unbounded(shape=S, dtype=D)` at all 17 sites, per this repo's "fix the class, not the instance" convention — the shared builders in `core/` cover most environments, but the live `base.py` files and `vectorized_sequential.py` each carry their own copy.

Worth adding a contract test asserting `spec.is_in(spec.rand())` and `torch.isfinite(spec.rand()).all()` for every env's `observation_spec`, so a future `Bounded(-inf, inf)` cannot reappear — `check_env_specs()` will not catch it.

Note any spec with genuinely finite bounds should stay `Bounded`; this is only about the infinite-bound ones.

BY571:
Closes #264.

## The bug

TorchRL samples a `Bounded` spec as `uniform() * (high - low) + low`. An infinite bound makes `.rand()` produce **NaN** (both sides infinite) or **inf** (one side), and inf poisons a lazily-built network just as thoroughly:

```python
>>> Bounded(low=-torch.inf, high=torch.inf, shape=(2,3)).rand()   # all NaN
>>> Bounded(low=0.0, high=torch.inf, shape=(4,)).rand()           # all inf
>>> nn.Linear(4, 8)(that)   # tensor([inf, nan, nan, nan, nan, nan, nan, nan])
```

`check_env_specs()` cannot catch either — it builds its dummy batch from `spec.zero()` and a real rollout, never `.rand()`. So every environment shipped a spec that could not be sampled, and the standard `spec.rand()` lazy-init idiom silently initialised networks on garbage. That is how it surfaced, in the offline IQL example (#263).

`Unbounded` is the correct spec and a verified drop-in — identical on `zero`, `is_in`, `project`, `expand`, `to`, `dtype`, `Composite` interaction, pickling across `ParallelEnv` process boundaries, `squeeze`/`unsqueeze`, `type_check` and `encode`.

## Scope

**21 sites across 12 files** — the shared spec builders in `core/`, the vectorized env, all five live exchange base classes, a transform, and both polymarket specs.

The issue as filed undercounted twice: its list came from a `low=-torch.inf` grep (missing the `float("inf")` spelling in polymarket) and my first sweep was driven by that *file list* (missing `transforms/binance_ohlcv.py` entirely). The final sweep is pattern-driven across the whole package.

**Polymarket's `market_state` is converted too.** I initially argued to keep it as `Bounded(low=0.0, high=inf)` because the lower bound looked like a real constraint. Review showed it still poisons lazy init via `inf`.

The bound was consulted — `check_env_specs` calls `spec.contains(td)` on a real rollout, and `tests/envs/polymarket/test_env.py` exercises that path — but it never fired: every fixture market uses positive prices, spreads, volumes and liquidity, so the `low=0.0` boundary was never tested even before this PR. Converting trades a validation that structurally could have caught a negative-liquidity glitch, but never did, for a spec that can actually be sampled. The guard now has **zero exemptions**.

## A side effect worth knowing

`RewardScaling.transform_reward_spec` and `RewardClipping.transform_reward_spec` both `raise NotImplementedError` unless `reward_spec` is `Unbounded`. Before this PR every env's reward spec was `Bounded`, so wrapping *any* env in either transform was silently broken. This fixes that.

## Guards

**Structural** — an AST scan of every `.py` under `torchtrade/` and `examples/`, rejecting any `Bounded` with an infinite bound on either side. Source-level rather than a list of environments, because hand-sweeping escaped me twice. It catches keyword and positional args, aliased imports, qualified `module.Bounded`, and `torch.inf` / `np.inf` / `float("inf")` / `1e999`. Its docstring states what it does **not** resolve — `partial()`, subclassing, name rebinding, `**kwargs` — rather than claiming unescapability.

**Behavioural** — 18 tests across 15 envs plus the transform, asserting each spec produces a sample it would itself accept. Every one of the 21 converted sites fails a behavioural test if poisoned, verified by mutation (with the `Bounded` import restored, so the mutation measures the bound and not a `NameError`).

Live envs are discovered via `__subclasses__` and built with `__new__` + the unbound spec builder, so exchange #6 cannot skip the test by being forgotten — with a companion assertion that discovery still finds all five, since an empty parametrize skips rather than fails.

## Verification

```
Full suite (CUDA_VISIBLE_DEVICES=""):  2224 passed, 18 skipped, 0 failed
CI (3.11):                             pass
```

Run against `.venv/bin/python` (torchrl 0.13.2, matching the pin) — not the conda interpreter on `PATH`, which carries torchrl 0.10.1.

## Review

Three full rounds of code-simplifier, pr-test-analyzer, torchrl-engineer, and test-justification, with findings fixed between each. The rounds earned their keep: round 1 found my regex had caused collateral damage and the guard had five escapes; round 2 found that closing those six opened a seventh (`1e999`) and that poisoning bybit, the transform or polymarket failed **zero** behavioural tests; round 3 closed the last three uncovered sites, including `core/live.py`'s `base_features` — the one spec shared by all five exchanges, whose own docstring records that drift there already cost 3 of 4 exchanges in #61.

### DIFF
--- a/examples/offline_rl/iql/utils.py
+++ b/examples/offline_rl/iql/utils.py
@@ -290,8 +290,8 @@ def make_discrete_iql_model(cfg, env, device):
 
     # init nets
 
-    # Real reset obs, not observation_spec.rand(): the specs are Bounded with infinite
-    # bounds, so sampling them yields NaN.
+    # A real observation rather than observation_spec.rand(): lazy init on actual data
+    # keeps the first forward pass representative of what training will see.
     example_td = env.reset().to(device)
     with torch.no_grad(), set_exploration_type(ExplorationType.RANDOM):
         for net in model:
--- a/examples/offline_rl/iql/utils.py
+++ b/examples/offline_rl/iql/utils.py
@@ -290,8 +290,6 @@ def make_discrete_iql_model(cfg, env, device):
 
     # init nets
 
-    # A real observation rather than observation_spec.rand(): lazy init on actual data
-    # keeps the first forward pass representative of what training will see.
     example_td = env.reset().to(device)
     with torch.no_grad(), set_exploration_type(ExplorationType.RANDOM):
         for net in model:
--- a/tests/envs/test_spec_contracts.py
+++ b/tests/envs/test_spec_contracts.py
@@ -34,8 +34,8 @@ from torchtrade.envs.offline import (
     VectorizedSequentialTradingEnvSLTPConfig,
 )
 
-# examples/ too: the .rand() lazy-init idiom lives there, and that is where this bug
-# was found.
+# examples/ is scanned too: that is where this bug was found, and example code is where
+# the spec.rand() lazy-init idiom tends to reappear.
 REPO_ROOT = pathlib.Path(torchtrade.__file__).parent.parent
 SCAN_ROOTS = [REPO_ROOT / "torchtrade", REPO_ROOT / "examples"]
 
@@ -50,9 +50,9 @@ def _infinite(node):
     if isinstance(node, ast.Call) and isinstance(node.func, ast.Name) and node.func.id == "float":
         return bool(node.args) and _infinite(node.args[0])
     if isinstance(node, ast.Constant):
-        return isinstance(node.value, (int, float)) and math.isinf(node.value) or (
-            isinstance(node.value, str) and node.value.lstrip("-").lower().startswith("inf")
-        )
+        if isinstance(node.value, str):
+            return node.value.lstrip("-").lower().startswith("inf")
+        return isinstance(node.value, float) and math.isinf(node.value)
     return False
 
 
@@ -107,9 +107,9 @@ def test_no_spec_is_bounded_by_infinity():
 def _assert_specs_sample_finite(env, label):
     """A spec must produce a sample it would itself accept.
 
-    isfinite is the load-bearing assertion; is_in is near-tautological for Unbounded
-    and Categorical (it checks shape/dtype only) but would catch a shape or dtype
-    declaration that disagrees with what the spec generates.
+    isfinite is the load-bearing assertion. is_in is a strict subset of it -- it rejects
+    the NaN a two-sided-infinite Bounded produces, but accepts the inf a one-sided one
+    does -- kept as a second net on the NaN case, not as independent coverage.
     """
     for name in ("observation_spec", "reward_spec", "action_spec", "full_done_spec"):
         spec = getattr(env, name)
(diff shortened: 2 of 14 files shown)


BY571:
@CharlieHelps review

BY571:
@CharlieHelps review

PR #266, opened 2026-07-30
test: guard against HuggingFace dataset ids rotting

BY571:
Closes the last follow-up from #263.

## Why

Five HuggingFace dataset ids referenced from library code, configs, docs and tests were **404 for an unknown length of time**, and nothing caught it:

- the tests that touched them sat behind `skipif(not available())`, so a dead dataset read as **skip**, not failure
- the configs, docs and README were never checked at all

#263 fixed those five ids. This guards the general case, so a sixth cannot rot silently.

## What it does

Resolves every `Torch-Trade/*` and `Sebasdi/*` id appearing in git-tracked `.py`, `.yaml` and `.md` against the Hub. Metadata-only via `HfApi().dataset_info`, so it costs **one request per distinct id** — currently 2, across 31 references — not a download.

## Verification

Mutation-tested against the three ids that actually died:

```
reintroduce Torch-Trade/AlpacaLiveData_LongOnly-v0            -> 1 failed
reintroduce Torch-Trade/BTCUSD_sport_1m_12_2024_to_09_2025    -> 1 failed
reintroduce Sebasdi/TorchTrade_btcusd_spot_1m_12_2024_...     -> 1 failed
(control) unmodified                                           -> 1 passed
```

```
Full suite: 2226 passed, 18 skipped, 0 failed
```

## Two deliberate design points

**It scans git-tracked files only.** Hydra `outputs/` dirs and gitignored scratch both carry stale ids that are history rather than bugs — the first draft walked the tree and tripped on an elided `btcusdt_spot_1m_...` in an ignored planning note. `git ls-files` excludes both by construction.

**Skipping when the Hub is unreachable is the exact mechanism that hid the original rot**, so the resolve test honours `TORCHTRADE_REQUIRE_HF=1` — which CI already sets, from #263. Locally without network it skips; in CI it fails.

The extraction is additionally asserted **offline**, so a scan that silently stopped matching cannot pass by finding nothing — which is the failure mode a network-gated test of this shape would otherwise have.

## Stated boundary

The pattern matches only this project's two publishing orgs. Matching every `a/b` string would drown in false positives, so a dataset published elsewhere is not covered. That is a known limit, not an oversight.

### DIFF
--- a/tests/test_dataset_ids.py
+++ b/tests/test_dataset_ids.py
@@ -10,11 +10,13 @@ What this checks, and what it does not:
 * It resolves each id against the Hub with `HfApi().dataset_info` -- metadata only,
   one request per distinct id, no download. Existence is all that proves. A repo can
   resolve here and still break `datasets.load_dataset(...)`: no `train` split (which
-  `load_torch_trade_dataset` hardcodes), a required `config_name`, or a schema the
+  `load_torch_trade_dataset` defaults to), a required `config_name`, or a schema the
   offline samplers cannot read.
-* Ids assembled at runtime -- f-strings, concatenation -- are invisible to a text
-  scan, as are file types outside {.py, .yaml, .yml, .md} and anything not yet
-  git-tracked.
+* Ids assembled at runtime are not resolved. A literal prefix before an interpolation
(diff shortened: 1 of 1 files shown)


BY571:
@CharlieHelps review

Issue #280, opened 2026-08-07
SL/TP fills at the exact bracket price, ignoring gaps


BY571:
\`sequential_sltp.py:384-388\` and \`replay/order_executor.py:97-100\` set \`execution_price = self.stop_loss\` (or take_profit) when triggered. If the bar **gaps past** the bracket (long entry 100, SL 97.5, next bar opens at 85), the fill is booked at 97.5 — a real stop-market fills near 85. Overstates PnL on exactly the moves that hurt most.

Enhancement: fill at \`min(open, stop)\` for longs / \`max(open, stop)\` for shorts when the bar opens beyond the bracket. Applies to scalar SLTP, vectorized SLTP, OneStep, and replay — all four share the trigger rule, so this should land as one change in a shared helper (see also the three-divergent-copies note in the audit: \`sequential_sltp.py:213-294\`, \`onestep.py:429-482\`, \`vectorized_sequential_sltp.py:259-282\`).

BY571:
Surfaced again during review of #267 (execution-bar aggregation), with a measured end-to-end case that is worth recording here.

15Min `execute_on`, long at 100 with SL at 98, a data gap over the 03:00 bar, market reopens at 90:
- 03:00 bar: `{open:100, high:100, low:100, close:100, volume:0}` — the gap-bar collapse from #267 behaving correctly
- 03:15 bar: `90/90/90/90` → exit booked at **98**, portfolio **9800**
- first tradable price after the gap was **90** → true portfolio **9000**

8% of capital of silent optimism, and it contradicts the module's own stated bias (`sequential_sltp.py:219-222`: *"we assume the worst case (SL). This pessimistic bias ensures backtesting underestimates performance"*).

Pre-existing and deliberately **not** fixed in the #267 PR — that PR is a one-line aggregation fix and this changes fill semantics. But #267 turns gap bars into a designed concept with a documented rationale, so this fill now sits directly downstream of a deliberate decision rather than of an accident.

Related note for whoever picks this up: the `open_price` branches in `_check_sltp_trigger` (`sequential_sltp.py:246-251`, `onestep.py:453-478`) are unreachable — after #267, `low = min` over sub-bars means `low <= open` structurally, so the low/high checks always fire first and both branches return the same value. A gap-aware fill has to change the execution price, not add another trigger branch.

PR #291, opened 2026-08-07
fix(envs): aggregate execution-bar OHLC instead of taking the last sub-bar

### LINKED ISSUE #267: Offline sampler: execution-bar OHLC built with .last() — stop/liquidation range ~12x too narrow
## Bug

\`torchtrade/envs/offline/infrastructure/sampler.py:149\`:

\`\`\`python
execute_base_raw = self.df.resample(execute_on.to_pandas_freq()).last()
\`\`\`

\`Resampler.last()\` takes the last valid value **per column**, i.e. the final sub-bar's OHLCV — not the aggregated bar. The correct aggregation dict already exists at \`sampler.py:88-89\` (\`ohlcv_agg\`/\`full_agg\`) and is used for \`resampled_dfs\`; only \`execute_base_raw\` skips it. \`close\` is coincidentally correct; \`open/high/low/volume\` are wrong.

## Measured impact

1-min source, \`execute_on="1Hour"\`: mean intrabar range 0.065% (env) vs 0.767% (true) — **11.8x too narrow**. On a 30-day BTC-like series a 0.5% stop that should fire 187 times fires **0** times.

Blast radius (every consumer of \`get_base_features\`/\`execute_base_tensor\`): \`sequential.py\` \`_check_liquidation\`, \`sequential_sltp.py\` \`_check_sltp_trigger\`, \`onestep.py\`, \`vectorized_sequential(_sltp).py\`, \`replay/observer.py\`. Observation windows are unaffected (they use \`resampled_dfs\`). Net effect: systematic optimism — stops and liquidations that fire in reality never fire in backtest.

Invisible to the suite because every SLTP/replay test uses \`execute_on\` == source frequency, where \`.last()\` is coincidentally correct.

## Fix

Build \`execute_base_raw\` with the same \`full_agg\` dict. Add a regression test with \`execute_on\` coarser than the source frequency asserting high/low match the true aggregate.

BY571:
Closes #267.

## The bug

`sampler.py` built the execution timeframe's bars with `resample(...).last()`, which takes the last valid value **per column** — the final sub-bar's OHLCV, not the bar's aggregate. `close` is coincidentally correct; `open`, `high`, `low` and `volume` are not.

Every SL/TP and liquidation check reads this bar's high/low range, so the range was systematically too narrow. Measured on 30 days of synthetic 1-minute data resampled to 1 hour:

| | buggy (`.last()`) | correct |
|---|---|---|
| mean intrabar range | 0.139% | 0.688% (**5x wider**) |
| 0.5% stop-loss hits detected | 73 | 155 |

So roughly **half of all stop-hits were invisible**, and the error is one-directional: stops and liquidations that trigger in reality never triggered in backtest. The exact ratio depends on the volatility of the series — the mechanism, not the multiplier, is the point.

The correct aggregation dict already existed and was used for the observation windows; only the execution bar skipped it. Observations were never affected, so this is not a leakage fix.

**Why it survived:** every SLTP and replay test uses `execute_on` equal to the source frequency, where `resample` is an identity op and `.last()` is coincidentally correct.

## The fix

Three changes in `sampler.py`, all in the same block:

1. Aggregate with `ohlcv_agg` (`open:first, high:max, low:min, close:last, volume:sum`).
2. Exclude aux columns from the execution bar. Nothing ever read them from this frame (aux reaches observations via `resampled_dfs`), so this makes the `row[:5]` positional contract exact rather than conventional, and deletes a now-dead `fillna` block.
3. Collapse bars with no source rows to flat. Empty bins are forward-filled, so such a bar used to inherit the previous bar's high/low — and widening that range meant it could re-trigger SL/TP on the excursion the position already lived through, stopping it out twice on one move. A bar with no trades has no range of its own.

Point 3 keys on a genuinely empty bin, not on `nan_mask` (which means "no usable close"). A bar that traded but lacks a usable close still has a real high/low, and flattening it would re-create the too-narrow range this PR exists to fix.

The vectorized envs read `execute_base_tensor` directly, so they inherit the fix with no changes.

## Tests

Each was verified by mutation — revert the fix, watch the specific test fail, with the mutation itself asserted to have applied.

| Test | Covers |
|---|---|
| `test_execution_bar_aggregates_its_sub_bars` | OHLCV vs an independently recomputed aggregate at coarse `execute_on` |
| `test_gap_bar_is_flat_rather_than_a_replay_of_the_previous_range` | the collapse |
| `test_bar_without_a_usable_close_keeps_its_real_range` | the collapse does *not* over-reach |
| `test_exit_fires_on_a_mid_bar_wick_when_execute_on_is_coarse` | end-to-end: stop-loss at 1x and liquidation at 10x, exit type and balance both pinned |
| `test_sl_fires_on_a_mid_bar_wick_when_execute_on_is_coarse` | vectorized SLTP — an independent consumer that slices the tensor by hardcoded column index |

The wick sits mid-bar, not on the final sub-bar, so it is visible only once the bar aggregates. Two tests whose assertions were subsumed were deleted; both were pinned at the degenerate config, which is why neither could ever have caught this.

Full suite: **2246 passed, 18 skipped.**

## Notes for the reviewer

- **This PR does three things**, not one. Points 2 and 3 above are not what #267 asked for — they are consequences the fix exposed (3 is a regression the fix would otherwise introduce; 2 removes a wart it uncovered). Happy to split if you'd rather review them separately.
- **Gap-through fill price** is a separate, pre-existing problem: an exit still books at the bracket price even when the market gaps past it, measured at 8% of capital in one case. Deferred to #280 with the evidence recorded there rather than expanded into this PR.
- The `open_price` branches in `_check_sltp_trigger` are unreachable — `low = min` over sub-bars means the low/high checks always fire first. Noted on #280, since a gap-aware fill has to change the execution price rather than add a trigger branch.
- The OHLCV column reorder in `__init__` is no longer what enforces `row[:5]` (pandas orders `.agg(dict)` output by dict key). Kept deliberately as a canonical order for the public `self.df`, with its comment corrected to stop claiming a guarantee it no longer provides.

### DIFF
--- a/tests/envs/offline/test_sampler.py
+++ b/tests/envs/offline/test_sampler.py
@@ -433,6 +433,34 @@ class TestSamplerBaseFeatures:
             assert source["high"].max() > source["high"].iloc[-1]
             assert source["low"].min() < source["low"].iloc[-1]
 
+    def test_gap_bar_is_flat_rather_than_a_replay_of_the_previous_range(self, sample_ohlcv_df):
+        """A bar with no source rows must not inherit the previous bar's high/low.
+
+        Forward-filling the range would hand SL/TP the excursion the position already
+        lived through one bar earlier, letting a single move stop it out twice.
+        """
+        execute_on = TimeFrame(15, TimeFrameUnit.Minute)
+        gap_start, gap_end = pd.Timestamp("2024-01-01 03:00"), pd.Timestamp("2024-01-01 03:15")
+        df = sample_ohlcv_df[
+            (sample_ohlcv_df["timestamp"] < gap_start) | (sample_ohlcv_df["timestamp"] >= gap_end)
+        ]
+
+        with pytest.warns(UserWarning, match="DATA GAPS"):
+            sampler = MarketDataObservationSampler(
+                df=df, time_frames=[execute_on], window_sizes=[5], execute_on=execute_on,
+            )
+
+        gap = sampler.get_base_features(gap_start)
+        previous = sampler.get_base_features(gap_start - pd.Timedelta("15min"))
+
+        assert gap["close"] == pytest.approx(previous["close"])
+        assert gap["open"] == pytest.approx(gap["close"])
+        assert gap["high"] == pytest.approx(gap["close"])
+        assert gap["low"] == pytest.approx(gap["close"])
+        assert gap["volume"] == 0.0
+        # Only meaningful if the bar it copies from actually had a range to inherit.
+        assert previous["high"] > previous["low"]
+
 
 class TestSamplerHelperMethods:
     """Tests for helper methods."""
--- a/tests/envs/offline/test_sampler.py
+++ b/tests/envs/offline/test_sampler.py
@@ -395,17 +395,14 @@ class TestSamplerObservations:
 class TestSamplerBaseFeatures:
     """Tests for base feature retrieval."""
 
-    @pytest.mark.parametrize("execute_on,sub_bars", [
-        (TimeFrame(1, TimeFrameUnit.Minute), 1),
-        (TimeFrame(1, TimeFrameUnit.Hour), 60),
-    ], ids=["source-frequency", "coarser"])
-    def test_execution_bar_aggregates_its_sub_bars(self, sample_ohlcv_df, execute_on, sub_bars):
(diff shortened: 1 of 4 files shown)


BY571:
@CharlieHelps review

Issue #293, opened 2026-08-07
Scalar vs vectorized SLTP diverge at high leverage for notional/quantity trade modes


BY571:
## Bug

Randomised-action equivalence sweeps between `SequentialTradingEnvSLTP` and `VectorizedSequentialTradingEnvSLTP` show mismatches at `leverage=10` across `fractional`, `notional` and `quantity` trade modes — both float drift in portfolio value/reward and, in some runs, outright **trigger-timing divergence** (one env exits a bar before the other).

At `leverage=1` every combination of trade mode, `lock_position_until_sltp` and fee shows zero mismatches. The divergence is leverage-dependent.

## Not a regression

Confirmed pre-existing: the identical fixed-seed sweep run against a `main` worktree reproduces the same mismatch counts, step numbers and magnitudes byte for byte. Found while reviewing #268, which is unrelated to it.

## Why it matters beyond float noise

Pure float32-vs-float64 drift would be a rounding annoyance. A *binary* trigger-timing difference is not: it means the two envs disagree about whether a stop fired on a given bar, so a policy evaluated on the vectorized env and deployed against scalar-env assumptions (or vice versa) sees different exits. The vectorized env's `docstring` claims it "passes extensive equivalence tests", which is only true for the axes currently covered.

## Coverage gap that hides it

`tests/envs/offline/test_vec_sltp_scalar_equivalence.py` never varies:
- `lock_position_until_sltp` (always default)
- `trade_mode` (always the default `fractional`)

Both were exercised only by ad-hoc sweeps during review. Adding them to the permanent harness is the first step, and will likely turn this from "found by chance" into a reproducible failing test.

## Suggested direction

1. Extend the equivalence harness over `trade_mode` and `lock_position_until_sltp`.
2. Decide whether the vectorized env should compute money in float64. It stores state as `(num_envs,)` float32 tensors; leverage multiplies the absolute magnitudes, so the same relative epsilon lands near bracket boundaries and flips comparisons.
3. Until resolved, treat the "extensive equivalence tests" claim in `vectorized_sequential.py` as scoped to spot/fractional.

Issue #292, opened 2026-08-07
SLTP envs: a trigger on bar N+1 cancels the agent's action from bar N's close


BY571:
## Bug

In `sequential_sltp.py::_step` the priority chain is an `if/elif`:

```python
elif self.position.position_size != 0:
    sltp_trigger = self._check_sltp_trigger(base_features)   # bar N+1
    if sltp_trigger is not None:
        trade_info = self._execute_sltp_close(...)           # agent action never runs
    elif self.config.lock_position_until_sltp:
        ...
    else:
        trade_info = self._execute_sltp_action(...)          # only reached if no trigger
```

So when a held position's SL/TP triggers on bar N+1, the agent's action is **discarded entirely** — not merely unrecorded. Same structure in `vectorized_sequential_sltp.py` via `active = ~triggered_mask` in `_execute_sltp_trades`.

This is backwards chronologically. The agent observes bar N and its order executes at **close(N)**. Bar N+1 unfolds *afterwards*. A position the agent already decided to close or flip at close(N) should not then be stopped out during a bar it should no longer have been holding.

## Reproduced

Futures, leverage 2, `stoploss_levels=[-0.025, -0.50]`, `takeprofit_levels=[0.025, 0.50]`, flat 100.0 series with `high=120` on bar 20.

Open `("long", -0.025, 0.025)` (TP at 102.5) at bar 15, then at bar 19 the agent selects a short:

- **Actual:** the long's TP fires on bar 20 → balance 10500, account **flat**, `action_types[-1] == "sltp_tp"`. The requested short never opened.
- **Expected:** the long closes at close(19)=100 (flat P&L), the short opens at 100, and *then* bar 20's range applies to the short.

The agent ends the step in a different position than it asked for, at a price it did not choose, and must re-issue the order next step.

## Cost

The exit is booked at the bracket price rather than the market close the agent selected, and the intended position is silently missing for at least one bar. Under `lock_position_until_sltp` the interaction is different (the action is meant to be ignored while a position is open), so any fix must preserve that.

## Fix direction

Execute the agent's action at close(N) first, then run the bar N+1 exit checks against the resulting position — which is the ordering #268 established for freshly-opened positions. Doing it for the incoming position too makes the whole step chronological. Not bundled into #268 because it changes which trades happen, not just when they are checked, and needs its own review.

Found while verifying the #268 fix.

PR #294, opened 2026-08-07
fix(envs): check SLTP exits on the bar a position is opened into

### LINKED ISSUE #268: SequentialTradingEnvSLTP (+vectorized): SL/TP/liquidation never checked on a position's entry bar
## Bug

\`torchtrade/envs/offline/sequential_sltp.py:371-398\` (identical structure in \`vectorized_sequential_sltp.py\`): \`_step()\` advances the sampler to bar N+1, checks liquidation/SL/TP against bar N+1, and only **then** opens the new position at bar N's close. A position opened at close(N) is therefore first checked at bar **N+2** — bar N+1's range is consumed by a check that ran while flat.

## Reproduced

Flat 100.0 series, crash bar (low=80) at index 20, long entry at 100 with SL -2.5%:
- open at bar 19 (crash is the next bar): **stop never fires**, position rides through low=80 untouched
- open at bar 18: fires correctly on bar 20, balance 9750

\`OneStepTradingEnv\` on the same data and entry bar handles it correctly (balance 9750, reward -0.0253) — the two SLTP implementations disagree by a full stop-loss on identical input. This also breaks the \`lock_position_until_sltp\` "OneStep policy evaluation parity" claim: under locking, the exit check is exactly the one that is off by one.

Base \`SequentialTradingEnv._step()\` (\`sequential.py:527\`) gets the ordering right (checks the cached current bar before advancing).

## Fix

Move the trigger check so a freshly opened position is evaluated against the first bar it lives through, matching OneStep. Apply to both scalar and vectorized SLTP. Regression test: entry immediately before a crash bar must trigger the stop.

BY571:
Closes #268.

## The bug

Both SLTP envs advanced the sampler and ran the liquidation and SL/TP checks against bar N+1 **before** opening the position at bar N's close. Those checks therefore only ever saw the position held *coming into* the step. A position opened at close(N) was first checked on bar N+2, and bar N+1 — the first bar it is actually exposed to — was never applied to it.

Reproduced on a flat series with one crash bar (low 80, stop at 97.5):

| Entry | Result |
|---|---|
| Two bars before the crash | Stop fires, balance **9750** |
| **One** bar before the crash | Position **rides straight through**, balance 0, still holding |

So the two paths disagreed by a full stop-loss on identical data. `OneStepTradingEnv` already handled this correctly, which is why the fix also restores the parity `lock_position_until_sltp` exists for.

The same blind spot applied to **liquidation** — a freshly opened levered position could ride through a bar that gapped past its liquidation price.

## The fix

Both envs re-run the checks against the bar **already fetched this step**, scoped to positions opened during it.

- No lookahead: the action was chosen on bar N's close alone, and the bar is not re-fetched.
- No retroactive stop: bar N's own range still cannot close a position that exists only from its close. Verified explicitly in review.
- The vectorized env had the checks inline, so extending them would have forked a second copy. They moved into `_apply_exit_checks`, called once for incoming positions and once for newly opened ones.

## Tests

Every test below was verified by mutation, with the mutation asserted to have applied.

| Test | Catches |
|---|---|
| `test_exit_fires_on_the_first_bar_the_position_is_exposed_to` (scalar, 10 cases) | the shared bug, exit-before-observation ordering, exact fills and fees |
| `TestSLTPScalarVecEquivalenceEntryBar` (+ switch case) | scalar/vectorized divergence, take-profit, direction switches |
| `test_entry_bar_check_targets_the_env_that_opened` | per-env mask misalignment |

The division of labour matters: the equivalence suite passed on `main` because the bug was **symmetric** — both envs skipped the entry bar identically. Value-pinning tests catch the shared bug; the equivalence suite catches drift; the `num_envs=2` test catches misalignment.

Three gaps found during review that the first version of this PR did not cover, each now closed and each verified to have been invisible before:

- **Exit-before-observation ordering.** Moving the new block after the observation build left the internal balance correct, so all 56 scalar tests passed — while the policy was handed a 10x-levered open long, 9.6% from liquidation, on a step the env had already liquidated. The test now asserts the returned `account_state`, `reward` and `terminated`.
- **`opened_mask` across envs.** Aiming the check at a sibling env passed all 658 offline tests, because every test touching it had all envs doing the same thing.
- **Fees and the take-profit fill.** A double-charged exit fee and a 1%-wrong take-profit fill both passed the suite; exact values now pin them.

Full suite: **2261 passed, 18 skipped.**

## Notes for the reviewer

- **A same-bar round trip records only the exit.** `HistoryTracker` has one row per step and the exit determines the ending state, matching what already happens when a previously-held position triggers — but it is now reachable one bar earlier. `action_types` is consumed only by the plot markers in `_get_action_markers`, so nothing numeric under-counts; the futures branch infers exit direction from `position_history[i-1]`, which is now `0` for these, so such an exit draws no marker. Pinned in the test so it stays a decision.
- **`eligible=opened_mask` is inert today** — dropping it passes all 658 offline tests, since re-checking a first-pass survivor on the same bar gives the same answer. Kept deliberately: that holds only while the check stays idempotent, and nothing enforces that. Noted at the call site.
- **#292 filed**, found while verifying this: when a held position triggers on bar N+1 the agent's action from close(N) is discarded outright, so a requested long→short flip can leave the account flat instead. Same ordering family, but it changes *which trades happen*, so it needs its own review.
- **#293 filed**: scalar and vectorized diverge at `leverage=10` for non-fractional trade modes, including trigger-timing disagreement. Verified byte-identical on `main`, so it predates this work; the equivalence harness never varies `trade_mode` or `lock_position_until_sltp`.

### DIFF
--- a/tests/envs/offline/test_sequential_sltp.py
+++ b/tests/envs/offline/test_sequential_sltp.py
@@ -397,17 +397,21 @@ class TestSLTPRegression:
 
     @pytest.mark.parametrize("bars_after_entry", [1, 2], ids=["next-bar", "two-bars-later"])
     @pytest.mark.parametrize(
-        "leverage,stoploss_pct,wick_low,fee,expected_exit,expected_balance,expected_terminated", [
+        "leverage,sl_pct,tp_pct,wick_low,wick_high,fee,"
+        "expected_exit,expected_balance,expected_terminated", [
             # The fee rows pin exact values because the next-bar path is the only place
             # an entry fee and an exit fee land in one step; "cheaper than free" would
             # hold just as well if one were charged twice.
-            (1, -0.025, 80.0, 0.0, "sltp_sl", 9750.0, False),
-            (1, -0.025, 80.0, 0.001, "sltp_sl", 9730.519481, False),
-            (10, -0.50, 88.0, 0.0, "liquidation", 400.0, True),
-            (10, -0.50, 88.0, 0.001, "liquidation", 306.534653, True),
-        ], ids=["stop-loss", "stop-loss-fee", "liquidation", "liquidation-fee"])
+            (1, -0.025, 0.50, 80.0, None, 0.0, "sltp_sl", 9750.0, False),
+            (1, -0.025, 0.50, 80.0, None, 0.001, "sltp_sl", 9730.519481, False),
+            (10, -0.50, 0.50, 88.0, None, 0.0, "liquidation", 400.0, True),
+            (10, -0.50, 0.50, 88.0, None, 0.001, "liquidation", 306.534653, True),
+            # The fix repeats the trigger->fill-price mapping, so pin the take-profit
+            # side too: a wrong fill here is invisible to every other assertion.
+            (1, -0.50, 0.025, None, 120.0, 0.0, "sltp_tp", 10250.0, False),
+        ], ids=["stop-loss", "stop-loss-fee", "liquidation", "liquidation-fee", "take-profit"])
     def test_exit_fires_on_the_first_bar_the_position_is_exposed_to(
-        self, bars_after_entry, leverage, stoploss_pct, wick_low, fee,
+        self, bars_after_entry, leverage, sl_pct, tp_pct, wick_low, wick_high, fee,
         expected_exit, expected_balance, expected_terminated
     ):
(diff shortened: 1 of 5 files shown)


BY571:
@CharlieHelps review

Issue #295, opened 2026-08-07
Live envs: surface exchange-status staleness to the policy and truncate on a prolonged outage


BY571:
Follow-up to #270, which makes an unknown exchange status stop reading as a flat account. This covers what the policy *sees* during an outage, deliberately split out because it changes the observation contract.

## Why separate

#270 is pure safety: no observation-spec change, no effect on trained checkpoints. Everything below alters what the env emits, so it invalidates policies trained before it and needs its own review.

## 1. The observation cannot say "I don't know"

`account_state` is a fixed 6-element vector shared with the offline envs. During an outage the honest answer is neither "flat" nor a confirmed open position. #270 falls back to the last known cached state, which is the best available truth but is silently stale.

There is no room for a staleness flag inside the 6 elements without breaking offline/live parity — the offline envs have no analogous concept, so a 7th slot would desync them. The right shape is a separate additive observation key (e.g. `status_unknown`) alongside `account_state`.

## 2. A prolonged outage should truncate, not terminate

All five live envs currently hardcode `truncated=torch.tensor([False])`, so the channel is dead weight. A long outage should end the episode via `truncated=True`, never `terminated=True`: value estimators read `terminated` as "true return-to-go is 0" and `truncated` as "bootstrap from the final observation". Terminating would teach the critic that an unreachable API and a blown account carry the same value target.

Suggested shape: a consecutive-unknown counter mirroring the existing `hold_counter` in `PositionState`, with a config threshold mirroring `bankrupt_threshold`/`done_on_bankruptcy` in `core/live.py`.

## 3. Unverified, needs checking first

Whether `SyncDataCollector`'s `auto_reset` handles `truncated`-without-`terminated` correctly for these stateful, non-vectorized live envs. Confirm against how `examples/` actually drives them before relying on the semantics.

BY571:
## Part 3 checked — and the premise needs updating before parts 1 and 2 are designed

The issue asks whether `SyncDataCollector`'s `auto_reset` handles `truncated`-without-`terminated` for these envs. Verified against the pinned torchrl (**0.13.2**), and three things are off:

**1. There is no `auto_reset` parameter.** `Collector.__init__` exposes `reset_at_each_iter`, `reset_when_done` and `set_truncated` — no `auto_reset`. So the specific mechanism the issue is worried about does not exist in the version this repo pins.

[...]

Issue #272, opened 2026-08-07
All 10 live-exchange envs write 'truncated' but omit it from full_done_spec — check_env_specs fails


BY571:
## Bug

None of \`torchtrade/envs/live/{alpaca,binance,bitget,bybit,okx}/{env,env_sltp}.py\` declares \`full_done_spec\`, so they inherit TorchRL's default (\`done\` + \`terminated\` only) — yet every \`_step()\` writes a \`"truncated"\` key (e.g. \`alpaca/env.py:154\`). Reproduced with mocked clients:

\`\`\`
check_env_specs(env) → AssertionError: keys_in_real_not_in_fake={('next', 'truncated')}
\`\`\`

The offline base (\`core/offline_base.py:171-176\`) and polymarket declare it correctly. The codebase's own comment at \`core/live.py:227-230\` warns about exactly this failure mode: a collector pre-allocating from the spec silently drops the key.

Why it shipped: **no live-exchange test calls \`check_env_specs\`** — coverage exists only for offline envs, transforms and polymarket.

## Fix

Declare the 3-key \`full_done_spec\` once in the shared live base (\`TorchTradeLiveEnv\`), and add spec-contract tests for the live envs (mocked clients), ideally via the existing \`__subclasses__\` discovery pattern in \`tests/envs/test_live_env_base.py\` so future envs are covered automatically.

PR #296, opened 2026-08-08
fix(envs): stop an unreachable exchange from reading as a flat account

### LINKED ISSUE #270: Live get_status(): API failure encoded as flat account — enables position doubling during outages
## Bug

All five exchanges catch \`Exception\` in \`get_status()\` and set \`position_status = None\`:
\`binance/order_executor.py:348-350\`, \`bitget/order_executor.py:586-588\`, \`bybit/order_executor.py:358-360\` (+319-322), \`okx/order_executor.py:378-380\` (+329-332), \`alpaca/order_executor.py:306-308\`.

\`None\` is the **only** encoding of flat (\`position_direction_from_status(None) == 0\`), so a timeout/429/non-zero retCode is indistinguishable from an empty account.

## Traced failure (binance, 20x, holding 0.5 BTC long)

1. \`futures_position_information\` times out → \`position_status = None\`
2. sync (\`binance/env.py:162\`): observed 0 ≠ cached +1 → resets \`current_action_level = 0.0\`
3. agent repeats action 1.0 → duplicate guard compares 1.0 == 0.0 → does not short-circuit
4. \`_get_current_position_quantity()\` re-queries, outage persists → 0.0
5. delta = 0.5 → **BUY 0.5 BTC → 1.0 BTC at 20x**, repeated every bar of the outage

The dust rule itself is honoured — it is being fed a lie. Fix belongs at the boundary: raise, or return an UNKNOWN sentinel the env refuses to trade on (skip step, keep cache).

BY571:
Closes #270.

## The bug

All five live exchanges collapsed a failed position fetch to `position_status = None` — which is also how a **confirmed-flat account** is encoded. An API timeout was therefore indistinguishable from an empty account.

Traced on binance at 20x holding 0.5 BTC long:

1. `futures_position_information` times out → `position_status = None`
2. the sync sees "flat", so it resets `current_action_level`
3. the agent repeats its action → the duplicate-action guard no longer matches
4. the quantity re-query also reads 0 mid-outage
5. delta = 0.5 → **buys another 0.5 BTC** — and repeats every bar until the outage ends

The strongest evidence this was an oversight rather than a design choice: `get_account_balance()` and `get_mark_price()` in those same classes already **raise** on failure. Only the position path was fail-open.

## What the blast-radius map found beyond the issue

- **It isn't only the `except` blocks.** Bybit and OKX also mapped a non-zero API return code to `None` without raising at all, and OKX did the same for an account holding positions it cannot represent. Two existing tests asserted that as correct.
- **There are two `_sync_position_from_exchange` forks.** The SLTP one returns "was the position closed?", so an outage looked like an SL/TP fill and would clear brackets the exchange still holds.
- **Alpaca's `_get_portfolio_value`** dropped a held position's market value on the flat branch, feeding cash alone to the bankruptcy check — an outage could terminate the episode as bankrupt.

## The fix

`get_status()` returns `POSITION_UNKNOWN` on failure, distinct from `None`. It is truthy (so `if position_status:` cannot take the flat branch), `position_direction_from_status` raises on it instead of returning 0, and reading any field off it raises with the reason — which is what makes the trade-sizing paths fail closed without editing four call sites.

Alpaca is the awkward one: **its API signals a flat account by raising**, so flat and failure share one channel. Only error code `40410000` means flat; a 429, a 5xx or an unparseable body counts as unreachable. Its mock raised a bare `Exception` and so could not express that distinction — it now raises `APIError` like the real client.

## Behaviour change

An outage now **raises out of `_step`/`_reset`** rather than trading on a phantom flat account. Verified: nothing in `examples/` or TorchRL's collector catches it, so it stops the driver rather than being swallowed. Riding an outage out instead — a stale-but-marked observation, truncating on a prolonged outage — is **#295**, which has to stop `_get_observation` raising before any of it works.

## Tests

Mutation-driven across three review rounds: the final sweep ran 46 mutants with 43 killed, and every remaining survivor is provably equivalent.

| Test | Covers |
|---|---|
| `test_an_outage_stops_the_step_before_it_can_trade` | the traced scenario end to end, 5 exchanges × `env`/`env_sltp`, asserting **no order placed** before asserting it failed closed |
| `test_a_failed_position_fetch_does_not_report_flat` | all five real executors with a failing client |
| `test_alpaca_only_treats_its_not_found_error_as_flat` | 40410000 vs 429 / 5xx / unparseable |
| `test_neither_sync_fork_treats_an_outage_as_flat` | the contract #295 will edit |
| plus alpaca's own observation, portfolio value and outer-handler paths, okx's multi-position branch, and the sentinel's identity/truthiness contracts |

## Notes for the reviewer

- **Polymarket is out of scope by construction**, not oversight: it subclasses `EnvBase` directly rather than `TorchTradeLiveEnv`, and its executor has no `position_status` concept. It is invisible to the discovery guard for that reason.
- **`check_env_specs` fails on these envs** with a `('next','truncated')` mismatch — pre-existing on `main` and tracked as **#272**, unrelated to this branch.
- After **#295** softens the sync forks, alpaca's non-SLTP `_step` will be stopped solely by its price read, which nothing currently pins. Worth carrying a test there when #295 lands.

### DIFF
--- a/tests/envs/bybit/test_futures_order_executor.py
+++ b/tests/envs/bybit/test_futures_order_executor.py
@@ -2,6 +2,7 @@
 
 import pytest
 from unittest.mock import MagicMock
+from torchtrade.envs.core.state import POSITION_UNKNOWN
 
 
 class TestBybitFuturesOrderClass:
@@ -586,11 +587,15 @@ class TestBybitFuturesOrderClass:
         assert result is expected
 
     def test_get_status_validates_retcode(self, order_executor, mock_pybit_client):
-        """get_status must return position_status=None on non-zero retCode."""
+        """A non-zero retCode is a failed call, not an empty account (#270).
+
+        Reporting None here would let the env trade as though it were flat while a
+        real position sits on the exchange.
+        """
         mock_pybit_client.get_positions = MagicMock(return_value={
             "retCode": 10001,
             "retMsg": "Invalid parameter",
             "result": {"list": []},
         })
         status = order_executor.get_status()
-        assert status["position_status"] is None
+        assert status["position_status"] is POSITION_UNKNOWN
--- a/tests/envs/okx/test_futures_order_executor.py
+++ b/tests/envs/okx/test_futures_order_executor.py
@@ -235,6 +235,19 @@ class TestOKXFuturesOrderClass:
         })
         assert order_executor.get_status()["position_status"] is POSITION_UNKNOWN
 
+    def test_multiple_open_positions_is_not_flat(self, order_executor, mock_okx_account_client):
+        """LONG_SHORT mode is unsupported here, which is not the same as holding nothing.
+
+        Reading two positions the env cannot represent as "flat" means it opens a third.
+        """
+        mock_okx_account_client.get_positions = MagicMock(return_value={
+            "code": "0", "msg": "", "data": [
+                {"instId": "BTC-USDT-SWAP", "pos": "1", "posSide": "long"},
+                {"instId": "BTC-USDT-SWAP", "pos": "-1", "posSide": "short"},
+            ],
+        })
+        assert order_executor.get_status()["position_status"] is POSITION_UNKNOWN
+
     def test_account_balance_empty_raises(self, order_executor, mock_okx_account_client):
         """get_account_balance raises when data is empty."""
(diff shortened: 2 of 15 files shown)


BY571:
@CharlieHelps review

PR #297, opened 2026-08-08
fix(envs): run the agent's SLTP action before bar N+1's trigger check

### LINKED ISSUE #292: SLTP envs: a trigger on bar N+1 cancels the agent's action from bar N's close

BY571:
Closes #292.

## The bug

The agent observes bar N and its order fills at `close(N)`; bar N+1 unfolds afterwards. But `_step` advanced the sampler and checked the position held *coming into* the step against bar N+1 **before** running the agent's action — so a held position whose bracket fired on N+1 discarded that action outright. Not deferred, discarded.

The account ended the step in a position it had explicitly asked to leave, at a price it never chose, and had to re-issue next step.

The premise is verifiable in the code rather than inferred: `cached_price` is captured before the sampler advances, commented *"Bar N price — where the agent's action would execute"*. The obvious counter-argument — a resting bracket order fills intrabar before a new decision round-trips — is real in live trading but does not transfer here, because the agent's own close also executes at `close(N)`, strictly earlier. There is no latency model in the offline env to appeal to.

## Reproduced

Long open at 100 with TP 102.5; agent requests a flip to short on the bar before a spike to 120.

| | `action_type` | direction | portfolio |
|---|---|---|---|
| before | `sltp_tp` | flat | 10500 |
| after | `sltp_sl` | flat | 9500 |

After the fix the long closes flat at `close(N)=100`, the short opens at 100, and the spike stops the **short** out at its own SL. Worth noting the old behaviour was **not** conservative — here it handed the account a win it never asked for, so this is not a one-directional safety margin.

## The fix

Both offline SLTP envs shared the shape — scalar via the `if/elif` priority chain, vectorized via `active = ~triggered_mask` masking triggered envs out of every action mask. Both now:

1. execute the agent's action at `close(N)`
2. apply bar N+1 once, to whatever is held after it

That second step is exactly what the #268 entry-trigger block already did for freshly opened positions; it becomes the *only* check rather than a bolt-on, so a step now makes strictly fewer trigger calls than before.

`lock_position_until_sltp` is applied up front by forcing the action to `None`, so a locked position is still what bar N+1 gets checked against.

## Scope note

**Liquidation had the identical defect for the identical reason** and is folded into the same post-action check. The issue names only the SL/TP chain, so this is deliberately broader than the text — but splitting liquidation back out would reintroduce the redundant double-check the merge removes.

## Live envs are unaffected

`env_sltp.py::_step` calls `_sync_position_from_exchange` **before** decoding the agent's action, so the exchange is the source of truth for whether a bracket already filled. There is no synthetic future-bar check to get backwards — the "exchange truth beats the cache" invariant this codebase already documents. No change to `sltp_mixin.py` or any `env_sltp.py`.

## Tests

No existing test encoded the bug — verified rather than assumed. One test (`test_entry_bar_exit_after_direction_switch`) had been written to *dodge* it, with a comment citing #292; that workaround note is now stale and updated.

Added a regression test through the existing scalar/vectorized equivalence harness, so it pins the behaviour in both envs and their parity at once. It was confirmed to fail for the right reason before the fix (`sltp_tp` recorded where `sltp_sl` is expected).

## Verification

```
Full suite (CUDA_VISIBLE_DEVICES=""):  2296 passed, 18 skipped, 0 failed
Offline env suite:                     659 passed
```

Run against `.venv/bin/python` (torchrl 0.13.2, matching the pin).

## Review

The mandated `torchrl-engineer` design pass ran before implementation, and three full rounds of code-simplifier / pr-test-analyzer / torchrl-engineer / test-justification ran after it, with findings fixed between each.

The rounds earned their keep — every round found a defect in the previous round's fix:

- **Round 1**: the two broadenings beyond the issue text (liquidation, and the money the exit is booked at) had *no* coverage — reverting either left the suite fully green. The one regression test pinned which action type fired, never a balance.
- **Round 2**: removing the outer position guard made the callee guard in `_check_sltp_trigger` load-bearing with nothing holding it — deleting it silently rewrote the agent's close to `hold` in history. The dead-test shape fixed in the scalar env was still present in the vectorized one. The affordability gate had no test at all.
- **Round 3**: the affordability test added in Round 2 was itself passing for the wrong reason — `assert (_balances >= 0).all()` cannot fail, because `clamp_` runs inside the call.

Final differential fuzz against `main`: 6,000 scalar and 3,500 vectorized scenarios, 1,464 and 1,859 divergences, **100% explained by the intended reorder, zero unexplained**.

## Not included

`#298` — the documented "liquidation takes priority over the bracket" rule is unenforced in both envs: swapping the two checks leaves the suite green, because no test builds a bar where both would fire. Pre-existing on `main`, filed separately.

### DIFF
--- a/tests/envs/offline/test_sequential_sltp.py
+++ b/tests/envs/offline/test_sequential_sltp.py
@@ -1245,7 +1245,6 @@ class TestLockPositionUntilSLTP:
         assert config.lock_position_until_sltp is False
 
 
-
 HOLD = (None, None, None)
 CLOSE = ("close", None, None)
 LONG_TIGHT = ("long", -0.025, 0.025)
@@ -1266,7 +1265,7 @@ class TestAgentActionPrecedesNextBarTrigger:
     """
 
     def _run(self, actions, *, leverage, sl_levels, tp_levels, wick_high=None,
-             wick_low=None, include_close=False, wick_bar=20):
+             wick_low=None, include_close=False):
         import numpy as np
         import pandas as pd
 
@@ -1278,9 +1277,9 @@ class TestAgentActionPrecedesNextBarTrigger:
             "close": flat.copy(), "volume": np.ones(n) * 1000,
         })
         if wick_high is not None:
-            df.loc[wick_bar, "high"] = wick_high
+            df.loc[20, "high"] = wick_high
         if wick_low is not None:
-            df.loc[wick_bar, "low"] = wick_low
+            df.loc[20, "low"] = wick_low
 
         config = SequentialTradingEnvSLTPConfig(
             leverage=leverage, initial_cash=10000,
@@ -1351,8 +1350,10 @@ class TestAgentActionPrecedesNextBarTrigger:
             leverage=2, sl_levels=[-0.025, -0.50], tp_levels=[0.025, 0.50],
             wick_high=120.0, include_close=True,
         )
-        assert "sltp_tp" not in env.history.action_types, (
-            "the bracket fired instead of the agent's close"
-        )
+        # The close must be RECORDED as a close. A stale bracket firing on the now-flat
+        # account is a financial no-op (_execute_sltp_close returns early when flat), so
+        # it moves no balance -- it just overwrites the agent's close with a phantom
+        # hold, and history feeds binarize_action_type and the reward function.
+        assert env.history.action_types[-3] == "close"
         assert env.position.position_size == 0
         assert env.balance == pytest.approx(10000.0), "a flat round trip must not move the balance"
--- a/tests/envs/offline/test_sequential_sltp.py
+++ b/tests/envs/offline/test_sequential_sltp.py
@@ -1314,10 +1314,9 @@ class TestAgentActionPrecedesNextBarTrigger:
             [HOLD] * 5 + [LONG_TIGHT] + [HOLD] * 3 + [SHORT_TIGHT] + [HOLD] * 2,
             leverage=2, sl_levels=[-0.025, -0.50], tp_levels=[0.025, 0.50], wick_high=120.0,
         )
-        assert "sltp_tp" not in env.history.action_types, (
-            "the incoming long's TP fired, so the agent's switch was discarded"
-        )
-        assert env.history.action_types[-3] == "sltp_sl", "the SHORT should have been stopped out"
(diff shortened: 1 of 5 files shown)


BY571:
@CharlieHelps review

PR #301, opened 2026-08-09
fix(envs): compute vectorized offline money in float64 (#293)

### LINKED ISSUE #293: Scalar vs vectorized SLTP diverge at high leverage for notional/quantity trade modes

BY571:
Fixes #293.

## The bug

`VectorizedSequentialTradingEnvSLTP` tracked balances, positions and bracket prices in **float32** while `SequentialTradingEnvSLTP` used Python floats (float64). float32 carries ~7 significant digits; leverage scales the absolute magnitudes, so the accumulated relative epsilon lands on a threshold and flips a comparison.

Reproduced deterministically at `leverage=25`, `trade_mode="fractional"`, `lock_position_until_sltp=True`:

```
step 39: scalar balance 100.00000000   vec balance 99.99975586
         bankruptcy threshold = 1000 * 0.1 = 100.0, compared with strict `<`
scalar:  100.00000000 < 100.0  ->  False  ->  episode continues
vec:      99.99975586 < 100.0  ->  True   ->  TERMINATED
```

The two engines disagree about whether the episode ended, so a policy evaluated on one and deployed against the other's assumptions sees different exits.

## Why float64 and not a tolerance

The obvious alternative is a relative tolerance on the threshold comparisons. Measured, it does not hold up: the reproduced gap is `2.441e-06` relative, so the natural `1e-6` constant **still terminates** — the tolerance must exceed the accumulated error, which grows with leverage and episode length, making any fixed constant a guess.

I also tested the hypothesis that the divergence came from the two engines computing position size by differently-ordered but algebraically identical formulas. Aligning them changes the reproduced balance **not at all** — same value to the last bit. The cause is plain accumulated float32 rounding.

After the change the two engines agree to better than **1e-12** across the entire equivalence suite.

## Cost

Pinned `.venv`, CPU, `OMP_NUM_THREADS=1`, best of 5:

| num_envs | float32 | float64 | delta |
|---|---|---|---|
| 64 | 154,006 steps/s | 155,058 steps/s | +0.7% (noise) |
| 256 | 547,732 steps/s | 534,276 steps/s | −2.5% |
| 1024 | 1,575,547 steps/s | 1,501,097 steps/s | −4.7% |

Up to ~5% at the largest batch, still 1.5M steps/s. Observations and reward are cast back to float32 at emission, so specs and networks are unchanged.

## Seeded reproducibility is preserved (a near-miss)

`uniform_` consumes more generator bits at float64, so drawing initial cash and slippage noise straight into `MONEY_DTYPE` shifted every subsequent draw. Seed 42 went from cash `[1382.27, 1415.00, 882.86, 1459.31]` / starts `[231, 233, 238, 41]` to `[558.15, 562.91, 623.59, 552.58]` / `[229, 77, 253, 181]` — silently different trajectories for anyone with a pinned seed.

The draws now stay float32 and are widened afterwards, so the RNG stream is **bit-identical to main**. The sampled value is an input; it was the accumulation that needed the wider type.

## Two further bugs found by getting the dtype right

- **`can_afford` tolerance.** The vectorized envs used `1e-5` with a comment justifying it as a float32 workaround; the scalar envs used `1e-9`. The vectorized env was *accepting opens the scalar env refused*. All five sites now share `AFFORDABILITY_REL_TOL`. Genuinely coupled to the dtype: under float32, `1e-9` fails.
- **Slippage noise** was drawn in float32 and multiplied into a float64 price. Every equivalence test runs `slippage=0.0`, which is why nothing caught it — that branch had **zero** coverage suite-wide.

## Tests

The harness now varies `trade_mode` × `lock_position_until_sltp` × leverage (12 cells) with randomised fixed-seed actions, and the equivalence tolerance is tightened from `atol=5e-4, rtol=1e-3` to `1e-9`. The old value was close to vacuous on money: it swallowed a real 1.8e-4 balance divergence inside a computed tolerance of 0.24, so only the binary done-flag comparison caught anything. Measured worst case is under 1e-12.

New coverage for previously-unguarded lines: the float32 RNG stream, the vectorized slippage branch, `_action_levels_tensor`'s dtype (unfalsifiable before, since every level in the suite was in `{-1,0,1}` — exact in both dtypes), and the affordability slack on both the SLTP and scale-up paths.

Mutation matrix at the tip:

| Mutation | Tests red |
|---|---|
| `MONEY_DTYPE` → float32 | 75 |
| `account_state` `.float()` removed | 63 |
| `_action_sl/tp_pcts` → float32 | 43 |
| `_base_tensor` cast removed | 42 |
| `_sl/_tp_prices` → float32 | 36 |
| SLTP grid brackets widened back | 12 |
| `AFFORDABILITY_REL_TOL` → 0.0 | 8 |
| `AFFORDABILITY_REL_TOL` → 1e-5 | 2 |
| slippage branch skipped | 2 |
| `_action_levels_tensor` → float32 | 1 |
| initial-cash draw widened to float64 | 1 |
| clean baseline | 0 |

Full suite: 2324 passed, 18 skipped.

## Review

Three rounds, four reviewers each. Each round found real defects in the previous round's fixes:

- **Round 1** — the RNG-stream break above; two of the twelve new grid cells were vacuous (sizing no leverage-1 balance could afford, so they compared two idle envs); a comment named the wrong mechanism for the SL/TP level dtype.
- **Round 2** — the grid fired **zero brackets** in all twelve cells (the fixture spans +5.9%/−1.7% against ±5% levels), and nine cells opened once then held for 99 steps, which the round-1 guard read as maximally healthy. Guard now counts opens and bracket exits.
- **Round 3** — a docstring claiming the base harness does not cover intermediate `action_levels`, made false by a test added later in round 2; and a refutation of my own dismissal of an affordability finding: I checked `_open_position`, found `cost == pv * |action|` algebraically, and generalised to all three base sites without checking `_increase_position_size`, which sizes the delta from portfolio value but pays from balance and *does* reach the 1e-9..1e-5 window (position 9.9999/balance 500 vs position 19.9998/balance 0.0, with a "Negative balance detected" log). Now pinned.

## Follow-up

**#302** — the scalar and vectorized envs disagree by 1.5% of portfolio in 8 steps on intermediate `action_levels`: the scalar resizes in place with a weighted-average entry, the vectorized always closes and reopens. Pre-existing, roughly five orders of magnitude larger than the divergence this PR fixes, and deliberately out of scope. The class docstring now scopes the equivalence claim to exclude it rather than asserting coverage it lacks.

### DIFF
--- a/tests/envs/offline/fundamental/test_margin_accounting.py
+++ b/tests/envs/offline/fundamental/test_margin_accounting.py
@@ -11,6 +11,7 @@ Tests critical calculations that MUST be correct:
 
 import pytest
 import pandas as pd
+import torch
 from torchtrade.envs.offline.sequential import SequentialTradingEnv, SequentialTradingEnvConfig
 
 
@@ -594,6 +595,50 @@ class TestMultiTimeframeHandling:
         pv_multi = env_multi._get_portfolio_value()
         assert abs(pv_single - pv_multi) < 1, \
             f"Single TF PV {pv_single:.2f} != Multi TF {pv_multi:.2f}"
+
+
+class TestAffordabilitySlackOnScaleUp:
+    """Scaling up an existing position must respect AFFORDABILITY_REL_TOL.
+
+    An open from flat is algebraically pinned: it sizes from portfolio value and pays from
+    balance, and those are the same number while flat, so cost == pv * |action| and the
+    overshoot is ~1 ULP. Scaling up is not. It sizes the *delta* from portfolio value but
+    still pays from balance, and once a position is open those differ -- so the relative
+    shortfall is continuous in the price move since entry and sweeps straight through the
+    1e-9..1e-5 window that separated the scalar and vectorized envs before #293.
+
+    The base env's other two call sites are equivalent mutants (the vectorized env never
+    resizes in place -- it closes and reopens, so every env in open_mask is flat there).
+    This is the one that can drift, and at 1e-5 it accepts an open it cannot pay for and
+    logs "Negative balance detected" before clamping to zero.
+    """
+
+    def test_scale_up_respects_the_shared_slack(self):
+        drift = 1e-6
+        prices = [100.0 * (1 + drift) ** i for i in range(40)]
+        df = pd.DataFrame({
+            "timestamp": pd.date_range("2024-01-01", periods=40, freq="1min"),
+            "open": prices, "high": prices, "low": prices, "close": prices,
+            "volume": [1000.0] * 40,
+        }, index=range(40))
+        env = SequentialTradingEnv(df, SequentialTradingEnvConfig(
+            action_levels=[0.0, 0.5, 1.0], leverage=2, initial_cash=1000.0,
+            time_frames=["1Min"], window_sizes=[10], execute_on="1Min",
+            transaction_fee=0.0, slippage=0.0, seed=42, max_traj_length=30,
+            random_start=False,
+        ))
+        td = env.reset()
+        for action in (1, 2):  # half size, then ask for full one bar later
+            td["action"] = torch.tensor(action)
+            td = env.step(td)["next"]
+
+        # Half size is ~10 units against ~20 for full; the scale-up must be refused.
+        assert env.position.position_size < 15.0, (
+            f"the scale-up was accepted at position_size={env.position.position_size!r}, "
+            f"balance={env.balance!r} -- the base env's affordability slack has drifted "
+            "above AFFORDABILITY_REL_TOL"
+        )
+        env.close()
 
 
 if __name__ == "__main__":
--- a/tests/envs/offline/fundamental/test_slippage.py
+++ b/tests/envs/offline/fundamental/test_slippage.py
@@ -356,16 +356,12 @@ class TestSlippageOnClose:
 
 
 class TestVectorizedSlippage:
-    """The vectorized envs' slippage branch, which no test reached at all.
+    """The vectorized envs' slippage branch, which no test reached at all: raising on the
+    first line under `if self.slippage > 0:` left the whole suite green.
 
-    Verified by mutation: raising on the first line under `if self.slippage > 0:` in both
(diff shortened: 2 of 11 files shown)


BY571:
@CharlieHelps review

PR #303, opened 2026-08-09
feat(actor): add optional Adanos sentiment tool

alexander-schneider:
## Summary

Adds an optional `AdanosSentimentTool` to the LLM actor tool loop introduced in #234.

The tool gives a configured actor compact stock or crypto sentiment context from the Adanos Market Sentiment API. It uses the official `adanos` Python SDK, supports Reddit, X, News, and Polymarket for stocks plus Reddit for crypto, and accepts explicit inclusive UTC date windows. The result is presented as additional evidence rather than a trading signal.

## Design

- Reuses the existing `Tool` interface and fail-open live-step behavior.
- Keeps Adanos outside TorchTrade's core dependencies through a dedicated `[sentiment]` extra.
- Reads `ADANOS_API_KEY` only when the tool runs; no credential is stored in the repository.
- Bounds SDK requests with the configured timeout and closes the client in a `finally` block.
- Returns only compact aggregate fields, avoiding raw mention payloads in the model context.
- Includes a runnable `LocalLLMActor` example and updates the actor documentation.

## Verification

- `.venv/bin/python -m pytest tests/actor/test_tools.py tests/actor/test_local_llm_actor.py -q` (46 passed)
- `.venv/bin/python -m py_compile torchtrade/actor/tools.py examples/llm/tools/adanos_sentiment_decision.py`
- `uv pip compile pyproject.toml --extra sentiment` (resolved `adanos==2.7.0`)
- `git diff --check`

The tests cover stock source routing, crypto token routing, explicit date forwarding, timeout configuration, response minimization, missing credentials, API failures, client teardown, and unsupported crypto sources without live network calls.

### DIFF
--- a/docs/components/actors.md
+++ b/docs/components/actors.md
@@ -147,7 +147,7 @@ only when configured (live path); without them the actor is single-shot as befor
 
 ```python
 from torchtrade.actor import LocalLLMActor
-from torchtrade.actor.tools import GoogleNewsTool
+from torchtrade.actor.tools import AdanosSentimentTool, GoogleNewsTool
 
 actor = LocalLLMActor(
     model="Qwen/Qwen2.5-0.5B-Instruct", backend="vllm",
@@ -155,7 +155,10 @@ actor = LocalLLMActor(
     account_state_labels=env.account_state,
     action_levels=env.action_levels,
     symbol="BTC/USD",
-    tools=[GoogleNewsTool(symbol="BTC/USD")],
+    tools=[
+        GoogleNewsTool(symbol="BTC/USD"),
+        AdanosSentimentTool(symbol="BTC/USD", asset_type="crypto"),
+    ],
     max_tool_iters=3,
 )
 ```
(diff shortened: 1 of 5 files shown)


BY571:
Thanks for opening this, and for taking the care to match the existing conventions — the lazy import, the fail-open `run()`, the response field whitelist and the `_fetch` seam all line up with how `GoogleNewsTool` works. It's a clean patch.

We're not going to merge it though, for three reasons:

1. **We want to keep the core vendor-neutral.** Shipping a wrapper around one commercial data provider in-tree is a standing maintenance obligation for an API surface and pricing model we don't control.

2. **Point-in-time correctness.** TorchTrade is used primarily for offline RL training and backtesting. With `from_date`/`to_date` omitted — the default, and they're LLM-supplied so we can't bind them to the simulated clock — the API returns its current window. Replaying a historical episode would expose the policy to present-day sentiment, and that's lookahead bias we can't guard against generically. The plan-gated lookback also caps how far a point-in-time backtest could reach.

3. **The free tier doesn't cover the use case.** At 250 requests/month, a single symbol making hourly decisions exhausts it in about ten days, so in practice the feature needs a paid plan. That's a hard sell for a research-oriented repo.

None of this requires our repo, though. `torchtrade.actor.tools.Tool` is a deliberately small, stable interface — `name`, `description`, and `run(**kwargs) -> str`. You can ship `AdanosSentimentTool` in the `adanos` package itself and it will work with `LocalLLMActor` unchanged, released on your own cadence and documented where you can keep it current. If you do that, we're happy to link it from the actor docs as a third-party tool.

Thanks again for the time you put into it.

Issue #302, opened 2026-08-09
Scalar and vectorized envs disagree on intermediate action_levels: resize-in-place vs close-and-reopen


BY571:
## Bug

`SequentialTradingEnv` and `VectorizedSequentialTradingEnv` implement a *change* in position size differently:

- **Scalar** resizes in place — `_increase_position_size` / `_decrease_position_size` (`sequential.py:790-800`) keep the existing position and update entry price to a weighted average, preserving `hold_counter`.
- **Vectorized** has no such branch — `_execute_trades` (`vectorized_sequential.py:650-696`) always closes the old position and opens a new one, which realises P&L, pays a full round-trip fee, resets entry price, and resets `hold_counter`.

With `action_levels=[-1, -0.5, 0, 0.5, 1]`, leverage 5, fee 0.001, actions `[3,3,4,4,4,3,3,4]`:

```
step 3 act=  1.0 | bal s=   20.5609 v=    0.0000 | entry s= 99.6419 v= 99.1793 | hold s=2 v=1
step 6 act=  0.5 | bal s=  489.9001 v=  460.2324 | entry s= 99.6419 v= 98.3116 | hold s=4 v=1
step 8 act=  1.0 | bal s=   33.7112 v=    0.0000 | entry s= 98.8144 v= 97.9737 | hold s=5 v=1

portfolio: scalar=924.8964 vec=910.9522   rel diff=1.5077%
```

1.5% portfolio divergence in 8 steps, plus a wrong `hold_counter` and a wrong entry price feeding `unrealized_pnl_pct` into the observation.

For scale: #293 was about a 1.8e-4 absolute (~2.4e-6 relative) divergence. This is roughly five orders of magnitude larger.

## Why it was invisible

`tests/envs/offline/test_vec_scalar_equivalence.py` threads an `action_levels` parameter through `_make_pair` (`:23`) and `_run_sequence` (`:108`) to both configs — **and no caller ever passes it.** Every cell runs the default `[0, 1]` or `[-1, 0, 1]`, where every action is flat/full and the two implementations coincide.

Confirmed by mutation: patching the method enclosing `sequential.py:793` to `raise` leaves all 79 equivalence tests passing. Only the scalar-only `fundamental/test_margin_accounting.py` and `fundamental/test_weighted_entry.py` go red.

## Not a regression

[...]

Issue #285, opened 2026-08-07
Dead code sweep: verified-dead symbols, dead config fields, misleading margin_type, orphaned tests/mocks infrastructure


BY571:
All verified with repo-wide reference searches (AST symbol extraction + word-boundary grep across torchtrade/, tests/, examples/, docs/):

**Certain-dead symbols**
- \`alpaca/env.py:332\` \`_create_info_dict\` (zero refs)
- \`binance/env.py:269,281\` \`_handle_long_action\`/\`_handle_short_action\` — uncalled AND reference \`config.quantity_per_trade\` which doesn't exist on the config (pre-fractional leftovers)
- \`bitget/order_executor.py:489\` \`check_both_positions_open\` (zero refs)
- \`binance/order_executor.py:17\` \`PositionSide\` enum + its 3 \`__init__\` re-exports (never used anywhere)
- \`core/offline_base.py:306-349\` \`_get_portfolio_value\` — shadowed by its only subclass (\`sequential.py:367\`); the ~30 lines of per-step validation inside never run
- \`offline/sequential.py:247-250\` \`_reset_history\` — byte-identical override of the base
- \`utils/action_maps.py:126\` \`combinatory_action_map\` back-compat alias (zero refs)
- \`tests/mocks/{binance,bitget,base,conftest}.py\` — 244 LOC parallel mock infrastructure nothing imports (live fixtures are re-implemented in \`tests/envs/*/conftest.py\`); \`tests/envs/alpaca/mocks.py\` is a pure re-export shim

**Dead config that silently misleads**
- \`position_penalty\` field in alpaca/binance/bitget env configs — zero reads repo-wide
- \`SequentialTradingEnvConfig.margin_type\` (\`sequential.py:73\`) — stored, never read; \`MarginType.CROSS\` silently behaves as isolated. Either wire it or remove it and document isolated-only
- \`include_base_features\` (\`sequential.py:62\`) — never read by any offline env

**Minor**: unused locals \`alpaca/base.py:223-224\`, \`alpaca/env_sltp.py:221\`, \`bitget/order_executor.py:330\`; commented-out code in \`alpaca/observation.py:96,125-127\`; ~7 unused test imports (ruff F401/F841 list).

[...]

Issue #275, opened 2026-08-07
Offline holding_time freezes at 1 when a position is resized


BY571:
## Bug

\`sequential.py:749\` sets \`hold_counter = 1\` on open; it is incremented only in the two no-trade branches (\`:657\`, \`:672\`). \`_increase_position_size\`/\`_decrease_position_size\` never touch it.

\`\`\`
step 0  level=1.0  holding_time=1
step 1  level=0.5  holding_time=1
...
step 7  level=0.5  holding_time=1    ← continuously open 8 steps, reports 1
\`\`\`

\`account_state[3]\` is documented as "steps since position opened". The offline envs never adopted \`advance_hold_counter\` (\`core/state.py:169\`) — the canonical helper the live envs were already fixed to use. Classic fix-the-instance-not-the-class.

## Fix

Route offline hold counting through \`advance_hold_counter\` (resize = still holding → increment). Regression test: resized position must report cumulative holding time.

BY571:
Fixed by #304 (merged). GitHub only auto-closed #274 from that PR's `Fixes #274, #275` line.

All four offline engines now age a position once per step from the post-trade direction, through `advance_hold_counter_from_size` (scalar, scalar-SLTP) or `_advance_hold_counters` (vectorized base and SLTP) — the tensor form of the same rule. Seventeen hand-rolled sites were removed, so a resize keeps ageing instead of freezing, and a future exit path cannot forget to age by omission.

Two things this issue's diagnosis got slightly off, worth recording:
- The scalar env did not freeze at 1; it froze at whatever value the counter already held. `_increase_position_size`/`_decrease_position_size` never touched it, so a position opened and resized on the next bar happened to read 1 — the common case, not the rule.
- The vectorized envs were wrong in the other direction: they reset to 1 through the close-and-reopen, discarding the age entirely.

Regression cover: a cell in `test_sequential_holding_time_sequence` pins that a resized position keeps ageing, and a `tolerance-hold` equivalence cell pins the one branch nothing else entered with a position open.

PR #304, opened 2026-08-10
fix(envs): trade only the delta on a partial resize, and age positions canonically (#274, #275)

### LINKED ISSUE #274: Vectorized env close-and-reopens on partial target changes — double fees, wrong entry price; equivalence docstring overclaims
## Bug

Scalar \`SequentialTradingEnv\` partially adjusts positions, trading only the delta and maintaining a quantity-weighted average entry (\`sequential.py:762-858\`). \`VectorizedSequentialTradingEnv._execute_trades\` (\`vectorized_sequential.py:630-677\`) has no equivalent: **any** target change closes 100% and reopens.

Measured with \`action_levels=[-1,-0.5,0,0.5,1]\`, fee 0.001, alternating 1.0/0.5 for 6 steps:
\`\`\`
scalar PV = 9959.2252  entry=100.5252
vector PV = 9909.5859  entry=100.7660   → 0.498% portfolio divergence after 6 steps
\`\`\`
Divergent entry price also means divergent liquidation price and \`unrealized_pnl_pct\`.

The class docstring (\`vectorized_sequential.py:105-109\`) claims "passes extensive equivalence tests against SequentialTradingEnv" — only true for \`action_levels=[-1,0,1]\` where no intermediate target exists.

## Related divergence (same PR)

Start-of-data padding: scalar sampler zero-pads short windows (\`sampler.py:287-293\`); vectorized repeats the first bar (\`vectorized_sequential.py:452\` \`clamp(min=0)\`). Same state, different observations.

## Fix

Implement delta-trading + weighted-average entry in \`_execute_trades\`, align the padding, and extend the equivalence tests to multi-level action spaces (which currently would have caught this).

BY571:
Fixes #274 and #275. Both were invisible to the suite: 2321 tests passed before this change and 2321 after, unchanged.

## 1. Partial resize traded as a full round trip (#274)

`VectorizedSequentialTradingEnv._execute_trades` sent **every** target change through close-then-reopen, including a same-direction resize. That pays a full round trip of fees and throws the entry price away — which also moves the liquidation price and `unrealized_pnl_pct`.

Measured against `SequentialTradingEnv`, `action_levels=[-1,-0.5,0,0.5,1]`, fee 0.001:

| scenario | before | after |
|---|---|---|
| lev 1, oscillate full↔half, 6 steps | 0.4982% of portfolio | **0.0000%** |
| lev 5, oscillate full↔half, 8 steps | 3.4566% | **0.0000%** |
| lev 5, gradual resize, 8 steps | 1.4983% | **0.0000%** |

Entry prices are now bit-equal too.

The fix classifies each trading env into four disjoint masks — resize / switch / close-to-flat / open-from-flat — and narrows the close and open masks to the last three. The resize branch uses the scalar engine's exact formulas: fee and margin on the delta only, a quantity-weighted entry on an increase, and an **untouched** entry on a decrease (partially closing does not re-price the remainder, so the liquidation price is unchanged).

## 2. holding_time did not survive a resize (#275)

`account_state[3]` is documented as "steps since position opened". Both engines got it wrong, in opposite directions:

- **scalar** hand-rolled the counter at five sites and never gave the resize path one, so a position resized every bar reported `holding_time=1` forever
- **vectorized** reset it to 1 through the reopen, discarding the age

`advance_hold_counter` in `core/state.py` has specified this rule since the live envs were fixed to share it — and no offline env ever called it. They only cited it in comments. All **four** offline engines now age once per step from the post-trade direction: the two scalar engines through `advance_hold_counter_from_size` (extracted to `core/state.py` in review, so the size→direction conversion is not itself duplicated), the two vectorized engines through `_advance_hold_counters`, its tensor form. Seventeen hand-rolled sites were removed.

**That part was not optional, and a test proved it.** My first attempt incremented inside the resize branch instead. A new equivalence cell caught it immediately: on a *refused* resize (unaffordable at leverage 5) the vectorized position stayed un-aged while the scalar aged it — because the scalar's call sits in `_step` and runs whether or not the trade executed. Six branches and three early returns is exactly the shape `core/state.py` warns about.

## Cost

~20% throughput on the vectorized env at worst-case churn (uniform random over five levels, so nearly every step is a resize):

| num_envs | main | this PR |
|---|---|---|
| 256 | 579,558 steps/s | 457,880 |
| 1024 | 1,659,959 | 1,332,503 |
| 4096 | 3,308,434 | 2,686,819 |

A trained policy holds far more often than uniform-random, so this is an upper bound. Still ~2.7M steps/s, orders of magnitude above the collector ceiling.

## Tests

`action_levels` was already threaded through `_make_pair` and `_run_sequence` and **no caller ever passed it** — every cell ran the default `{-1,0,1}` where every action is flat-or-full and the resize branch is unreachable. That is the axis this bug lived behind. Now covered by `TestScalarVecEquivalenceResize`: four equivalence cells (`oscillate-spot`, `oscillate-levered`, `oscillate-short`, `through-flat`), plus `test_an_unaffordable_increase_changes_nothing_but_the_age` and `test_resize_charges_the_delta_fee_exactly` — the latter pinning the absolute fee on *both* engines, so a joint regression to close-and-reopen still fails. One new cell in `test_sequential_holding_time_sequence` pins that a resized position keeps ageing.

Mutation-verified across 23 mutations of the changed production lines — resize arithmetic, mask classification, and the ageing rule — with **no survivors**. Per-cell, `oscillate-short` is the sole killer of dropping the sign in the decrease branch (which carries through `closed_qty`, `pnl` and `freed_margin`), and `through-flat` is the sole killer of both switch-classified-as-resize and close-to-flat. `oscillate-spot` is not the sole killer of anything; it is kept as cheap coverage of a supported mode, not claimed as load-bearing.

## On #302

I filed #302 while reviewing #301, not having found #274 first — same mechanism, same file, same lines. It is closed as a duplicate, and the class docstring it was cited in (`vectorized_sequential.py`) now points at the coverage this PR adds rather than at a closed issue. Two things #302 contributed that #274 did not name are carried here: the `hold_counter` divergence (fixed as #275), and `margin_type` being dead config on both engines (left for #285, the dead-code sweep).

## Not fixed, deliberately

#274 also names a start-of-data padding divergence (scalar zero-pads, vectorized repeats the first bar). **Both branches are unreachable.** `sampler.py`'s `min_start_time` guarantees every window has full lookback — verified across five multi-timeframe configs, every `first_idx >= window_size - 1` — and `TestUndersizedWindowPadding` has to monkeypatch `_obs_indices` to reach the branch at all. It is a code inconsistency, not a live bug, and aligning it is defence-in-depth rather than a fix.

### DIFF
--- a/tests/envs/offline/test_holding_time.py
+++ b/tests/envs/offline/test_holding_time.py
@@ -53,9 +53,6 @@ def _drive(env, actions):
     # keeps ageing. Before the fix this reported 1 forever -- the scalar env incremented
     # only in its two no-trade branches and the resize path never got one.
     (5, [-1.0, -0.5, 0.0, 0.5, 1.0], [4, 3, 4, 3, 4], [0, 1, 2, 3, 4, 5]),
-    # resize, then a direct flip OUT of the resized position, then resize the new short:
-    # the flip cell above never leaves a resized position, so it cannot catch this.
-    (5, [-1.0, -0.5, 0.0, 0.5, 1.0], [4, 3, 1, 0], [0, 1, 2, 1, 2]),
 ])
 def test_sequential_holding_time_sequence(sample_ohlcv_df, leverage, action_levels, actions, expected):
     """The opening bar reads 1, holds increment, close resets to 0, a reopen restarts at 1."""
--- a/tests/envs/offline/test_holding_time.py
+++ b/tests/envs/offline/test_holding_time.py
@@ -49,6 +49,13 @@ def _drive(env, actions):
     (10, [-1, 0, 1], [2, 2, 2, 1, 0, 0], [0, 1, 2, 3, 0, 1, 2]),
     # futures DIRECT flip: open long, hold, flip straight to short (no flat bar), hold
     (10, [-1, 0, 1], [2, 2, 0, 0], [0, 1, 2, 1, 2]),
+    # RESIZE every bar (#275): a half/full oscillation is one position all along, so it
+    # keeps ageing. Before the fix this reported 1 forever -- the scalar env incremented
+    # only in its two no-trade branches and the resize path never got one.
+    (5, [-1.0, -0.5, 0.0, 0.5, 1.0], [4, 3, 4, 3, 4], [0, 1, 2, 3, 4, 5]),
+    # resize, then a direct flip OUT of the resized position, then resize the new short:
+    # the flip cell above never leaves a resized position, so it cannot catch this.
+    (5, [-1.0, -0.5, 0.0, 0.5, 1.0], [4, 3, 1, 0], [0, 1, 2, 1, 2]),
 ])
 def test_sequential_holding_time_sequence(sample_ohlcv_df, leverage, action_levels, actions, expected):
     """The opening bar reads 1, holds increment, close resets to 0, a reopen restarts at 1."""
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps review

PR #310, opened 2026-08-10
fix(envs): check liquidation against the bar before serving it to the policy (#281)

### LINKED ISSUE #281: Offline: liquidation applied one step after the agent has already observed the breaching bar
\`sequential.py:527\` checks liquidation at the *start* of the next \`_step()\`, but the observation for that bar was built at the *end* of the previous step — before the check. Reproduced (5x long, liq 80.40, bar 10 low=70):

\`\`\`
acted_on_bar=9  → obs shows bar 10: dir=+1  dist_to_liq=0.196  pv=10000.00   ← position already dead
acted_on_bar=10 → dir=0  pv=200.00
\`\`\`

The policy sees a healthy levered long on a bar that already breached liquidation, and reward/termination land one step late. Cumulative return telescopes correctly; per-step attribution and the observation do not.

Fix direction: check liquidation against the newly fetched bar *before* building its observation (the SLTP env's post-#268 ordering should be the template — coordinate these two fixes).

BY571:
Fixes #281.

## The bug

`SequentialTradingEnv._step` checked liquidation at the **start** of the next step — against the bar whose observation had already been handed to the policy one call ago — and left the freshly-fetched bar unchecked until the call after that. So a bar whose wick breached the liquidation price was served to the policy as a healthy position:

```
step 9:  dir=+1  dist_to_liq=0.196  pv=10000.00  hold          <- low 70, liq price 80.40
step 10: dir=+0  dist_to_liq=1.000  pv=  200.00  liquidation
```

`dist_to_liq=0.196` is not merely stale — it is the opposite of true. The bar it describes had already wiped the position. The exit now lands on step 9.

**The wick-only shape is load-bearing.** My first reproduction breached on the close and showed nothing: the observation reads unhealthy anyway, so the ordering bug is invisible. It only surfaces when the low breaches and the close recovers.

## Two defects, not one

The scalar env had **both** defects the SLTP env had before #294 and #297, and the issue only names one. Its liquidation and its trade were a mutually-exclusive `if/else`, so a legitimate close or switch was silently discarded whenever liquidation fired — the #292 defect. Fixing only the observation ordering would have left that in place.

The order now mirrors the SLTP env's post-#294-and-#297 form: fetch bar N+1, run the agent's action at bar N's close **unconditionally**, then apply bar N+1's wick to whatever that left open.

## Both engines, together

`VectorizedSequentialTradingEnv` carried both defects too, and deliberately — its docstring read *"Matches scalar SequentialTradingEnv order: check liquidation at current bar (before trade)"*. The 1e-9 equivalence harness kept the two engines agreeing while both were wrong. Fixing only the scalar turns that harness red immediately (verified), so both move in this commit.

The two SLTP engines were already correct and are untouched.

## The existing regression test was dead

`test_liquidation_intrabar_wick` claims to cover exactly this scenario — *"intrabar wick that recovers must still trigger liquidation"* — and never did. Its "hold" step submitted action index `1`, which under `action_levels=[-1, 0, 1]` is value `0.0`: a **close**, not a hold.

Instrumented, it flattened the position on a clean bar and then asserted `position_size == 0`:

```
after open : size=1000.0000  liq=90.40
after 'hold': size=0.0000    balance=10000.00
history     = ['hold', 'long', 'flat']
LIQUIDATED? False
```

Balance back at exactly 10000 — the position was closed at no loss, the wick bar never reached. It passed for an unrelated reason, and would have kept passing after this fix. It now holds by re-submitting the opening action and asserts a liquidation is actually recorded; verified to fail against the old ordering.

## Tests

Two new tests pin what the issue is actually about — the step the exit is *observed* on — one per engine, each verified to fail against its own engine's old ordering. Suite: 2331 passed, 21 skipped.

## Scope

#280 (SL/TP fills at the exact bracket price, ignoring gaps) is **independent** — it concerns what price a trigger fills at once it is decided, not which bar the check runs against. Worth flagging for whoever picks it up: `_execute_liquidation()` in `sequential.py` fills at exactly `self.liquidation_price` with the same gap-blindness #280 describes for SL/TP, and #280's writeup does not mention it.


BY571:
@CharlieHelps review this PR thoroughly

PR #321, opened 2026-08-11
test(alpaca): delete the dead SLTP mock trader (#319)

### LINKED ISSUE #319: MockSLTPTrader.simulate_price_movement is dead, and carries the #280 bug shape

`tests/envs/alpaca/test_torch_env_sltp.py:46` `MockSLTPTrader.simulate_price_movement` books its exit at `self.active_stop_loss` — a fifth SL/TP fill site with the same bug shape #311 fixed in the other four (a stop is a market order once touched, so a bar that gaps past it fills at the open).

It is also **dead**: grep finds exactly one hit, the definition itself.

So this is a dead mock that models the wrong fill semantics — if someone revives it as a reference for live-env SLTP behaviour, it teaches the bug back. Delete it, or fix it and give it a caller.

Left out of #311 to avoid widening a fill-pricing fix into the Alpaca test mocks.

BY571:
Closes #319. Fallout from #311.

## The problem

`MockSLTPTrader.simulate_price_movement` booked its exit at `self.active_stop_loss` — the exact bug shape #311 just fixed in the four real engines (a stop is a market order once touched, so a bar that gaps past it fills at the open).

It had **no caller**: grep found only the definition. So it was a dead mock modelling the wrong fill semantics, waiting for someone to revive it as a reference for live-env SLTP behaviour and learn the bug back.

## Why the whole subclass goes

Deleting the simulator orphans everything else in the class:

- `_close_position_at_price` was called only from `simulate_price_movement`.
- `active_stop_loss`, `active_take_profit`, `bracket_order_active` were written in `__init__` and `trade`, and read **only** by the simulator. (Every other hit on those names in the suite is `env.active_stop_loss` — the environment, a different object.)
- What remained was a `trade()` override that recorded that write-only state and delegated the rest. `MockTrader.trade` already accepts `**kwargs`, so it swallows `take_profit`/`stop_loss` unchanged.

The eight construction sites now use `MockTrader` directly.

## Testing

Tests only — no production code touched. `tests/envs/alpaca`: **132 passed**. Full suite: **2396 passed, 19 skipped**.

### DIFF
--- a/tests/envs/alpaca/test_torch_env_sltp.py
+++ b/tests/envs/alpaca/test_torch_env_sltp.py
@@ -60,29 +60,6 @@ class TestCombinatorActionMap:
 class TestAlpacaSLTPTradingEnvInitialization:
     """Tests for environment initialization."""
 
-    def test_init_with_mocks(self):
-        """Test initialization with injected mocks."""
-        config = AlpacaSLTPTradingEnvConfig(
-            symbol="BTC/USD",
-            window_sizes=[10],
-            paper=True,
-            stoploss_levels=(-0.05, -0.1),
-            takeprofit_levels=(0.05, 0.1),
-        )
-
-        mock_observer = MockObserver(window_sizes=[10])
-        mock_trader = MockTrader(initial_cash=10000.0)
-
-        env = AlpacaSLTPTorchTradingEnv(
-            config=config,
-            observer=mock_observer,
-            trader=mock_trader,
-        )
-
-        assert env.config == config
-        assert env.observer is mock_observer
-        assert env.trader is mock_trader
-
     def test_action_spec_size(self):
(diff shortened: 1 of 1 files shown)


BY571:
@CharlieHelps please review.

Short version: `MockSLTPTrader.simulate_price_movement` booked SL/TP exits at the bracket price — the bug #311 just fixed in all four real engines — and had no caller. Deleting it orphaned the rest of the subclass (private helper called only from there; three attributes read only from there), so the whole class goes and the 8 construction sites use `MockTrader`, which already accepts the same kwargs.

Worth your attention: whether swapping `MockSLTPTrader` → `MockTrader` at those 8 sites silently changes what any of those tests exercise. Tests-only diff; full suite 2396 passed.

Issue #320, opened 2026-08-11
Coarse-timeframe END shift uses a fixed period, so a tz-aware Day frame is visible before it closes


BY571:
## Bug

The anti-lookahead END shift for coarse timeframes uses a **fixed** `pd.Timedelta`:

```python
if tf > execute_on:
    offset = pd.Timedelta(tf.to_pandas_freq())
    resampled.index = resampled.index + offset
```

For a Day-or-longer *observation* timeframe on a tz-aware index, that fixed offset does not match the bin's real width. `Day` is not a Tick offset in pandas, so tz-aware `1D` bins follow **local midnight** and are 23h or 25h wide across DST transitions.

Confirmed with pandas on `America/New_York`: the bar starting `2024-11-03 00:00-04:00` is 25h wide, but the fixed 1D shift labels it `2024-11-03 23:00-05:00` — so it becomes visible **an hour before it actually closed**. That is lookahead, not staleness.

## Scope

This is the *observation-frame* mirror of the hazard #282 guarded on the *execution* side. That PR added:

```python
if df.index.tz is not None and execute_on >= TimeFrame(1, TimeFrameUnit.Day):
    raise ValueError(...)
```

which covers `execute_on >= Day`. It does **not** cover a Day-or-longer observation frame under a sub-day `execute_on` — e.g. `time_frames=[1Hour, 1Day], execute_on=1Hour` on a tz-aware index.

## Why it is filed separately

The END shift predates #282 — it comes from #10 — so this is not a regression from that PR, and fixing it means touching the shift itself rather than the fine-frame lookup #282 introduced. Bundling it would mix a pre-existing defect into a branch whose production diff is deliberately narrow.

## Fix

Either extend the existing guard to reject a tz-aware index when **any** timeframe is Day-or-longer, or make the shift width per-bin (`resampled.index.shift(1, freq=...)`-style) so it follows the real bin boundary rather than a fixed duration.

The first is a one-line change to a guard that already exists. The second is more correct but touches the shift used by every coarse timeframe.

## Reachability

No dataset in this repo is tz-aware today — `grep` for `tz_localize|tz_convert|utc=True` across `torchtrade/` and `examples/` returns only the #282 guard itself. But tz-aware input is accepted and working: `pd.to_datetime(..., utc=True)` is a standard market-data idiom, alpaca-py returns tz-aware, and HuggingFace/Arrow round-trips `timestamp[us, tz=UTC]` straight through.

Found during the #282 review.

Issue #10, opened 2025-12-28
Higher timeframe observations may contain future data due to pandas resampling behavior


BY571:
## Problem Description

When using multi-timeframe observations in `MarketDataObservationSampler`, higher timeframes (5min, 15min, 1hour, etc.) may contain data from timestamps **after** the current execution time. This is a form of lookahead bias that can lead to unrealistically optimistic backtesting results.

### Root Cause

Pandas resampling indexes bars by their **START** timestamp, but the bar contains aggregated data through the **END** of the period:

```
5-minute bar at 00:25:00 contains:
  - open: from 00:25:00
  - high: max of 00:25:00 - 00:29:59
  - low:  min of 00:25:00 - 00:29:59  
  - close: from 00:29:59  ← FUTURE DATA when queried at 00:25:00
  - volume: sum of 00:25:00 - 00:29:59
```

### Current Behavior

When `get_observation(timestamp=00:27:00)` is called:

1. The sampler uses `torch.searchsorted(idx_tensor, ts_t, right=True)` to find bars
2. For the 5-minute timeframe, this returns the bar starting at 00:25:00
3. This bar's close value is from 00:29:59 - **2 minutes in the future**

### Code Location

`torchtrade/envs/offline/sampler.py`, lines 145-180:

```python
def get_observation(self, timestamp: pd.Timestamp) -> Dict[str, torch.Tensor]:
    # ...
    pos_t = torch.searchsorted(idx_tensor, ts_t, right=True)
    pos = int(pos_t.item())
    idx_pos = pos - 1  # last index <= timestamp
    # ...
    window = arr[start: idx_pos + 1]  # This window may include "future" data
```

---

### Why This MAY Be Critical

1. **Backtesting Validity**: Models trained with lookahead bias will show inflated performance metrics that won't replicate in live trading.

2. **Feature Engineering**: If features are computed from higher timeframe data (e.g., "5-min close > 5-min open"), these features use future information.

3. **Decision Making**: An RL agent that sees the 5-minute close at minute 25 has unfair knowledge of where the price will be at minute 29.

4. **Subtle Bug**: This only affects higher timeframes in multi-timeframe setups, making it easy to miss.

### Why This MAY NOT Be Critical

1. **Execution Timeframe is Clean**: The execution timeframe (typically matching the base data frequency) does NOT have this issue. Tests confirm 1-min observations at time T contain only data ≤ T.

[...]

PR #322, opened 2026-08-11
fix(envs): align every observation timeframe to the fill instant, and compare timeframes by duration (#282)

### LINKED ISSUE #282: Sampler multi-timeframe alignment: sub-execute_on windows stale by one period; TimeFrame comparison is not a total order
## Two coupled sampler defects

**1. Fine timeframes lag execution** — \`sampler.py:107\` shifts only \`tf > execute_on\`. Measured with \`[1Min, 1Hour, 4Hour]\`, \`execute_on=1Hour\`, exec label 14:00: execution price = minute bar 14:59's close, but the 1Minute window's last close = minute bar **14:00** — 59 minutes stale, while the 1Hour window contains the execution price itself. The fine window — whose whole purpose is recent detail — is the stale one. Not lookahead, but the alignment is not what the config implies.

**2. \`TimeFrame\` ordering is inconsistent** — \`utils/timeframe.py:66-76\`: \`@total_ordering\` with \`__eq__\` on (value, unit) but \`__lt__\` on total minutes. Result: \`60Min == 1Hour\` is False while \`1Hour > 60Min\` AND \`60Min > 1Hour\` are both True. The anti-lookahead shift at \`sampler.py:107\` depends on this comparison: a same-duration-differently-spelled timeframe gets shifted forward by its own period and goes one bar stale. \`normalize_timeframe_config\` warns but does not canonicalize.

Fix: compare on total duration consistently (make \`__eq__\` duration-based or canonicalize units at construction), and align sub-execute_on windows to the execution bar's close.

BY571:
Closes #282 — both defects it filed.

## ⚠️ Breaking for trained policies

Two categories of config see different observations after this. Both need retraining.

| config | before | after |
|---|---|---|
| a timeframe **finer** than `execute_on` | up to a full execution bar stale | ends at the instant the trade fills |
| **same duration, different spelling** (`60Minute` alongside `execute_on=1Hour`) | wrongly shifted as if coarser — a bar stale | recognised as equal duration, not shifted |

The first covers the "Multi-Timeframe" and "Hierarchical" patterns in `docs/guides/sampler.md`.
Coarse-only configs, single-timeframe configs, and the replay path are unaffected.

## The convention this rests on

The env fills at **the last close in `obs[execute_on]`** — the observed bar's close, not its
label. Verified by intercepting `_execute_trade_if_needed` at its call site (with
`execute_on=5Minute`, the steps labelled minute 20 and 25 fill at minutes 24 and 29), and now
pinned by `TestExecutionPriceConvention`.

That was previously implicit, and getting it backwards is expensive in both directions: read
the label as the decision instant and the aggregation looks like it leaks a bar when it does
not; read it a bar late and real lookahead goes unnoticed.

## Defect 1 — every timeframe now ends at the same instant

The resample loop shifts only `tf > execute_on` to END labels, and `_obs_indices` searched
every timeframe with the exec bin's **start** label. The `execute_on` frame lands correctly on
that key only by accident — its bin is already aggregated through the bin's end. A finer index
has real per-bar granularity and got the bar that merely *starts* there, so a 1Minute window
sat 59 minutes behind the 1Hour window sampled beside it.

Finer frames are now looked up at **the last bar that CLOSES at or before the bin ends**
(`exec_ts + exec_period − fine_period`). Not the last bar that *starts* inside it: pandas
aggregates a bar across its whole span, so a 7-minute bar starting at minute 70 closes at 76,
past a bin ending at 75.

`get_observation()` ran its own search without this rule, so it disagreed with the precomputed
path — reachable through the exhaustion fallbacks in `sequential.py` and `onestep.py`, where a
truncated episode bootstraps its value estimate. Both paths now share `_search_ts`.

## Defect 2 — timeframes compare by duration, not spelling

`TimeFrame` orders on duration but compares equal on `(value, unit)`. Those are deliberately
different keys: `60Minute` and `1Hour` produce different observation keys, and
`parse_timeframe_string` already warns that models do not transfer between them.

`@total_ordering` cannot express that. It derives `__gt__` as `not (self < other or self ==
other)`, so for equal-duration spellings neither branch held and **both** directions returned
`True`. All four comparisons are now explicit on duration; the decorator is gone. `__eq__` and
`__hash__` are unchanged — duration-based equality would disagree with the `obs_key_freq()`
identity the sampler's dicts actually run on.

The slip had **three** sites, not the one #282 names: the anti-lookahead shift, the
`max_offset` accounting (which discarded leading data for frames that were never shifted), and
a third introduced by this PR itself and caught in round 3.

## Also

A tz-aware index with `execute_on >= 1Day` is now refused. `Day` is not a Tick offset, so
tz-aware daily bins follow local midnight and are 23h/25h wide across DST while the search
bound uses a fixed 24h — reaching an hour past the bin's real close, which is lookahead rather
than staleness. The guard is deliberately broader than the hazard: it also rejects fixed-offset
zones like UTC that are provably safe. Narrowing it means enumerating zone transitions against
the data's span, which is more machinery than a case with no current user, and the error names
the fix. `Minute`/`Hour` are Tick offsets, DST-immune, and explicitly still allowed.

## Testing

`2395 passed, 25 skipped`. Every assertion added here was verified by mutation — reverted, run,
and confirmed to fail the right cases and only those:

| mutant | caught by |
|---|---|
| revert the `get_observation` parity fix | `test_both_lookup_paths_agree_for_a_finer_frame` (sole detector) |
| "snap the fine grid to the bin" — safe but stale | the 3 non-divisor straddling cases; divisor correctly passes |
| remove the tz guard | the 2 `Day` cases; `Hour`/`Minute` correctly pass |
| `tf < execute_on` → `tf != execute_on` | the coarse-freshness bound in the three-timeframe test |
| restore the old `bin_end − 1ns` bound | the 3 non-divisor cases (6/30/20 min of lookahead) |

Three rewritten tests in `TestSamplerNoFutureLeakage` previously bounded on the bar's **label**
and so read this correction as a regression, reporting "Future leakage detected". They now bound
on the decision instant. Review confirmed by mutation that this made them *stronger*: the old
one-sided label bound could only detect data arriving too early, and was structurally blind to
the staleness this PR fixes.

## Review

Three rounds, all four reviewers each round, per CLAUDE.md. Each round found a defect in the
previous round's fix:

- **Round 1** — my defect-1 fix introduced real lookahead for non-divisor fine timeframes
  (measured: 6, 30 and 20 minutes). Only divisor pairs were tested, so nothing caught it.
- **Round 2** — the parity fix was shipped untested (reverting it survived all 712 tests), and
  the bound asserted safety but not freshness, so a stale-but-safe implementation passed.
- **Round 3** — a third spelling-vs-duration site, unpinned; and `docs/guides/sampler.md` still
  described the leaky bound that round 1 replaced.

Round 3's fixes changed no production logic — tests, docs and one docstring only — so the chain
ends there rather than by assertion.

## Follow-up filed, deliberately not here

**#320** — the coarse-timeframe END shift uses a fixed `pd.Timedelta`, so a Day-or-longer
*observation* frame on a tz-aware index becomes visible an hour before it closes. Same DST
hazard, mirrored to the observation side, and it predates this branch (#10).

### DIFF
--- a/docs/guides/sampler.md
+++ b/docs/guides/sampler.md
@@ -107,10 +107,16 @@ Execute at:              ↑         ↑         ↑
                         t=5      t=10      t=15
 ```
 
-At t=10 (executing on 5-minute bar B):
-- **1-minute data**: Last 12 bars (from recent history)
-- **5-minute data**: Bar A (completed at t=5)
-- **15-minute data**: Bar X (completed at t=0)
+Each execution step is labelled by its bin's **start**, but the trade fills at that
+bin's **close**. So the step labelled t=10 spans [10, 15) and executes at minute 14 —
+verified by intercepting the executed price, and pinned by
+`tests/envs/offline/test_sequential.py::TestExecutionPriceConvention`.
+
+Every observation window is therefore as-of minute 14, the instant money changes hands:
+
+- **1-minute data**: last 12 bars, ending at minute 14
+- **5-minute data**: bar C [10, 15), whose close *is* the execution price
+- **15-minute data**: bar X [0, 15) only once it has closed; until then, the previous one
 
 ### Lookahead Bias Prevention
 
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps review this PR

PR #324, opened 2026-08-11
test(envs): give the suite a fixture that can actually gap (#315)

### LINKED ISSUE #315: No test fixture can construct a price gap: every conftest dataset sets open = previous close

## The blind spot

Every OHLCV fixture in `tests/conftest.py` builds opens with:

```python
open_prices = np.roll(close_prices, 1)
```

so `open[i] == close[i-1]` **exactly**, for `sample_ohlcv_df`, `trending_up_df`, `trending_down_df` and the rest. A price gap is by definition `open[i] != close[i-1]`. The entire suite therefore cannot construct one.

This is why #280 (stops filling at the bracket instead of the open on a gapping bar) survived ~2,300 tests, in four independent engines, indefinitely.

It is also why the fixture that existed specifically to cover this did not. `price_gap_df` ended with:

```python
# Make the gap visible in open/close
open_prices[50] = initial_price      # == close[49]
open_prices[150] = initial_price * 0.90   # == close[149]
```

The line labelled "make the gap visible" set the open equal to the previous close, which is the definition of *no* gap. Its only consumer, `TestSLTPPriceGaps::test_gap_triggers_stop_loss`, asserted just that the position had closed, so it passed under every fill-price mutation. Both were deleted in #311 and replaced by `tests/envs/offline/fundamental/test_gap_fills.py`, which builds real gaps.

## What is still missing

`test_gap_fills.py` pins the rule directly, at one config (leverage 2, one SL level, fee 0, slippage 0, fractional). What it does *not* do is give the rest of the suite the ability to see gaps at all. In particular the scalar-vs-vectorized equivalence harness (`test_vec_sltp_scalar_equivalence.py`) is blind to this whole class: reverting the stop rule in *both* engines leaves all 92 of its cells green, because it only detects divergence *between* engines, not a shared wrong answer.

## Proposal

Add a gapping fixture to `tests/conftest.py` (jump gaps at a known fraction of bars, both directions) and parametrize the existing equivalence tests over it. Verified during #311 review that both engines agree to 1.8e-12 across leverage {1,5} x fee {0, 0.001} x lock {on, off} x trade_mode {fractional, notional} on gappy data, so this should pass as-is — it is a guard against future regressions, not a bug report.

Deliberately left out of #311 to keep a fill-pricing fix from turning into a shared-fixture change.

BY571:
Closes #315. Fallout from #311.

## The blind spot

Every OHLCV fixture in `tests/conftest.py` builds `open = np.roll(close, 1)`, so `open[i] == close[i-1]` **exactly**. A price gap is by definition `open[i] != close[i-1]`. No fixture in the repo could construct one — which is why the gapped-stop mispricing in #280 survived the whole suite, in four independent engines.

## The fixture

`gappy_ohlcv_df` — the same random walk, with periodic 2% jumps in alternating directions.

**Gap size is bounded on both sides**, and that is the non-obvious part:

- **Large** against the walk's own 0.1%-per-bar volatility, so a gap clears a bracket. Below ~0.3% gaps stop clearing it at all (0.2% → zero gapped fills).
- **Small** against the liquidation band at the highest leverage under test — at 25x that band is `1/25 - 0.004 maintenance = 3.6%`. First liquidations appear at ~3%; my first attempt at 6% degenerated a cell into a liquidation test and tripped the harness's own vacuity guard.

## Where it is wired, and why *this* justification

Two cells on the scalar-vs-vectorized SLTP equivalence harness.

The obvious justification is wrong, and review caught it: this is **not** needed to catch a one-engine drift of the gap rule — `test_gap_fills.py` pins that directly. What only gappy data reaches is the **downstream consumers** of a gapped fill price. Fee, margin and PnL are all computed off `exec_price`:

| | fee mutant (charge the exit fee against the bracket, PnL fix intact) |
|---|---|
| `test_gap_fills.py` | passes — runs at `transaction_fee=0.0`, so no consumer is observable |
| existing equivalence grid | passes — has fees, but never gaps, so `exec_price == bracket` and the mutation is inert |
| **these 2 cells** | **fail** |

Measured: 2 failed, 754 passed across offline + replay.

## Two cells, not nine

The first draft added a `continuous`/`gappy` fixture axis over the whole 12-cell grid. Mutation testing across 11 gap mutations showed `trade_mode` is pure duplication on gappy data and leverage nearly so — `lev25 × notional` alone kills every one. Only the lock axis and the presence of shorts change what is reached. A separate 2-cell test also leaves the existing grid untouched.

The continuous arm stays, and now has a reason on record: max `hold_counter` is 26 there versus 6 on gappy data (2% gaps stop positions out fast), and two hold-counter saturation mutants are killed **only** by continuous cells.

## Scope

This does not catch a revert of **both** engines — the harness only detects divergence between them, so a shared wrong answer is invisible here by construction (measured 18/18 green). That case is pinned by `test_gap_fills.py`.

## Testing

Tests only. `test_vec_sltp_scalar_equivalence.py` + `test_gap_fills.py`: 64 passed, 3 skipped. Offline + replay: 756 passed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -71,18 +71,20 @@ def sample_ohlcv_df():
 def gappy_ohlcv_df():
     """Random walk whose bars periodically GAP: open[i] != close[i-1].
 
-    Every other OHLCV fixture in this file builds `open = np.roll(close, 1)`, so
-    `open[i] == close[i-1]` exactly and a bar can never open beyond a bracket set off the
-    previous close. That is a structural blind spot, not an oversight in any one test: it
-    is why the gapped-stop mispricing in #280 survived the whole suite in four separate
-    engines (#315).
+    Every other OHLCV fixture in this file has `open[i] == close[i-1]` exactly, so a bar
+    can never open beyond a bracket set off the previous close. That is a structural blind
+    spot, not an oversight in any one test: it is why the gapped-stop mispricing in #280
+    survived the whole suite in four separate engines (#315).
 
     Gap size is bounded on both sides. It must be large against the walk's own volatility
     (0.1% per bar) so a gap clears a bracket outright rather than grazing it, and small
     against the liquidation band at the highest leverage under test -- at 25x that band is
-    ~4%, and a gap wider than it liquidates the position before any bracket can fire,
-    which degenerates the cell into a liquidation test. 2% satisfies both. Direction
-    alternates so long and short brackets are each exercised.
+    1/25 - 0.004 maintenance = 3.6%, and a gap wider than it liquidates the position
+    before a bracket can fire, degenerating the cell into a liquidation test.
+
+    2% sits between measured cliffs: below ~0.5% gaps stop clearing the bracket, and at
+    ~3% the first liquidations appear. Direction alternates so long and short brackets are
+    each exercised.
     """
     rng = np.random.default_rng(20260811)
     n = 1440
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -82,9 +82,10 @@ def gappy_ohlcv_df():
     1/25 - 0.004 maintenance = 3.6%, and a gap wider than it liquidates the position
     before a bracket can fire, degenerating the cell into a liquidation test.
 
-    2% sits between measured cliffs: below ~0.5% gaps stop clearing the bracket, and at
-    ~3% the first liquidations appear. Direction alternates so long and short brackets are
-    each exercised.
+    2% sits between measured cliffs: gaps stop clearing the bracket below ~0.3% (0.2%
+    yields zero gapped fills, 0.5% already degrades to 30 of 51), and the first
+    liquidations appear at ~3%. Direction alternates so long and short brackets are each
+    exercised.
     """
     rng = np.random.default_rng(20260811)
     n = 1440
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -68,6 +68,49 @@ def sample_ohlcv_df():
 
 
 @pytest.fixture
+def gappy_ohlcv_df():
+    """Random walk whose bars periodically GAP: open[i] != close[i-1].
+
+    Every other OHLCV fixture in this file builds `open = np.roll(close, 1)`, so
+    `open[i] == close[i-1]` exactly and a bar can never open beyond a bracket set off the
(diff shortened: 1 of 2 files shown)


BY571:
@CharlieHelps please review.

Adds `gappy_ohlcv_df` to `tests/conftest.py` — the first fixture in the repo where `open[i] != close[i-1]`. Every existing one uses `np.roll(close, 1)`, so no bar could open beyond a bracket; that is why #280 survived the suite in four engines.

1. **Gap sizing is load-bearing.** 2% is chosen to clear the ±0.5% brackets while staying inside the ~4% liquidation band at 25x. My first try at 6% liquidated every position before a bracket fired, degenerating the cells into liquidation tests. Is 2% robust, or is it wedged between two constraints that a future config change would break silently?
2. **Scope claim.** I claim this catches a *one-engine* drift but explicitly NOT a both-engine revert (the harness only sees divergence between them). Verified: vectorized-only revert → 9 failures, all gappy-arm, 0 continuous-arm. Please check I have not overclaimed.

BY571:
Substantially reduced after round 2 — please re-review head `06a27b4` rather than the original diff.

Two changes worth flagging:
1. **9 gappy cells → 2.** Mutation testing across 11 gap mutations showed `trade_mode` is pure duplication on gappy data and leverage nearly so; `lev25 × notional` alone kills every one. It is now a separate 2-cell test instead of a fixture axis over the 12-cell grid, so that grid is untouched.
2. **The justification I gave was refutable.** I claimed this catches a one-engine drift of the gap rule — `test_gap_fills.py` already does that. The real unique value is *downstream consumers* of the gapped fill price: a mutant charging the exit fee against the bracket (PnL fix intact) kills exactly these 2 cells and nothing else in offline/replay. `test_gap_fills` runs at zero fee so it cannot see fee consumers; the grid has fees but never gaps.

Please sanity-check claim 2 — it is the load-bearing one now.

BY571:
@CharlieHelps the stale "6%" comment you flagged is already gone at head `06a27b4` — round 2 restored the 12-cell grid to its original form entirely, taking that comment with it, and moved the gappy coverage into a separate 2-cell test. `grep -n "6%"` over both changed files returns nothing.

Please re-review the current head; the diff is materially smaller than the one you approved.

PR #325, opened 2026-08-11
refactor(envs): delete OneStep's forks of the parent's exit logic (#316)

### LINKED ISSUE #316: OneStep re-forks the SLTP trigger rule under a near-identical name (_check_sltp_triggers vs _check_sltp_trigger)

## The duplication

- `SequentialTradingEnvSLTP._check_sltp_trigger` (singular) — `torchtrade/envs/offline/sequential_sltp.py`
- `OneStepTradingEnv._check_sltp_triggers` (plural) — `torchtrade/envs/offline/onestep.py:441`

Two implementations of the same intrabar SL/TP trigger rule, distinguished only by a trailing `s`. #311 had to apply the identical gap-aware fill in both, in three separate places, and the OneStep copy shipped with **zero** test coverage — reverting all three of its call sites left the whole suite green until #311 added cells for it.

This is exactly the case `CLAUDE.md` calls out: *"a re-forked copy that has not drifted YET is still a bug waiting to happen"*. Unlike the scalar-vs-vectorized pair, which is pinned by a shared equivalence harness, nothing forces these two to agree.

Confirmed during review that they are currently behaviourally identical, so this is a consolidation, not a bug fix.

## Related: both copies contain unreachable conditions

```python
if (open_price <= self.stop_loss or
    low_price  <= self.stop_loss or
    close_price <= self.stop_loss):
```

Given the OHLC invariant `low <= min(open, close)`, terms 1 and 3 are subsumed by term 2 — the whole condition reduces to `low_price <= self.stop_loss`. Mirrored through `high` for shorts. Present at `onestep.py:455-457` and `:485-487`, and in the parent at `sequential_sltp.py:267-270` and `:293-296`.

Reducing them is what makes the two implementations visibly identical, so it is best done as part of the merge rather than separately.

## Scope

Merge to a single implementation on the parent, delete the OneStep fork, and let `tests/envs/offline/fundamental/test_gap_fills.py` (which now covers both) guard the merge.

BY571:
Closes #316. Fallout from #311.

## The duplication

OneStep carried **two** forks of the parent's exit logic:

| OneStep | parent | justification given |
|---|---|---|
| `_check_sltp_triggers` | `_check_sltp_trigger` (one letter apart) | "to return trade_info directly" |
| `_check_liquidation_in_rollout` | `_check_liquidation` | *the identical sentence* |

Neither had drifted — but they are why the gapped stop in #280 had to be fixed in **two separate scalar places**, and why the OneStep half shipped with zero coverage until #311 added cells for it. `CLAUDE.md` names this case: *"a re-forked copy that has not drifted YET is still a bug waiting to happen."*

The second fork was found in round 2, twenty lines above the first. It also re-forked the enable guard by hand as `if self.leverage == 1` instead of the parent's `has_liquidation` property.

## The consolidation

The parent grows one method — `_apply_bar_exits(ohlcv)` — that owns the **entire** exit surface: liquidation-first priority, bracket detection, gap-aware fill pricing, and booking. Both the parent's `_step` and OneStep's `_rollout` call it. OneStep now defines none of it.

That seam also inlines the `_sltp_execution_price` helper the first draft added, which review correctly identified as a one-call-site helper once the delegation moved.

## Equivalence — verified, not assumed

- The two SLTP rules are identical over **300k well-formed bars**; they can only part on a bar whose `close` falls outside `[low, high]`, which the sampler cannot produce (its aggregation preserves the ordering, and its synthesized empty bars set `open=high=low=close`).
- SL-before-TP priority preserved on **3,780/3,780** bars where both could fire.
- **1,380 OneStep episodes** (leverage 1/3/10 × 12 seeds × 5 starts × every action, gap-heavy data) exercising **600 stops + 600 take-profits, 564 gapped**: reward and balance **byte-identical** to `main`, max delta **0.0**.

## The durable fix

A structural guard — `test_onestep_does_not_refork_the_exit_rules`, 5 cells — asserting OneStep **inherits** each of `_apply_bar_exits`, `_check_liquidation`, `_execute_liquidation`, `_check_sltp_trigger`, `_execute_sltp_close`. Behavioural equivalence cannot catch a re-fork that has not drifted yet, which is exactly what both of these were. The live envs already carry this shape of guard in `tests/envs/test_live_env_base.py`; its absence on the offline side is how these survived.

It asserts *identity against the parent*, not `name not in vars(...)` — the absence form passes for any name matching nothing, so a renamed parent method would leave every cell green.

## Two dead-assertion tests repaired

Both on paths this PR touches:

- **`test_futures_liquidation_during_rollout`** asserted only `done`, trivially true for every OneStep episode. A zero balance doesn't discriminate either — measured, the position wipes out with liquidation disabled too. What liquidation does is *cap* the loss: it fires at bar 107 for −2.19, where riding the trend to bankruptcy runs to bar 135 for −3.30. Now bounded on both sides, and it fails when the check is stubbed.
- **`test_transaction_costs_included`** had **zero assertions**. Now compares against the same episode run fee-free — and needs `sampler.seek(0)`, because OneStep forces `random_start` and the first repair passed on start-bar randomness rather than on fees.

## Testing

`tests/envs/offline` + `tests/envs/replay`: **761 passed, 3 skipped**.

### DIFF
--- a/tests/envs/offline/test_onestep.py
+++ b/tests/envs/offline/test_onestep.py
@@ -499,6 +499,30 @@ class TestOneStepIntegration:
 class TestOneStepRegression:
     """Regression tests for known OneStep issues."""
 
+    @pytest.mark.parametrize("method", [
+        "_check_sltp_trigger",     # intrabar SL/TP detection
+        "_sltp_execution_price",   # which price a fired bracket fills at
+        "_execute_sltp_close",     # booking the close
+    ])
+    def test_onestep_does_not_refork_the_bracket_rules(self, method):
+        """OneStep must inherit these, never redefine them (#316).
+
+        It used to carry its own copy of the detection rule under
+        _check_sltp_triggers -- one letter from the parent's _check_sltp_trigger -- along
+        with its own fill pricing. The copies had not drifted, but they meant #280 had to
+        be fixed in two scalar places and the OneStep half shipped untested.
+
+        A structural guard rather than a behavioural one: the live envs have the same
+        shape of check in tests/envs/test_live_env_base.py, and its absence here is how
+        the fork survived. Behavioural equivalence cannot catch a re-fork that has not
+        drifted yet.
+        """
+        assert method not in vars(OneStepTradingEnv), (
+            f"OneStepTradingEnv redefines {method}. Bracket detection, pricing and "
+            "booking live on SequentialTradingEnvSLTP so a fill rule cannot be fixed in "
+            "only one of them -- delegate instead of re-forking."
+        )
+
     def test_done_flag_always_true_after_step(self, onestep_env):
         """Done flag should always be True after step (one-step setting)."""
         td = onestep_env.reset()
--- a/tests/envs/offline/test_onestep.py
+++ b/tests/envs/offline/test_onestep.py
@@ -11,7 +11,11 @@ Uses parametrization to test both trading modes with maximum code reuse.
 import pytest
 import torch
 
-from torchtrade.envs.offline import OneStepTradingEnv, OneStepTradingEnvConfig
+from torchtrade.envs.offline import (
+    OneStepTradingEnv,
+    OneStepTradingEnvConfig,
+    SequentialTradingEnvSLTP,
+)
 from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
 from tests.conftest import simple_feature_fn, validate_account_state
 
@@ -223,12 +227,18 @@ class TestOneStepRolloutSimulation:
         action_td["action"] = torch.tensor(4)  # Long 100%
         next_td = env.step(action_td)
 
-        # Should terminate (liquidation or natural end)
-        assert next_td["next"]["done"].item()
-
-        # Reward should reflect loss
-        reward = next_td["next"]["reward"]
-        # Typically negative due to downtrend
+        # `done` is trivially true for every OneStep episode, and so is a zero balance --
+        # measured, the position wipes out either way. What liquidation actually does is
+        # CAP the loss by exiting early: it fires at bar 107 for a terminal reward of
+        # -2.19, where a position that rides the same downtrend to bankruptcy runs to bar
+        # 135 for -3.30. The upper bound is what makes this test fail when the liquidation
+        # check is stubbed out; the lower bound keeps it honest that a wipeout happened.
+        assert env.position.position_size == 0.0
+        reward = next_td["next"]["reward"].item()
+        assert -3.0 < reward < -1.0, (
+            f"reward {reward:.3f}: a 20x long into a downtrend should be liquidated, "
+            "capping the loss short of what riding the trend to bankruptcy costs"
+        )
         env.close()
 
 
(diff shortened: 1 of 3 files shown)


BY571:
@CharlieHelps please review — this one touches money code, unlike the last three.

It deletes OneStep's fork of the SLTP trigger rule (`_check_sltp_triggers`, one letter from the parent's `_check_sltp_trigger`) and makes it delegate. Fill-price selection also moves onto the parent as `_sltp_execution_price`, since the parent's `_step` had its own copy of the same ternary — that duplication is what forced #280 to be fixed twice in scalar code.

1. **Equivalence.** I brute-forced both rules over 9,604 well-formed OHLC × bracket × direction combinations: 0 mismatches. They differ only on malformed candles (`high < max(open, close)`). Is treating that as undefined input the right call, or does some code path legitimately feed malformed bars?
2. **Priority ordering.** The parent checks SL before TP (deliberate pessimism). Please confirm delegation preserves that, and that liquidation still outranks brackets in `_rollout`.

Behavioural evidence: 1,380 OneStep episodes across leverage 1/3/10 with 1,200 bracket exits (564 gapped) — reward and balance byte-identical before/after, max delta 0.0.

BY571:
Scope grew materially after round 2 — please re-review head `c4f1975`.

Round 2 found the fork's **twin** twenty lines above the one I deleted: `_check_liquidation_in_rollout` re-implements the parent's `_check_liquidation`, with the *identical* docstring justification the deleted `_check_sltp_triggers` gave. Shipping a PR titled "delete OneStep's fork" while leaving its twin would have been the overclaiming this PR is meant to end.

Both are gone. The seam moved: the parent grows `_apply_bar_exits` owning the whole exit surface (liquidation priority → bracket detection → gap-aware pricing → booking), used by both `_step` and `_rollout`. That inlines the `_sltp_execution_price` helper the first draft added.

Still byte-identical to main: 1,380 OneStep episodes, 1,200 bracket exits (564 gapped), max reward/balance delta 0.0.

Worth your scrutiny: (1) `_apply_bar_exits` is now the single place liquidation-outranks-brackets is expressed for the scalar engines — please confirm the vectorized engine still agrees; (2) the structural guard asserts identity against the parent for 5 methods — is that list complete, or is there other shared exit/money logic OneStep could still re-fork?

BY571:
@CharlieHelps resolved at head `5f8b264` — thank you, that was a real gap my reviewers missed. They verified the sampler's *aggregation* preserves well-formed OHLC; they did not check that *ingestion* validates it. Nothing does.

I tried your first option, validating at ingestion. It fails **30 existing tests** — a substantial part of the suite builds impossible candles. That's a real finding but too big for this PR, so I filed it as #326 with the diff that produces it.

So I took your second option, and made it the surviving rule rather than a special case. Where the two rules differed, the parent deferred its close test until *after* the take-profit checks, so a malformed bar could be handed to the take-profit; OneStep's form tests open/extreme/close together with the stop first. **OneStep's is the strictly more pessimistic one**, which is what this module documents itself as converging on — so the consolidation can now only ever be more conservative, never less.

Verified against the deleted fork over **all 32,768** open × high × low × close × sl × tp × direction combinations *including malformed*: **0 mismatches**. And the 1,380-episode OneStep probe is still byte-identical to main (max reward delta 0.0).

Please re-review.

PR #327, opened 2026-08-11
chore(envs): drop 10 hardcoded truncated writes that #272 made no-ops (#313)

### LINKED ISSUE #313: Live envs: drop the 10 hardcoded truncated=False writes, now no-ops after #272
## Cleanup

Every one of the 10 live-exchange `_step()` methods writes a hardcoded `truncated=False`:

```python
next_tensordict.set("done", torch.tensor([done], dtype=torch.bool))
next_tensordict.set("truncated", torch.tensor([False], dtype=torch.bool))
next_tensordict.set("terminated", torch.tensor([done], dtype=torch.bool))
```

Once #272 declares `truncated` in `TorchTradeBaseEnv.full_done_spec`, those writes are provably no-ops.

## Evidence

Two independent verifications during the #272 review:

- **Traced**: `EnvBase._complete_done` runs after every `_step`. With `done` and `terminated` both present, neither infer-from-sibling branch fires, and the missing `truncated` falls through to `data.set("truncated", item.zero(leading_dim))` — where `item` is the `Categorical(2, dtype=torch.bool, shape=(1,))` from the shared spec. That produces a `(1,)` all-`False` bool tensor: byte-for-byte what the hardcoded line writes.
- **Mutated**: deleting the write from one live env and running the whole envs suite → 2074 passed, zero failures.

This is not a new dependency either. None of the 10 `_reset()` methods write `truncated` at all — they already rely on exactly this auto-fill. Deleting the `_step` writes would make `_step` consistent with what `_reset` already does.

## Scope

- [ ] Delete the `truncated` write from all 10 live `_step()` methods
- [ ] Confirm `check_env_specs` and `test_a_collector_batch_carries_truncated` still pass (they are the backstop that makes this safe)

Note that after deletion there is no test pinning the absence — the write and its removal are behaviourally identical, so a test would be pinning code that provably cannot change behaviour. That is the argument for deleting rather than testing.

## Why not in #272

That PR's production diff is exactly two files. Reopening `_step` in 10 live env files would reintroduce production surface into a review that closed clean, and mix a "declare the spec" bugfix with an unrelated "remove now-redundant write" cleanup in one history. Both reviewers who looked at it independently recommended a follow-up.

Depends on #272 landing first — the deletion is only safe once the shared spec declares the key.

BY571:
Closes #313. Follow-up to #272/#317.

## The cleanup

All 10 live-exchange `_step()` methods wrote:

```python
next_tensordict.set("truncated", torch.tensor([False], dtype=torch.bool))
```

Since #272 declared `truncated` in the shared `full_done_spec`, `EnvBase._complete_done` fills it after every step. Verified the auto-fill is byte-identical to the deleted line:

```
auto-filled : value=[False] dtype=torch.bool shape=(1,)
deleted line: value=[False] dtype=torch.bool shape=(1,)
IDENTICAL: True
```

The 10 `_reset()` methods never wrote it and have always relied on exactly this path, so `_step` is now consistent with `_reset`.

## One nuance the issue got wrong

It called the writes *"provably no-ops"*. That holds for the real `step()` path but **not** for `_step()` called directly — and two tests did exactly that while asserting `truncated` was present:

- `test_torch_env.py::test_step_contains_done`
- `test_torch_env_sltp.py::test_step_contains_reward_and_done`

They were pinning the hardcoded write rather than the spec that makes the collector work. Both now go through `step()`. Asserting the done family on `_step`'s raw output tests an implementation detail — emitting it is `EnvBase`'s job, not `_step`'s.

## Testing

`tests/envs`: **2111 passed, 19 skipped**. `check_env_specs` and the collector-truncated backstop from #272 are what make this safe, and both still pass.

### DIFF
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -271,14 +271,19 @@ class TestAlpacaTorchTradingEnvStep:
         assert "reward" in td_out.keys()
 
     def test_step_contains_done(self, env):
-        """Test that step returns done flag."""
-        env.reset()
-        td_in = TensorDict({"action": torch.tensor(1)}, batch_size=())
-        td_out = env._step(td_in)
+        """A stepped env emits the whole done family (#272).
 
-        assert "done" in td_out.keys()
-        assert "terminated" in td_out.keys()
-        assert "truncated" in td_out.keys()
+        Goes through step(), not _step(): truncated is filled by EnvBase from the done
+        spec, so asserting it on _step's raw output would pin a hardcoded write rather
+        than the spec that makes the collector work (#313).
+        """
+        td = env.reset()
+        td["action"] = torch.tensor(1)
+        nxt = env.step(td)["next"]
+
+        for key in ("done", "terminated", "truncated"):
+            assert key in nxt.keys(), f"{key} missing from the emitted step"
+        assert nxt["truncated"].dtype is torch.bool
 
     def test_step_buy_action(self, env):
         """Test buy action (action=2)."""
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -159,9 +159,15 @@ class TestAlpacaTorchTradingEnvReset:
             trader=mock_trader,
         )
 
+    def test_step_emits_the_whole_done_family(self, env):
+        from tests.envs.base_exchange_tests import (
+            assert_the_step_emits_the_whole_done_family as assert_done_family,
+        )
+        assert_done_family(env)
+
     def test_check_env_specs_passes(self, env):
         """check_env_specs compares the emitted step against every declared spec;
-        catches a done spec missing the truncated key each _step writes (#272)."""
+        catches a done spec missing a key the emitted step carries (#272)."""
         with patch.object(type(env), "_wait_for_next_timestamp"):
             check_env_specs(env)
 
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -167,7 +167,8 @@ class TestAlpacaTorchTradingEnvReset:
 
     def test_check_env_specs_passes(self, env):
         """check_env_specs compares the emitted step against every declared spec;
-        catches a done spec missing a key the emitted step carries (#272)."""
+        the done family comes from the spec on both sides here, so a narrowed done spec
+        is NOT what this catches -- see assert_the_step_emits_the_whole_done_family."""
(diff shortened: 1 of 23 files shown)


BY571:
@CharlieHelps please review.

Deletes 10 hardcoded `truncated=False` writes from the live `_step()` methods, now that #272 declares `truncated` in the shared `full_done_spec` and `EnvBase._complete_done` fills it. Verified byte-identical: `[False]`, `torch.bool`, shape `(1,)`.

Worth your scrutiny — the issue claimed these were "provably no-ops" and that was **not quite right**: two tests called `_step()` directly and asserted `truncated` was present, so deleting the writes broke them. I moved those tests onto `step()` on the grounds that emitting the done family is `EnvBase`'s job, not `_step`'s. Is that the right call, or is there a legitimate reason a caller would invoke `_step` directly and expect the full done family?

BY571:
@CharlieHelps please review.

Deletes 10 hardcoded truncated=False writes from the live _step() methods, now that https://github.com/TorchTrade/torchtrade/issues/272 declares truncated in the shared full_done_spec and EnvBase._complete_done fills it. Verified byte-identical: [False], torch.bool, shape (1,).

Worth your scrutiny — the issue claimed these were "provably no-ops" and that was not quite right: two tests called _step() directly and asserted truncated was present, so deleting the writes broke them. I moved those tests onto step() on the grounds that emitting the done family is EnvBase's job, not _step's. Is that the right call, or is there a legitimate reason a caller would invoke _step directly and expect the full done family?

BY571:
@CharlieHelps head is now `e9ea9a4` — the PR grew substantially after two review rounds; please review the current diff.

**Round 1 found this PR was a net loss as it stood.** While every live `_step` wrote `truncated` by hand, a narrowed done spec showed up in `check_env_specs` as `keys_in_real_not_in_fake`. Delete the writes and both real and fake rollouts lack the key, so `check_env_specs` can't see a narrowed spec at all — the #272 tripwire disarms itself. Same mutation: **fails on main, passes after the deletion**. Binance and binance-SLTP had zero residual coverage.

Fixed by adding `assert_the_step_emits_the_whole_done_family` to `base_exchange_tests.py`, wired into all 10 live env test files.

**Round 2 found three false prose claims I'd just written**, the sharpest being that my "sole source of the key" comment sits in the class shared by live *and* offline — and offline envs still write `truncated` meaningfully. All scoped correctly now.

Worth your scrutiny: the shared body absorbed 3 pre-existing duplicates (net −3 test functions) while asserting dtype+shape for all three done keys. Dropping `truncated` from the spec now fails 26 tests; narrowing its dtype fails 21.

PR #329, opened 2026-08-11
test(examples): kill the example, not the shell that started it (#312)

### LINKED ISSUE #312: test_examples: a timed-out example is orphaned, not killed — the flake aggravates itself
## Bug

`run_command` in `tests/test_examples.py:467` runs each example with `shell=True` and, on `TimeoutExpired`, calls `process.kill()`. That kills the `/bin/sh` it forked — **not** the training run the shell started. The child survives as an orphan.

Verified empirically: after `p.kill()`, the shell died and the child kept running.

## Why it matters

The failure is self-aggravating. A timed-out example leaves a full training run consuming CPU for the remainder of the suite — under exactly the heavy-load condition that caused the timeout in the first place. One trip can cascade into further timeouts.

Observed in practice: under load from concurrent work, `test_example_commands[grpo_online-...]` exceeded the 300s budget in a full-suite run that took 627s overall. The same test passes in **23s** in isolation, and the suite is green in 115s when the machine is quiet — 13x headroom, so this is load sensitivity rather than a slow example.

## Fix

Either:

```python
process = subprocess.Popen(..., shell=True, start_new_session=True)
...
except subprocess.TimeoutExpired:
    os.killpg(os.getpgid(process.pid), signal.SIGKILL)
```

or at minimum `process.communicate()` after the kill so the child is reaped.

Dropping `shell=True` in favour of `shlex.split` would also work — the module already imports `shlex`.

## Context

Found while reviewing #272. Pre-existing, unrelated to that change: `tests/test_examples.py` is not in its diff, and the affected examples use offline envs whose done spec already declared `truncated` before it.

BY571:
Closes #312.

## The bug

`run_command` ran each example with `shell=True` and, on `TimeoutExpired`, called `process.kill()`. That kills the `/bin/sh` it forked — **not** the training run the shell started. (`sh` forks rather than execs for these commands; verified.)

Self-aggravating: a timed-out example leaves a full training run consuming CPU for the rest of the suite, under exactly the load that caused the timeout.

## The fix

- **`shlex.split` instead of `shell=True`** — none of the six commands contains a shell metacharacter, so the shell was pure overhead. Popen's child is now the example itself.
- **`start_new_session=True` + `os.killpg`** — reaches whatever the example spawned in turn. It is also what makes the `killpg` *safe*: without it `getpgid` returns pytest's own group and the kill takes down the whole test session.
- **`except BaseException`, not `TimeoutExpired`** — `start_new_session` removes the example from pytest's process group, so a terminal Ctrl-C no longer reaches it. Handling only the timeout would close one door and open another.
- **`wait()`, not `communicate()`** — `communicate` reads stdout to EOF, and a descendant that escaped the group kill holds that pipe open: 25s on a 3s timeout, unbounded against a long-lived escapee, with no `pytest-timeout` to stop it. CPython's own `subprocess.run` does `kill()+wait()`.

## What review changed

Two rounds, both of which found the fix incomplete in ways I had not:

1. **The first draft opened the abort door while closing the timeout one.** Measured: Ctrl-C leaked 0 processes before, 1 after.
2. **The reap could hang the suite indefinitely** — and the same `communicate()` made the new test half-vacuous, passing even with `killpg` reverted to `kill()`.
3. **A claim of mine was measurably false.** I justified `killpg` by saying `ParallelEnv` workers outlive their parent. They do not — killing the parent alone reaps every worker within 2s under both spawn and fork, because the worker blocks in `child_pipe.poll()` and the parent's death closes the last write end. `killpg` is still worth keeping; that just isn't why.

## Testing

Two tests, both mutation-verified. Writing the second caught the same vacuity trap one level down: my first version patched `communicate` to delegate to the real one with a timeout, which raised `TimeoutExpired` before reaching the `KeyboardInterrupt` — silently exercising the arm it was meant to be distinct from.

| mutation | result |
|---|---|
| `shell=True` + `process.kill()` (the original bug) | **fails** `grandchild N survived the timeout` |
| `killpg` → `process.kill()` | **fails** |
| `except BaseException` → `except TimeoutExpired` | **fails** |
| no reap after the kill | **fails** |

`tests/test_examples.py`: **23 passed** in 118s.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -467,7 +467,7 @@ EXAMPLE_COMMANDS = {
 
 def run_command(command: str, timeout: int = 300) -> int:
     """
-    Run a shell command and return the exit code.
+    Run an example command (deliberately with no shell) and return its exit code.
 
     Args:
         command: The command to run
@@ -486,9 +486,9 @@ def run_command(command: str, timeout: int = 300) -> int:
         argv[0] = sys.executable
 
     # No shell: with shell=True the child of Popen is /bin/sh, so killing it on timeout
-    # leaves the training run itself alive. start_new_session puts the example in its own
-    # process group, which is what makes the killpg below reach the run AND anything it
-    # spawned -- ParallelEnv workers outlive their parent otherwise (#312).
+    # leaves the training run itself alive -- the actual bug in #312. start_new_session
+    # puts the example in its own group so the killpg below reaches whatever it spawned
+    # in turn, whether or not that thing cleans up after its own parent.
     process = subprocess.Popen(
         argv,
         stdout=subprocess.PIPE,
@@ -504,10 +504,17 @@ def run_command(command: str, timeout: int = 300) -> int:
             print(f"Command failed with exit code {process.returncode}")
             print(stdout.decode() if stdout else "")
         return process.returncode
-    except subprocess.TimeoutExpired:
+    except BaseException:
+        # BaseException, not TimeoutExpired: start_new_session took the example out of
+        # pytest's process group, so a terminal Ctrl-C no longer reaches it. Catching only
+        # the timeout would close that door and open this one -- and aborting a slow run by
+        # hand is far more common than hitting the 300s budget.
+        #
+        # Not `finally`: on the success path communicate() has already reaped the child,
+        # and getpgid on a reaped pid raises ProcessLookupError, which would then crash
+        # every passing test.
         os.killpg(os.getpgid(process.pid), signal.SIGKILL)
-        # Reap it, or the timed-out example stays a zombie for the rest of the session.
-        process.communicate()
+        process.communicate()  # reap, or it lingers as a zombie for the session
         raise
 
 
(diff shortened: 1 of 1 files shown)


BY571:
@CharlieHelps please review.

`run_command` ran examples under `shell=True` and killed the shell on timeout, not the training run — so a timed-out example kept burning CPU for the rest of the suite, under exactly the load that caused the timeout.

Fix drops the shell (`shlex.split`; none of the six commands has a shell metacharacter) and adds `start_new_session=True` + `os.killpg`, which is the part that matters for TorchRL — `ParallelEnv` workers outlive their parent, so killing the Python process alone still leaks them.

Measured: old path leaks 2 processes past the kill, new path leaks 0. `tests/test_examples.py` 21 passed.

Worth your scrutiny: is `killpg` on a new session safe here, or could it ever reach the pytest process itself? And is dropping `shell=True` right, versus keeping it and killing the group?

BY571:
Two review rounds landed substantive changes — please review head `a9cc500` rather than the original diff. Notably: my `ParallelEnv` justification for `killpg` was **measurably false** (workers do not outlive a killed parent, under either start method), the first draft made Ctrl-C leak the example, and the reap could hang CI indefinitely. PR body rewritten with the corrected story.

PR #334, opened 2026-08-11
fix(envs): let a stop that price crosses first beat the liquidation (#300)

### LINKED ISSUE #300: SLTP envs: liquidation-beats-bracket is applied unconditionally, including where the bracket must fill first
Raised during the review of #299 (which pins the current behaviour). **Pre-existing, not a regression** — this is about whether the rule itself is right, not whether it is implemented as documented.

## The rule today

Both offline SLTP envs check liquidation before the bracket, unconditionally:

- `torchtrade/envs/offline/sequential_sltp.py::_step` — an `if/else` chain
- `torchtrade/envs/offline/vectorized_sequential_sltp.py::_apply_exit_checks` — the liquidation block zeroes the position first, so `has_position` excludes it

## Where that is sound, and where it is questionable

The env already documents a related pessimistic convention: **SL is checked before TP**. That one is unambiguously the right policy — SL and TP sit on *opposite* sides of entry, so a single bar's high and low genuinely cannot say which was touched first. Assuming the worse of two irreconcilable orderings is the only sound choice.

Liquidation vs a stop-loss on the same position is **not** the same situation. Both trigger on the same side of entry:

| case | bracket | liquidation | continuous path from entry | verdict |
|---|---|---|---|---|
| bracket nearer entry | SL 95 | 90.4 | must cross 95 before 90.4 | a broker fills the stop; **liquidation is arguably wrong** |
| bracket further out | SL 50 | 90.4 | must cross 90.4 before 50 | **liquidation correctly wins** |

The code does not distinguish these. In the first row the bar's own recorded low is enough to know the stop was reachable first, so the outcome is not a pessimistic choice between unknowns — it books an exit the data makes implausible, and the difference is large: at 10x with 10000 cash, liquidation leaves 400 where the stop would leave 5000.

The take-profit-vs-liquidation collision (a bar wicking both ways) *is* genuinely ambiguous and the current convention is right there.

## Options

1. Condition the precedence on which threshold the price reaches first, given the direction of travel from entry. Most correct, most work.
2. Keep it unconditional and document it explicitly as a pessimistic convention rather than a physical claim. Cheapest, and arguably fine for a backtest that wants a conservative bound.
3. Leave as-is.

Whichever way it goes, the tests added in #299 pin today's behaviour for all three quadrants (long-stop, short-stop, long-take-profit), so a deliberate change will surface as a clear test update rather than silent drift.

BY571:
Closes #300. Raised during the #299 review; pre-existing, not a regression.

## The rule today, and where it is wrong

Both offline SLTP engines take the liquidation unconditionally when a bar breaches it *and* a bracket.

Against a **take-profit** that is right, for the same reason SL beats TP: the two sit on **opposite** sides of entry, so a bar reaching both says nothing about the order. Assuming the worse of two irreconcilable orderings is the only sound choice.

A **stop-loss** is not that case:

| | bracket | liquidation | continuous path from entry | today | correct |
|---|---|---|---|---|---|
| stop inside liq distance | SL 95 | 90.4 | must cross 95 before 90.4 | liquidation | **stop** |
| stop beyond it | SL 50 | 90.4 | must cross 90.4 before 50 | liquidation | liquidation |

A stop sits on the **same** side as liquidation, so price cannot reach the further level without crossing the nearer one — and the bar's own extreme says which. Booking a liquidation where the stop was crossed first isn't pessimism, it's an outcome the data contradicts. At 10x that's **400 against 5000**, a 12x gap, in the *ordinary* configuration where a trader's stop sits inside their liquidation distance.

## The one exception

A bar that **opened** past the liquidation price crossed nothing on the way. The margin was gone before the bar began and no resting order could have worked first — so the liquidation stands. This is why the fix is a predicate rather than a bare comparison.

## Implementation

- Scalar: `_stop_is_reached_first` guards the liquidation branch in `_apply_bar_exits` (the single exit surface from #316).
- Vectorized: an **exemption mask** handed to the shared `_apply_liquidation`, rather than re-deriving the priority. The base env has no brackets and passes nothing, so it is unaffected.

## Testing

#299 deliberately pinned the old behaviour "so a deliberate change will surface as a clear test update rather than silent drift" — it did. Two of its four quadrants change answer; the two take-profit quadrants do not.

Mutation-checked, and two guards were unpinned until their own cell existed:

| mutation | result |
|---|---|
| scalar reverted to the old rule | 3 fail |
| **vectorized only** reverted (one-engine drift) | 1 fail |
| scalar drops the gapped-open guard | 1 fail *(0 before adding its cell)* |
| vectorized drops the gapped-open guard | 1 fail *(0 before adding its cell)* |

`tests/envs/offline` + `tests/envs/replay`: **775 passed, 3 skipped.**

### DIFF
--- a/tests/envs/offline/test_sequential_sltp.py
+++ b/tests/envs/offline/test_sequential_sltp.py
@@ -1270,12 +1270,11 @@ class TestLiquidationVsBracketOnADoubleBreachBar:
     sits on the same side as liquidation, so price cannot reach the further level without
     crossing the nearer one, and the bar's own extreme says which. Booking a liquidation
     where the stop was crossed first is not pessimism but an outcome the data contradicts:
-    400 instead of 5000 here.
+    400 instead of 5000 here, a 12.5x gap.
 
-    Nothing held this. A double-breach bar has existed since
-    TestSLTPRegression::test_liquidation_triggers_on_next_bar_futures, but no test
-    asserted anything that *distinguishes* the two exits, so swapping the checks left
-    the whole offline suite green. All four quadrants of direction x trigger are
+    Nothing held this before #298: double-breach bars existed in the suite, but no test
+    asserted anything that *distinguishes* the two exits, so swapping the checks left the
+    whole offline suite green. All four quadrants of direction x trigger are
     covered because each is invisible to the other three: a reorder scoped to one of
     them passes every test the others pin.
     """
--- a/tests/envs/offline/test_sequential_sltp.py
+++ b/tests/envs/offline/test_sequential_sltp.py
@@ -1278,9 +1278,6 @@ class TestLiquidationVsBracketOnADoubleBreachBar:
     the whole offline suite green. All four quadrants of direction x trigger are
     covered because each is invisible to the other three: a reorder scoped to one of
     them passes every test the others pin.
-
-    All four quadrants of direction x trigger are covered because each is invisible to
-    the other three: a reorder scoped to one of them passes every test the others pin.
     """
 
     def test_a_bar_that_gapped_past_liquidation_still_liquidates(self):
@@ -1302,6 +1299,8 @@ class TestLiquidationVsBracketOnADoubleBreachBar:
             f"a gapped-open bar must liquidate, not fill the stop -- got "
             f"{env.history.action_types}"
         )
+        # 400 is an #314 number: the liquidation books at liq_price even though the bar
+        # opened below it. When that is fixed this expectation moves with it.
         assert env.balance == pytest.approx(400.0)
 
     # The action tuple is spelled out rather than derived: create_sltp_action_map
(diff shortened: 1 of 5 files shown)


BY571:
@CharlieHelps please review — this changes money semantics in the offline SLTP engines.

A stop-loss sits on the **same side of entry** as liquidation, so price cannot reach the further level without crossing the nearer one. Booking a liquidation where the stop was crossed first is an outcome the bar's own extreme contradicts: 400 vs 5000 at 10x, in the ordinary configuration where a stop sits inside the liquidation distance. A take-profit is the opposite case (other side of entry, ordering unresolvable) and still loses to liquidation.

1. **Is the continuity argument sound for a backtest?** It assumes intrabar price moves continuously from entry. A bar that gapped past liquidation is excluded, but is there another path where both levels are reached without crossing the nearer one first?
2. **Is exempting lanes from the shared `_apply_liquidation` the right seam** for the vectorized engine, versus re-deriving priority in the SLTP subclass? The base env passes no exemption, so it is unaffected — but the shared method now has a parameter only one caller uses.

PR #337, opened 2026-08-11
fix(envs): count an unsettled position in the alpaca bankruptcy baseline (#284)

### LINKED ISSUE #284: Alpaca bankruptcy baseline reads cash before close_all_positions() proceeds settle
\`live/alpaca/base.py:98-104\`: construction calls \`close_all_positions()\` (submits market orders, does not block) and immediately reads \`account.cash\` as \`initial_portfolio_value\`. Constructed while holding $9k BTC + $1k cash → baseline = **1000**, so \`_check_termination\` (\`core/live.py:237-244\`) only fires below $100 instead of $1000.

The four futures bases read \`total_margin_balance\` and are immune.

Fix: poll until positions report closed (or read portfolio value including positions) before pinning the baseline.

BY571:
Closes #284.

## The bug

`AlpacaBaseTorchTradingEnv.__init__` calls `close_all_positions()` — which **submits market orders and returns**, it does not block — then immediately reads `account.cash` as `initial_portfolio_value`.

Reproduced against a trader whose close hasn't settled:

```
cash=1000.0  position value=9000.0
initial_portfolio_value = 10000.0   (cash-only gave 1000)
bankruptcy fires below   = 1000.0   (cash-only gave 100)
```

A 10× understatement of the account the check exists to protect.

## The deeper problem

Cash was never the right quantity. `_check_termination` compares against `_get_portfolio_value()`, which is **cash + position value** — so the baseline was measured in different units from the thing it bounds. *That* is why settlement timing could move it at all.

Using the env's own method makes both the same measurement by construction, and the fix no longer depends on the close having settled. The four futures bases already read an equity figure for exactly this reason (`bybit/base.py:89`: *"Bankruptcy baseline on total_margin_balance (equity)"*).

`_get_portfolio_value`'s own docstring already names the hazard — *"cash alone feeds `_check_termination`, and for a held position that is most of the portfolio missing"* — it just wasn't applied at construction.

## Testing

New test constructs the env with an unsettled close and asserts the baseline is 10000, not 1000. Reverting to `account.cash` fails it; nothing else in the 2139-test suite noticed.

### DIFF
--- a/tests/envs/alpaca/test_torch_env.py
+++ b/tests/envs/alpaca/test_torch_env.py
@@ -806,3 +806,37 @@ class TestAlpacaTorchTradingEnvSeed:
         # Should not raise
 
 
+
+
+def test_bankruptcy_baseline_counts_a_position_that_has_not_settled_yet():
+    """__init__ closes positions, then pins the baseline -- but the close does not block.
+
+    close_all_positions() submits market orders and returns. Reading account.cash at that
+    instant excludes whatever is still tied up in the unsettled position, so an env
+    constructed holding $9k of BTC against $1k cash pinned its baseline at 1000 and
+    _check_termination then fired below $100 instead of $1000 -- a 10x understatement of
+    the account it is protecting (#284).
+
+    The baseline must be the same quantity termination compares against, which is
+    _get_portfolio_value: cash PLUS position value. Reading account.cash made it a
+    different measurement, which is why settlement timing could change it at all.
+    """
+    class UnsettledTrader(MockTrader):
+        def close_all_positions(self):
+            return {}  # submitted, not filled -- the position is still open
+
+    trader = UnsettledTrader(initial_cash=1000.0)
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps please review — live-env money path.

`__init__` closes positions (non-blocking) then reads `account.cash` as the bankruptcy baseline, so an unsettled position is missing from it: constructed holding $9k BTC against $1k cash, the baseline pinned at 1000 and termination fired below $100 instead of $1000.

Fix uses the env's own `_get_portfolio_value()` (cash + position value), which is what `_check_termination` compares against — so the two are the same measurement and settlement timing can't move them apart.

Worth your scrutiny: **is calling `_get_portfolio_value()` inside `__init__` safe?** It calls `trader.get_status()` and assigns `self.balance`, and it raises `PositionUnknownError` on an unknown status — so a broker outage at construction now raises where it previously returned a (wrong) number. Is failing closed there correct, or should construction tolerate it?

Issue #277, opened 2026-08-07
Live: leverage and distance_to_liquidation fail open (swallowed set_leverage failures; empty liqPrice → 1.0)


BY571:
## Two coupled fail-open defects in account_state

**1. set_leverage failures swallowed, applied leverage never re-read**: \`bybit/order_executor.py:197-202\`, \`okx/order_executor.py:218-228\`, \`bitget/order_executor.py:341-345\`, \`binance/order_executor.py:121-127\`. Consequences: (a) sizing uses \`config.leverage\` — config 20x on an account still at 5x computes a target 4x the available margin; (b) \`futures_live_base.py:82\` reports \`config.leverage\` when flat but \`:89\` reports \`position_status.leverage\` when open, so \`account_state[4]\` flips between 20.0 and 5.0. The policy conditions on that element.

**2. distance_to_liquidation defaults to 1.0 on empty liqPrice**: \`futures_live_base.py:95-102\` — bybit returns \`liqPrice=""\` for unified/cross (\`bybit/order_executor.py:342\`), OKX \`liqPx=""\` for cross (\`okx/order_executor.py:362\`). A cross-margin 20x long one 5% move from liquidation reads **1.0** — identical to a flat spot account. The formula is correct; the default is the bug (CLAUDE.md invariant #3, "an open position must read open"). \`exposure_pct\` at \`:93\` has the same fail-open shape.

## Fix

Verify applied leverage after set (read back; raise on mismatch), and compute a fallback liquidation distance from leverage + maintenance margin when the venue omits liqPrice instead of defaulting to the safest possible value.

PR #342, opened 2026-08-12
fix(live): stop live envs reading fabricated numbers as safe (#277)

BY571:
Closes part of #277.

> **Scope note.** This started as the two `account_state` defaults #277 names. Eight review rounds later it is a sweep: no live env may read a fabricated, non-finite, or venue-unconfirmed number as a safe one. The PR was retitled to match. Everything the sweep turned up that was *pre-existing* is filed (#343, #344, #345, #347, #348, #349) rather than fixed here.

**The single worst bug was not in the observation at all.** Alpaca sized and submitted a real order from unguarded `cash`, `market_value` and a hand-rolled `float(position_status.qty)`. A NaN makes `delta_qty` NaN; `nan > 0` is `False` so it always falls to the **sell** branch; and in `trade_mode="notional"` a sell is intercepted as `close_position()`. One garbage tick liquidates the position.

**Most findings were in my own fixes.** Round 2's "simplification" reintroduced the bug on NaN. Round 3's `leverage <= 1` gate became a route in. Round 4 found the equity hardening had never landed, because three adapters *fabricate* the value rather than omitting the key. Round 5's finiteness loop covered only the fields I happened to think of. Round 8 found the `+inf` guard fixed one commit earlier still wrong on four lines the same PR authored. Each is recorded in its commit message.

## The bug

Defaults in the shared futures `_get_observation` answered a missing venue field with the safest possible number, so a position the account is actually holding read like no position at all — CLAUDE.md invariant #3, *"a flat account must read flat, AND an open position must read open"*.

| Site | Missing input | Old answer | Why it is wrong |
|---|---|---|---|
| `distance_to_liquidation` | `liquidation_price <= 0` | `1.0` | Same branch as "no position". bybit sends `liqPrice=""` for unified/cross, OKX `liqPx=""` for cross — a 20x long 4% from liquidation read exactly as safe as a flat spot account. |
| `exposure_pct` denominator | `total_margin_balance` key absent | `0` | All four adapters build the key unconditionally, so absence is an adapter bug. The default made every position read 0% exposed, permanently. |
| `exposure_pct` | equity `<= 0` with a position on | `0.0` | An underwater position held against wiped equity reported as flat. |

The venue does not omit `liqPrice` arbitrarily: under cross margin the whole account backs the position, so there is no per-position price to publish. The formula was never the problem — the default was.

## Same default, four more sites

`.get("total_margin_balance", 0)` also sat in `_get_portfolio_value` and in all four per-exchange bankruptcy baselines. Through `is_bankrupt`, which is `current < threshold * initial`, that one default breaks termination in **both** directions:

- as the **baseline**: `initial = 0` makes `current < threshold * 0` never true — the account can be wiped out and the episode trades on
- as the **current value**: `0 < threshold * initial` is true for any funded account — instant false bankruptcy

All five now index the key. A structural guard keeps them indexed.

## The fix

- **Estimate rather than default.** When the venue omits the price, derive it from *both* the isolated geometry and the account's equity, and use whichever sits nearer the mark on the losing side.

  Isolated alone is not the conservative choice it looks like — @CharlieHelps caught this. It sees only this position, so it answers the same number whether the account is flush or nearly empty; once losses elsewhere have eaten the collateral, cross liquidates *earlier* than isolated says and the estimate **overstates** the distance. That is the same fail-open, one level down. Cross alone is no better: it cannot see other positions' maintenance requirements or the venue's risk tiers, and prices a well-funded long at −903. Only the nearer-of-the-two cannot overstate.

  | case (100-notional, 20x) | isolated | cross | used |
  |---|---|---|---|
  | equity = initial margin (5.0) | 95.40 | 95.38 | 95.38 |
  | collateral eaten elsewhere (2.0) | 95.40 | **98.39** | **98.39** |
  | amply funded (1,000,000) | 95.40 | −903.61 | 95.40 |
- **Gated on `leverage == 1`**, matching offline's `has_liquidation`, as its own branch rather than a fall-through — a short at 1x would otherwise compute `(0 - price)/price = -1` and clamp to `0.0`, reporting an unlevered position as *at* liquidation. `== 1` and not `<= 1`, because the loose form handed 0, negative and fractional leverage the maximally-safe answer, and that is reachable: OKX blanks `lever` on cross positions and its adapter substitutes the *config* leverage.
- **Fail closed elsewhere**, matching the unknown-status path already in that function.
- The venue's own `liqPrice` is still used verbatim when present; the fallback is for an absent price, not a second opinion on a published one.

Guards are spelled `not (x > 0)` / `not (x >= 1)` rather than `x <= 0` / `x < 1`. NaN compares False to every operator, so the natural spelling skips the raise and lands on `max(0.0, nan)`, which returns `0.0` — the exact "reads as safe" answer, arrived at silently.

## Why a shared module

The formula already existed twice offline (scalar + tensorised). Those two are pinned to each other by the scalar/vectorized equivalence tests. A third live copy would have had **no such pin** — the shape that produced BUG 1 (three SLTP action-map implementations).

So the scalar rule moves to `envs/utils/liquidation.py`, is shared with `sequential.py`, and owns the `maintenance_margin_rate` default both offline configs now import. `test_live_distance_to_liquidation_agrees_with_the_offline_env` compares `account_state[5]` end to end over leverage 1/2/20 — end to end rather than price-to-price because the gates around the arithmetic are where the two actually diverged.

## The hardening didn't land the first time

Worth stating plainly, because it is the most useful thing in this PR: indexing `total_margin_balance` instead of `.get(..., 0)` **did nothing on three of the four exchanges**. bitget, bybit and okx *fabricate* the value — `.get('total', 0)`, `.get("totalEquity", 0)`, `.get("totalEq") or "0"` — and publish a well-formed `0.0` under the key. So the indexed read never raised, and two commits' worth of comments described a fix that had not happened.

Fixed at the adapter boundary, where the venue field is actually missing. Also: a genuinely empty account still produced a bankruptcy baseline of `0`, and `current < threshold * 0` is never true, so all four envs now refuse to start an episode on non-positive equity.

Two more routes found in the same round:

- **`int()` truncation defeated the `leverage == 1` branch** added one commit earlier. bybit and okx coerced venue leverage with `int(float(...))`, so anything in `[1, 2)` became exactly `1`, took the no-liquidation branch, and reported a levered position as maximally safe — on the exact cross-margin path the fallback exists for. `PositionStatus.leverage` is a float now.
- **NaN skipped every remaining comparison.** A NaN liquidation price passes `<= 0`, misses the fallback, reaches the arithmetic and clamps to distance `0.0` — a healthy 20x position reported as *at* liquidation on one garbage tick. The venue's numbers are now checked for finiteness once.

## And the round after that

- **The finiteness check guarded the fields I was thinking about.** `notional_value` and `unrealized_pnl_pct` are read in the same branch and go straight into `account_state` unchecked — a NaN notional makes `exposure_pct = nan/equity`, putting **NaN into the observation tensor and from there into the policy network**. Worse than the NaN liquidation price fixed the round before, which at least clamped to a number. The test had been parametrized over exactly the three fields already fixed, so it certified the fix rather than the property.
- **`or` is not `.get(key, default)` for a numeric zero.** The previous round's `float(pos.get("leverage") or self.leverage)` silently swaps a venue-reported leverage of `0` for the config value — inventing a liquidation distance from a leverage the venue never confirmed. bitget reads ccxt's unified `leverage`, a *number* not a REST string, so `0` arrives falsy there today. That defect survived the entire mutation run, because the leverage cells construct `PositionStatus` directly and never touch adapter parsing; it has a structural guard now.
- **Alpaca had the identical unguarded bankruptcy baseline** — it inherits the same `_check_termination` and `is_bankrupt` — and the guard added the round before excluded it *by name*, so the one env still carrying the bug was the one the test skipped.
- **`current < threshold * 0` is not "never true"** — it reduces to `current < 0`. The conclusion held, the stated mechanism didn't, in five copies.

## And the one after that

- **Alpaca had none of the guards.** The futures four got the finiteness checks and the held-position raise; the spot env kept `exposure_pct = ... if portfolio_value > 0 else 0.0` and no validation at all. A NaN `unrealized_plpc` goes into the tensor and on into the policy network; margin debt exceeding the position reads a held position as flat. The same shape as the bankruptcy-baseline bug one round earlier, where the guard excluded alpaca *by name*.
- **Equity itself was never checked for finiteness.** The held-position guard is keyed on direction, so a NaN equity on a **flat** account passed silently — and reaches `is_bankrupt()`, where `nan < threshold * initial` is `False`, disabling termination for the rest of the episode.
- **A comment I falsified two commits earlier.** It said bybit blanks `leverage` alongside `liqPrice` so its adapter yields `POSITION_UNKNOWN` before reaching the fallback. True until I stopped `float("")` raising. bybit now reaches it, on a leverage this env *asked for* rather than one the venue confirmed. The fallback stays — refusing would leave OKX and bybit cross accounts unable to produce an observation at all — but the comment says so now and names the dependency on #277's unfixed half.

## Review found three defects, two of them mine

Recorded because the second round earned its keep:

1. **The guards failed open on NaN** — including in a "simplification" I had just accepted, which replaced a three-branch block with `if total_balance <= 0 ... else 0.0`. Not equivalent: NaN equity skipped the raise *and* took the `0.0` arm, reporting a held position as flat.
2. **Live and offline diverged at leverage 1** — the live default. Offline gates on `has_liquidation`; the ungated live fallback priced a 1x long at `mmr*entry` for 0.996 against offline's 1.0.
3. **My parity test could not see it.** It hardcoded `has_liquidation=True` and passed the same rate to both sides, so it compared the arithmetic while the divergence lived in the gates. Rewritten end to end, sourcing the offline rate from the config default.

## Reviewer findings I did not fully take

@CharlieHelps also flagged that the non-positive-equity `ValueError` can strand an open live position, since no `_step` catches observation failures. Half taken: the transient route is closed at the boundary (bitget and four sizing paths coerced a missing equity field to `0`; they now raise where it is diagnosable). The raise itself stays — `_get_observation` already raises `PositionUnknownError` on that same path, and the only non-raising alternative is `exposure_pct = 0.0`, a held position reported as flat. A close/halt fail-safe applies to every existing raise, not just this one, so it is filed as #343.

## Scope I did not take

- **`set_leverage` failures are still swallowed** (part 1 of #277). Verifying applied leverage means a read-back in four executors and touches sizing, so #277 stays open for it. That is also why `account_state[4]` can still flip between `config.leverage` when flat and `position_status.leverage` when open.
- **Behaviour change beyond the issue:** the helper raises on `entry_price <= 0` / `leverage < 1` where offline previously returned `0.0`. A zero liquidation price is indistinguishable downstream from having no position. Follows invariant #4 (raise rather than absorb nonsense); covered by a test on the offline route too.
- **The four per-exchange zero-liq tests** were retargeted, not deleted. They pin things the shared test cannot: the `entry_price` field name across four different `PositionStatus` dataclasses, and config-leverage vs position-leverage (their fixtures differ, 5 vs 10).

## Verification

Every fix is mutation-checked — each mutant is killed by the test written for it:

| Mutant | Tests failed |
|---|---|
| Nearer-of-two → isolated only (Charlie's finding) | 2 |
| Nearer-of-two → cross only | 10 |
| Nearer/further flipped | 12 |
| Gate back to `leverage <= 1` (the OKX route) | 3 |
| Equity raise keyed on notional again | 1 |
| Offline `has_liquidation` forced False | 4 |
| Reintroduce a single-quoted equity default | 1 |
| Fallback uses config leverage, not the position's | 8 |
| Revert the liquidation fallback | 2 |
| Drop the leverage-1 gate | 2 |
| Drop the live distance clamp | 2 |
| Helper guards back to `< 1` / `<= 0` (NaN) | 2 |
| `_get_portfolio_value` default | 2 |
| Exposure guard back to `<= 0` (NaN) | 1 |
| One per-exchange bankruptcy baseline | 1 |
| Drift the shared maintenance rate | 2 |
| Adapter fabricates 0 equity again | 1 |
| Drop the non-finite venue check | 4 |
| One env drops the baseline validation | 1 |
| Restore `int()` leverage truncation | 1 |
| Loop back to 3 fields (NaN into the tensor) | 3 |
| Alpaca baseline guard removed | 1 |
| Leverage back to `or` (numeric 0) | 1 |
| Alpaca finiteness loop removed | 5 |
| Alpaca flat-reading guard removed | 1 |
| Equity finiteness check removed | 2 |

Full `pytest tests/` — **2551 passed, 19 skipped**. The offline extraction is behaviour-neutral.

Sanity check on the estimate: for the test position (entry 50000, mark 50100, 10x) the fallback gives 45200/54800 against the venue-published 45000/55000 in the neighbouring cells — the right neighbourhood, not a placeholder.

### DIFF
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -383,15 +383,17 @@ class TestBinanceFuturesTorchTradingEnv:
 
     @pytest.mark.parametrize("qty,liq_price,expected_dtl", [
         (0.001, 45000.0, pytest.approx(0.1018, rel=1e-2)),   # long normal: (50100-45000)/50100
-        (0.001, 0.0, 1.0),                                   # long zero-liq -> unknown -> 1.0
+        (0.001, 0.0, pytest.approx(0.0978, rel=1e-3)),   # long, venue omits liq: 50000*(1-1/10+0.004)=45200
         (-0.001, 55000.0, pytest.approx(0.0978, rel=1e-2)),  # short normal: (55000-50100)/50100
-        (-0.001, 0.0, 1.0),                                  # short zero-liq -> unknown -> 1.0 (the fix)
+        (-0.001, 0.0, pytest.approx(0.0938, rel=1e-3)),  # short, venue omits liq: 50000*(1+1/10-0.004)=54800
     ], ids=["long-normal", "long-zero-liq", "short-normal", "short-zero-liq"])
     def test_distance_to_liquidation(self, env, mock_trader, qty, liq_price, expected_dtl):
         """distance_to_liquidation (account_state[5]) across long/short x normal/zero-liq.
 
-        A zero/absent liq price (cross-margin, or .get(..., 0)) must read 1.0 -- for a short the
-        unguarded (0 - price)/price = 0.0 would falsely signal AT liquidation. Matches bybit/okx.
+        A zero/absent liq price (cross-margin) is estimated from leverage rather than read
+        as 1.0, which reported a levered position as being as safe as a flat account (#277).
+        The estimate also has to land before the subtraction: an unguarded (0 - price)/price
+        clamps to 0.0, the opposite lie, falsely signalling a short is AT liquidation.
         """
         from torchtrade.envs.live.binance.order_executor import PositionStatus
 
--- a/tests/envs/bitget/test_torch_env_futures.py
+++ b/tests/envs/bitget/test_torch_env_futures.py
@@ -361,15 +361,17 @@ class TestBitgetFuturesTorchTradingEnv:
 
     @pytest.mark.parametrize("qty,liq_price,expected_dtl", [
         (0.001, 45000.0, pytest.approx(0.1018, rel=1e-2)),   # long normal: (50100-45000)/50100
-        (0.001, 0.0, 1.0),                                   # long zero-liq -> unknown -> 1.0
+        (0.001, 0.0, pytest.approx(0.0978, rel=1e-3)),   # long, venue omits liq: 50000*(1-1/10+0.004)=45200
         (-0.001, 55000.0, pytest.approx(0.0978, rel=1e-2)),  # short normal: (55000-50100)/50100
-        (-0.001, 0.0, 1.0),                                  # short zero-liq -> unknown -> 1.0 (the fix)
+        (-0.001, 0.0, pytest.approx(0.0938, rel=1e-3)),  # short, venue omits liq: 50000*(1+1/10-0.004)=54800
     ], ids=["long-normal", "long-zero-liq", "short-normal", "short-zero-liq"])
     def test_distance_to_liquidation(self, env, mock_trader, qty, liq_price, expected_dtl):
         """distance_to_liquidation (account_state[5]) across long/short x normal/zero-liq.
 
-        A zero/absent liq price (cross-margin, or .get(..., 0)) must read 1.0 -- for a short the
-        unguarded (0 - price)/price = 0.0 would falsely signal AT liquidation. Matches bybit/okx.
+        A zero/absent liq price (cross-margin) is estimated from leverage rather than read
+        as 1.0, which reported a levered position as being as safe as a flat account (#277).
+        The estimate also has to land before the subtraction: an unguarded (0 - price)/price
+        clamps to 0.0, the opposite lie, falsely signalling a short is AT liquidation.
         """
         from torchtrade.envs.live.bitget.order_executor import PositionStatus
 
--- a/tests/envs/bybit/test_torch_env_futures.py
+++ b/tests/envs/bybit/test_torch_env_futures.py
@@ -445,12 +445,12 @@ class TestBybitZeroLiquidationPrice:
 
     @pytest.mark.parametrize("qty,liq_price,expected_dtl", [
         (0.001, 45000.0, pytest.approx(0.1018, rel=1e-2)),  # Long normal: (50100-45000)/50100
-        (0.001, 0.0, 1.0),                                   # Long zero liq: unknown → 1.0
+        (0.001, 0.0, pytest.approx(0.0978, rel=1e-3)),   # long, venue omits liq: 50000*(1-1/10+0.004)=45200
         (-0.001, 55000.0, pytest.approx(0.0978, rel=1e-2)),  # Short normal: (55000-50100)/50100
-        (-0.001, 0.0, 1.0),                                   # Short zero liq: unknown → 1.0
+        (-0.001, 0.0, pytest.approx(0.0938, rel=1e-3)),  # short, venue omits liq: 50000*(1+1/10-0.004)=54800
     ], ids=["long-normal", "long-zero-liq", "short-normal", "short-zero-liq"])
     def test_distance_to_liquidation(self, env, mock_env_trader, qty, liq_price, expected_dtl):
-        """Zero liquidation price must return 1.0 (consistent with Binance/Bitget)."""
+        """distance_to_liquidation, with and without a liquidation price from the venue."""
         from torchtrade.envs.live.bybit.order_executor import PositionStatus
 
         mock_env_trader.get_status = MagicMock(return_value={
--- a/tests/envs/offline/test_sequential.py
+++ b/tests/envs/offline/test_sequential.py
@@ -1126,3 +1126,25 @@ class TestExecutionPriceConvention:
                 "about which bar the decision belongs to"
             )
         env.close()
+
+
+def test_liquidation_price_refuses_a_zero_entry_price():
+    """The offline route of the #277 guard: 0.0 back would read as "no position".
+
+    `_calculate_liquidation_price` returns 0.0 for the two legitimate no-liquidation
+    cases (spot, and no position), so a third 0.0 meaning "the entry price was garbage"
+    is indistinguishable from them downstream. Re-inlining the arithmetic here without
+    the guard is what this catches.
+    """
+    from types import SimpleNamespace
+    from torchtrade.envs.offline.sequential import (
+        SequentialTradingEnv, SequentialTradingEnvConfig,
+    )
+
+    env = SimpleNamespace(
+        has_liquidation=True,
+        leverage=10,
+        maintenance_margin_rate=SequentialTradingEnvConfig().maintenance_margin_rate,
+    )
+    with pytest.raises(ValueError, match="entry_price must be positive"):
+        SequentialTradingEnv._calculate_liquidation_price(env, 0.0, 1.0)
--- a/tests/envs/okx/test_torch_env_futures.py
+++ b/tests/envs/okx/test_torch_env_futures.py
@@ -344,12 +344,12 @@ class TestOKXZeroLiquidationPrice:
 
     @pytest.mark.parametrize("qty,liq_price,expected_dtl", [
         (0.001, 45000.0, pytest.approx(0.1018, rel=1e-2)),
-        (0.001, 0.0, 1.0),
+        (0.001, 0.0, pytest.approx(0.0978, rel=1e-3)),   # long, venue omits liq: 50000*(1-1/10+0.004)=45200
         (-0.001, 55000.0, pytest.approx(0.0978, rel=1e-2)),
-        (-0.001, 0.0, 1.0),
+        (-0.001, 0.0, pytest.approx(0.0938, rel=1e-3)),  # short, venue omits liq: 50000*(1+1/10-0.004)=54800
     ], ids=["long-normal", "long-zero-liq", "short-normal", "short-zero-liq"])
     def test_distance_to_liquidation(self, env, mock_env_trader, qty, liq_price, expected_dtl):
-        """Zero liquidation price must return 1.0 (consistent with other exchanges)."""
+        """distance_to_liquidation, with and without a liquidation price from the venue."""
         from torchtrade.envs.live.okx.order_executor import PositionStatus
 
         mock_env_trader.get_status = MagicMock(return_value={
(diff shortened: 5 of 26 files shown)


BY571:
@CharlieHelps please review. Focus on the liquidation-estimate fallback in `futures_live_base.py` (is understating the distance under cross margin the right direction?) and on the two new fail-closed raises — is raising on "position held against non-positive equity" safe on a live path, or does it kill a session at the worst moment?

BY571:
Thanks — both were real. Fixed in 3rd commit; second one partially, with reasoning.

## 1. Cross estimate not reliably conservative — you were right, fixed

My comment claimed the isolated estimate "understates the distance". That only holds when the rest of the account is net-positive collateral. As you say, once losses elsewhere have consumed it, cross liquidates *earlier* than isolated and the estimate **overstates** distance — the same fail-open, one level down. The claim was wrong, not just imprecise.

The threshold now uses account equity too, solving `equity + size*(P - mark) = mmr * |size| * P`, and takes **whichever of the two sits nearer the mark on the losing side**. Neither dominates: cross can't see other positions' maintenance requirements or the venue's risk tiers (it prices a well-funded long at -903), so it isn't trusted alone either.

Measured:

| case (100-notional, 20x) | isolated | cross | used |
|---|---|---|---|
| equity = initial margin (5.0) | 95.40 | 95.38 | 95.38 |
| collateral eaten elsewhere (2.0) | 95.40 | **98.39** | **98.39** |
| amply funded (1,000,000) | 95.40 | -903.61 | 95.40 |

Row 2 is your scenario: liquidation is 3 points nearer than isolated claimed. Reverting to isolated-only now fails 2 tests.

I did **not** take the "represent as explicitly unknown" option: `account_state` has no channel for it, which is why this codebase raises for unknown position status rather than encoding it — and a sentinel there would be read by the policy as a number regardless.

## 2. The ValueError stranding a live position — narrowed, not removed

You're right that it was reachable from partial adapter data, and that route is now closed at the boundary rather than absorbed: bitget's `float(usdt_balance.get('total', 0))` and four `env.py` sizing paths that did the same have been indexed, so a missing equity field raises *there*, where it is diagnosable, instead of arriving here as a plausible `0`. That guard also caught four sizing paths where the `0.0` default silently stopped the env trading altogether.

I kept the raise for a genuine zero-equity-with-a-position reading, for three reasons:

- `_get_observation` **already** raises on this path (`PositionUnknownError`, same method, ~30 lines up), so this is an established contract rather than a new hazard.
- The alternative you'd get by not raising is `exposure_pct = 0.0` — a held position reported as flat, which is the exact defect this PR exists to remove. There is no conservative *finite* exposure for a position held against zero equity; the ratio is unbounded.
- At genuinely zero equity with a position on, the venue is already liquidating. Ending loudly beats trading on a fabricated observation.

Your underlying point stands though, and it is **broader than this PR**: no live env has a fail-safe that closes or halts when `_get_observation` raises — that's true of the existing `PositionUnknownError` too, and has been. I've filed it rather than bolting a half-version onto this PR: **#343**.

## Also fixed from the same review round

- `leverage <= 1` was itself a route in — it gave 0, negative and fractional leverage the maximally-safe answer. Reachable: OKX blanks `lever` on cross and its adapter substitutes the *config* leverage, so an env at the default 1x against a 20x account read `distance = 1.0`.
- The equity raise keyed on `position_value`, which binance defaulted to `0` — so a venue omitting the notional skipped the raise on a held position.

BY571:
@CharlieHelps please re-review — both findings are addressed in the latest commit (cross-margin equity now feeds the estimate and the nearer of the two is used; the transient route into the equity raise is closed at the adapter boundary, with the fail-safe itself filed as #343). Two further reachable routes found in the same round are also fixed: the `leverage <= 1` gate and the raise keying on notional rather than position direction.

BY571:
Verified and you're right — the claim was false, and I've removed it rather than softened it.

Your arithmetic checks out exactly: `5 + (P - 100) = 0.004*P + 4` → `P = 99/0.996 = 99.3976`, so **0.60% real room against the 4.60% both estimates report**. Neither the isolated nor the equity-based estimate carries another position's maintenance requirement, and taking the nearer of two incomplete numbers does not add the missing liability back. Calling it a bound was wrong.

What changed:

- `nearest_liquidation_price`'s docstring no longer claims it "cannot overstate the distance". It now states the limitation, your worked example, and what closing it requires.
- Same at the call site in `futures_live_base.py`.
- Filed **#344** for the real fix: expose the account-level maintenance requirement from `get_account_balance()` (binance `totalMaintMargin`, bybit unified MM, OKX `mmr`, bitget via ccxt — **no adapter parses any of them today**, which is why it isn't in this PR) and solve `equity + size*(P - mark) = account_maintenance(P)` instead.

On shipping it rather than routing to an unsafe/unknown path — my reasoning, and I'd genuinely like your read on it:

1. **It is a strict improvement on what it replaces, in every case.** The old code returned `1.0` for every cross position unconditionally. That is maximally wrong always, including the single-position case the new estimate gets exactly right. Your scenario makes it too optimistic; it does not make it worse than `1.0`.
2. **The single-position assumption is already framework-wide.** `exposure_pct`, `_get_portfolio_value` and the bankruptcy baseline all divide this position against account-wide equity and always have. If another cross position exists, those are wrong in the same direction — so routing only `distance_to_liquidation` to "unknown" would single out the one element that just got more accurate.
3. **There is no unknown channel to route to.** `account_state` is a 6-float tensor; a sentinel is read by the policy as a number. The codebase's existing answer to "we cannot know" is to raise (`PositionUnknownError`), and raising here would mean no OKX cross account can run at all.

If you think (1)–(3) don't hold — particularly if you'd rather this raise until #344 lands — say so and I'll take that route instead.

BY571:
Merging without waiting on the review dismissal. The `CHANGES_REQUESTED` here is about the cross-margin estimate ignoring other positions' maintenance requirements (#344) — and @CharlieHelps' own #350, which is stacked on this branch, is the fix for it. I reviewed and approved that separately: in the scenario raised here it reports 0.6% room where this PR reports 4.6%, which is the true value. #350 follows immediately.

PR #351, opened 2026-08-12
fix(offline): make set_seed reach the RNGs that decide an episode (#273)

### LINKED ISSUE #273: env.set_seed() is a no-op for episode starts and initial cash; slippage drawn from global RNG
## Bug

\`torchtrade/envs/core/base.py:124-136\` \`_set_seed\` only touches global \`np.random.seed\`/\`torch.manual_seed\`. The two RNGs that actually drive episode diversity are constructed from \`config.seed\` and never reseeded:
- sampler start-index RNG: \`infrastructure/sampler.py:33\` (\`self.np_rng\`)
- \`InitialBalanceSampler.np_rng\`: \`infrastructure/utils.py:69\`

Measured:
\`\`\`
set_seed(1)   start idxs: [7, 55, 46, 31, 31, 61]  cash: [5892, 12739, 11545, ...]
set_seed(999) start idxs: [7, 55, 46, 31, 31, 61]  cash: [5892, 12739, 11545, ...]  ← identical
\`\`\`

\`SerialEnv\`/\`ParallelEnv\` seed workers \`seed, seed+1, ...\` — so **every parallel worker collects the same trajectories**. Effective batch diversity is 1. (\`VectorizedSequentialTradingEnv._set_seed\` at \`vectorized_sequential.py:254\` does it correctly; the scalar env is the broken one.)

## Related (same PR)

Slippage sampling uses the global torch RNG with no \`generator=\` (\`sequential.py:632\`, \`sequential_sltp.py:474,481\`, \`onestep.py:509,523,530\`) — shares the stream with policy exploration, not per-env reproducible. \`vectorized_sequential.py:521\` correctly passes \`generator=self._rng\`.

## Fix

\`_set_seed\` must reseed \`self.sampler.np_rng\` and the initial-cash sampler; slippage draws get a per-env generator. Regression test: two envs seeded differently must produce different start indices; same seed → identical.

BY571:
Closes #273.

## The bug

`_set_seed` reseeded the global numpy and torch streams. Neither of the two RNGs that determine what an episode *is* ever reads them — the sampler draws its start index from `sampler.np_rng`, and the starting balance comes from `initial_cash_sampler.np_rng`. Both are built once from `config.seed` and were never touched again.

Reproduced before changing anything:

```
seed=1    [(69, 5892), (597, 12740), (505, 11546), (339, 9389), ...]
seed=999  [(69, 5892), (597, 12740), (505, 11546), (339, 9389), ...]   ← identical
```

**The damage is not "set_seed does nothing."** `SerialEnv`/`ParallelEnv` give each worker a *different* seed precisely to decorrelate them. Every worker therefore replayed the same start indices and the same starting cash regardless: a batch of N workers with an effective diversity of **1**. Any training run using parallel collection was learning from one trajectory distribution, N times over.

Verified end to end by review: `SerialEnv(4)` and `ParallelEnv(4, spawn)` now produce distinct per-worker trajectories and reproduce byte-identically across fresh processes.

After:

```
worker seed=1: [(12, 14412), (539, 9758), (639, 6437), (135, 11006)]
worker seed=2: [(192, 13994), (722, 14264), (370, 14274), (114, 7191)]
worker seed=3: [(489, 12121), (418, 6003), (102, 6591), (292, 11325)]
worker seed=4: [(192, 14403), (697, 14775), (564, 11084), (186, 14830)]
```
15 of 16 distinct start indices; same seed still reproduces exactly.

## Why `SeedSequence.spawn` and not `seed` / `seed+1` / `seed+2`

My first fix used fixed offsets. That re-introduces the same failure one step removed: with `seed`/`seed+1`, the *cash* stream of a run seeded 1 is bit-for-bit the *start-index* stream of a run seeded 2 — and `--seed $i` over 1..N is exactly how sweeps and multi-process launches are started. `SeedSequence.spawn` is what numpy provides for this, and it keeps the three streams independent of each other as well.

**Correction.** Two earlier commit messages and an earlier version of this description said torchrl seeds workers `seed, seed+1, ...`. That is false — it derives them through `seed_generator()`, a PCG64 hash: `set_seed(1234)` → `[1234, 2931408885, 2474697360, ...]`. The fix is unchanged and still right; the *reason* is the seed-sweep pattern, not a torchrl internal. Five docstrings built on the wrong premise were rewritten.

## Slippage

Drawn from the global torch stream (`sequential.py`, `sequential_sltp.py` ×2, `onestep.py` ×3), which the policy's own exploration sampling also consumes — so what a saved seed reproduced depended on how many times the policy had sampled in between. Now a per-env generator, as `vectorized_sequential.py` already did.

One non-obvious consequence: an **unseeded** config seeds that generator nondeterministically *on purpose*. `torch.Generator()`'s default seed is a constant, so simply constructing one would have made every env in a process draw the identical slippage sequence — the opposite of what `seed=None` asks for. The global stream it replaces was nondeterministic by accident of being shared; this is nondeterministic deliberately. Three existing slippage tests depend on that and pass unchanged.

## Verification

Mutation-checked:

| Mutant | Tests failed |
|---|---|
| Revert the `_set_seed` override (the bug) | 3 |
| `SeedSequence` → `seed`/`seed+1`/`seed+2` offsets | 1 |
| Slippage back to the global stream (sequential) | 1 |
| Slippage back to the global stream (sltp, 2 sites) | 1 |
| Slippage back to the global stream (onestep, 3 sites) | 1 |
| Delete the slippage generator reseed | 1 |
| Vectorized unseeded generator → constant default | 1 |

**Two of the tests I wrote first were worthless**, and both survived a mutation run before I caught them:
- the global-stream test reseeded `torch.manual_seed(1234)` before *both* sides of its comparison, so it compared identical things regardless of what the env did
- the stream-independence test pointed both RNGs at one seed and still passed, because the drawn *values* hide a shared stream — the two draw different ranges off the same bits

Stream independence is only observable on the bit-generator state, so that is what is asserted now.

`pytest tests/` — **2501 passed, 19 skipped**.

### What review found that mutation testing alone did not

- **Deleting the slippage generator reseed passed the entire suite.** The class helper records state after `reset()` and never steps, so slippage was invisible to every cell; the global-stream test proves the draw *left* the global stream, not that `set_seed` *reaches* the private one.
- **Five of the six `generator=` call sites were uncovered** — reverting the two in `sequential_sltp.py` or the three in `onestep.py` failed nothing. The generator is shared; the keyword is six hand-written sites.
- **The vectorized env had the same unseeded bug** this PR describes. Its comment claimed otherwise. Two independent unseeded instances produced byte-identical balances. Fixed there too, per CLAUDE.md's universality rule.
- `assert episodes() == episodes()` was **vacuous** — both envs come from the same config seed, so it held with `_set_seed` replaced by a bare `return`.

## Scope

`VectorizedSequentialTradingEnv._set_seed` was already correct and is untouched; this brings the scalar envs up to it.

### DIFF
--- a/tests/envs/offline/test_seeding.py
+++ b/tests/envs/offline/test_seeding.py
@@ -399,9 +399,10 @@ class TestSetSeedReachesTheEpisodeRNGs:
 
     The two RNGs that determine what an episode IS -- the sampler's start index and the
     initial cash -- are built once from `config.seed` and were never touched again. The
-    consequence was not "set_seed is a no-op": SerialEnv/ParallelEnv seed their workers
-    `seed, seed+1, ...` precisely to decorrelate them, so every worker replayed the same
-    start indices and the same starting cash. A batch of N workers had a diversity of 1.
+    consequence was not "set_seed is a no-op": SerialEnv/ParallelEnv give each worker a
+    DIFFERENT seed precisely to decorrelate them, so every worker replayed the same start
+    indices and the same starting cash regardless. A batch of N workers had a diversity of
+    1. (Worker seeds come from seed_generator(), a PCG64 hash -- not `seed+1`.)
     """
 
     @staticmethod
@@ -425,19 +426,17 @@ class TestSetSeedReachesTheEpisodeRNGs:
             out.append((int(env.sampler._sequential_idx), float(env.balance)))
         return out
 
-    def test_different_seeds_give_different_episodes(self, large_ohlcv_df):
-        """The measurement from the issue: these two were byte-identical."""
-        assert self._episodes(large_ohlcv_df, 1) != self._episodes(large_ohlcv_df, 999)
+    @pytest.mark.parametrize("a,b", [(1, 999), (1, 2)], ids=["distant", "adjacent"])
+    def test_different_seeds_give_different_episodes(self, large_ohlcv_df, a, b):
+        """(1, 999) is the measurement from the issue -- these were byte-identical.
 
-    def test_adjacent_seeds_give_different_episodes(self, large_ohlcv_df):
-        """Adjacent, because `--seed $i` over 1..N is how sweeps and multi-process launches
-        are actually started.
-
-        (Not because torchrl hands workers `seed+1` -- it derives them through
-        seed_generator(), a PCG64 hash. Adjacency comes from the caller.) A fix that only
-        decorrelates distant seeds would leave the common case broken.
+        (1, 2) is the pair that matters in practice: `--seed $i` over 1..N is how sweeps
+        and multi-process launches are started, so a fix that only decorrelated distant
+        seeds would leave the common case broken. (Not because torchrl hands workers
+        `seed+1` -- it derives them through seed_generator(), a PCG64 hash. Adjacency
+        comes from the caller.)
         """
-        assert self._episodes(large_ohlcv_df, 1) != self._episodes(large_ohlcv_df, 2)
+        assert self._episodes(large_ohlcv_df, a) != self._episodes(large_ohlcv_df, b)
 
     def test_same_seed_still_reproduces(self, large_ohlcv_df):
         """The point of the fix is reproducibility, not merely difference."""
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps please review. Two things worth your attention specifically:

1. **`SeedSequence.spawn` vs fixed offsets.** My first attempt used `seed`/`seed+1`/`seed+2`, which re-couples across workers (worker N gets `seed + N`, so worker 1's cash stream is worker 2's start-index stream). I think spawn is right, but I'd like a second opinion on whether spawning inside `_set_seed` interacts badly with anything TorchRL does when it seeds workers.

2. **The unseeded case.** `torch.Generator()`'s default seed is a constant, so I call `.seed()` to make `config.seed=None` nondeterministic on purpose. That is a deliberate behaviour choice, not an oversight — is it the one you'd expect?

Issue #289, opened 2026-08-07
Cross-env parity gaps: binance observation base class + hedge surface, alpaca close_position_on_*, vectorized reward_function/include_base_features


BY571:
Gaps found by systematic diff of the five live integrations + scalar/vectorized offline envs (CLAUDE.md mandates feature parity):

- **binance**: \`observation.py\` doesn't inherit \`BaseFuturesObservationClass\` (236 LOC parallel implementation; bitget/bybit/okx all inherit). No \`PositionMode\`/hedge-mode config surface; \`trade()\` accepts a \`position_side\` param nothing ever sets — dead parameter with a stale comment.
- **alpaca**: no \`close_position_on_init\`/\`close_position_on_reset\` (all four futures exchanges have them) — no escape hatch for stale positions after restart.
- **vectorized offline envs**: no \`include_base_features\`; reward formula hardcoded inline (numerically matches \`log_return_reward\`, but no pluggable \`reward_function\` like both scalar envs); \`log_return_reward\` raises on \`old_value <= 0\` while vectorized clamps to 1e-10 — align the edge-case behavior.
- **naming**: binance \`margin_type: MarginType\` vs \`margin_mode: MarginMode\` elsewhere (each exchange redefines the enum locally) — breaks exchange-agnostic config code.
- **test depth**: vectorized SLTP has 10 tests vs scalar SLTP's 31, despite the "passes extensive equivalence tests" docstring (see also #274).

Best sequenced AFTER the dedup refactor (#288-style consolidation) so parity features land once in the shared base rather than 4-5 times.

BY571:
**Status after #403** — four of six bullets are closed, two remain, and one was stale. Verified by execution rather than by reading:

| bullet | state |
|---|---|
| binance `observation.py` doesn't inherit `BaseFuturesObservationClass` (236 LOC parallel impl) | **closed** — inherits, now 80 lines |
| binance `trade()` accepts a dead `position_side` param | **closed by #403** — removed from binance, bitget and the replay executor, along with four unreachable `if position_side != "BOTH"` branches |
| alpaca lacks `close_position_on_init`/`_on_reset` | **closed** — both present and honoured at `alpaca/base.py:107,341` |
| vectorized lacks `include_base_features` / pluggable `reward_function` | **closed** — `reward_function` wired at `vectorized_sequential.py:156` and `vectorized_sequential_sltp.py:238`; `include_base_features` now *raises* rather than silently doing nothing |
[...]

Issue #278, opened 2026-08-07
Replay fidelity: reward lags one bar; fractional+fee rejects 100% of trades; reset never rewinds; obs keys don't match live/offline


BY571:
## Four defects that break the offline↔replay comparison the replay pipeline exists for

**1. PV/reward one-bar lag**: live \`_step()\` order is trade → wait → read PV → get obs. In replay the clock only advances inside \`ReplayObserver.get_observations()\` (\`observer.py:59-65\`) — *after* the PV read. Measured: \`recorded PV[t] == true equity[t-1]\` exactly. Reward at step t belongs to the action at t-1, and an SL/TP close during the bar is invisible to that step's reward.

**2. Fractional mode + fee = 100% rejection**: \`replay/order_executor.py:163-170\` requires \`balance >= notional/leverage + fee\` with no headroom, but the live fractional sizer (\`bybit/env_sltp.py:266-271\`) reserves no fee — unlike offline \`utils/fractional_sizing.py:76-80\` (\`fee_multiplier\`). balance=10000, lev=5, fee=0.0004, fraction=1.0 → need 10020, have 10000 → every action is a silent no-op; replay reports a flat 'strategy'.

**3. reset() never rewinds**: \`ReplayObserver.reset()\` (\`observer.py:105-110\`) is called only from a test. Under a collector, episode 2 continues mid-stream with episode 1's balance, open position (brackets cancelled → orphaned), and pinned \`initial_portfolio_value\`.

**4. Spec/key mismatches**: \`observer.py:81-94\` — returns keys like \`market_data_1Minute\` where real observers produce \`market_data_1Minute_10\`; with no \`feature_preprocessing_fn\` the declared spec is \`(3, 0)\` vs emitted \`(3, 5)\`; only \`base_arr[-1]\` is filled (rest zeros). A policy trained offline cannot be fed the replay env without renaming in_keys.

Also: \`reduce_only=True\` discards \`quantity\` (full close), and the lot-size rounding helper (\`:267-275\`) is defined but never called.

BY571:
Investigated defect **2** (fractional + fee = 100% rejection). The diagnosis in the issue is right, and the fix is a little larger than it looks — recording it before it gets attempted casually.

**The affordability check is not the bug.** `replay/order_executor.py:232` requires `balance >= margin_required + fee`, which is arithmetically correct. The bug is the **caller**: all four SLTP envs size with

```python
quantity = balance * position_fraction * leverage / current_price
```

[...]

PR #352, opened 2026-08-12
fix(binance): round order quantities to the venue's LOT_SIZE step (#271)

### LINKED ISSUE #271: Binance order executor never applies exchange lot-size rounding (hardcoded round(qty, 3))
## Bug

\`torchtrade/envs/live/binance/order_executor.py:142-160\` (\`_fetch_price_precision\`) parses only \`PRICE_FILTER\` (tick size) from \`futures_exchange_info()\`, never \`LOT_SIZE\` (\`stepSize\`/\`minQty\`). Quantities are hardcoded \`round(quantity, 3)\` at lines 215, 260, 277, 437, 468.

Bitget, Bybit and OKX all fetch and cache the real per-symbol qty step (\`bybit/order_executor.py:123-166,443-477\`, \`bitget/order_executor.py:181-203\`, \`okx/order_executor.py:132-165,467-501\`). This is the same bug class as the already-fixed bitget lot-size bug — the fix was never applied to binance.

Smoking gun: \`tests/envs/binance/test_futures_order_executor.py:71\` mocks a \`LOT_SIZE\` filter that production code never reads — a dead fixture for wiring that never happened.

Any symbol whose step precision isn't exactly 3 decimals gets silently wrong order quantities (rejected orders or mis-sized positions).

## Fix

Parse \`LOT_SIZE\` alongside \`PRICE_FILTER\`, round quantities to the fetched step (mirroring the bybit implementation), and make the existing mock fixture actually exercised by a test that fails on \`round(q, 3)\`.

BY571:
Closes #271.

## The bug

`_fetch_price_precision` parsed `PRICE_FILTER` and stopped, so every order quantity went out as a hardcoded `round(quantity, 3)` — five call sites. Any symbol whose step is not exactly three decimals got a silently wrong size: a rejected order, or a mis-sized position.

Bitget, bybit and okx all fetch the real per-symbol step. Binance was the one exception, which is why #271 calls it "the same bug class as the already-fixed bitget lot-size bug — the fix was never applied to binance."

The mock in `tests/envs/binance/test_futures_order_executor.py:71` has carried a `LOT_SIZE` filter since before any production code read it. It does now.

## Two things the obvious implementation gets wrong

**A bare floor shaves exact multiples.** `0.29 / 0.01` is `28.999999999999996` in binary, so `math.floor` returns `0.28` — a whole step lost on a perfectly valid size. Measured: **207 of the first 400 exact multiples** of the four common steps (`0.0001`, `0.001`, `0.01`, `0.1`, `1`) land just under an integer. The epsilon is load-bearing, not defensive decoration.

**Steps are cached per symbol, not one per executor.** `close_all_positions` closes whatever the account holds, not just `self.symbol` — rounding ETHUSDT's quantity with BTCUSDT's step would be this same bug wearing a different hat. `futures_exchange_info()` returns every symbol anyway, so keying the cache costs nothing.

Floors rather than rounds to nearest, matching bybit: rounding **up** can ask for more than the available margin covers, and the venue rejects the whole order.

Renamed `_fetch_price_precision` → `_fetch_symbol_filters`, since "price precision" stopped describing what it does.

## What review found after the first commit

**The issue was not actually fixed by the executor change.** `binance/env.py` carried a *second* LOT_SIZE parser — re-querying `futures_exchange_info()` on every sizing decision and flooring with a bare `int(qty / step) * step`. No epsilon, so `0.29 → 0.28`: exactly what this PR's own test forbids, still live in the fractional path, and unreachable from the executor because the env discarded the step before calling `trade()`. Both sizing paths now delegate to the executor and the duplicate parser is deleted — one owner of lot-size knowledge, not two.

That made `round_quantity` part of the trader interface, which surfaced a genuine integration gap: `ReplayOrderExecutor` acts as a trader and didn't implement it, so binance-in-replay would have failed at runtime. It returns the quantity unchanged — replay has no venue to query, and inventing a step would make a backtest disagree with the exchange it stands in for in a second, invisible way (#278).

**Two regressions from my own first fix pass:**

- Scoping the `try` to the *symbol* meant one bad filter aborted that symbol's remaining filters — including its `PRICE_FILTER`. On the traded symbol that left `_tick_size` `None`, so SLTP stop prices go out at full float precision, draw `-1111`, and are swallowed by the non-fatal bracket handler: **a position open with no stop loss**. Construction didn't even raise, because the step was cached on the line before the failure. Scoped per *filter* now.
- Narrowing the `try` moved `info['symbols']` outside it, so a truncated or error-shaped body took down `__init__` with a bare `KeyError` — neither the fallback nor the deliberate `ValueError`, whose message advertises "the payload changed shape" for a case that couldn't reach it.

Also: the foreign-symbol fallback in `close_all_positions` was silent, and `close_position` logged the *requested* size rather than the submitted one — in a PR whose entire subject is that the two differ.

## Verification

Mutation-checked:

| Mutant | Tests failed |
|---|---|
| Stop parsing `LOT_SIZE` | 5 |
| Revert the five call sites to `round(quantity, 3)` | 2 |
| Drop the epsilon (shaves exact multiples) | 3 |
| Round-to-nearest instead of floor | 1 |
| Ignore the per-symbol argument | 1 |
| Unknown symbol falls back instead of raising | 1 |
| Zero-size guard removed | 2 |
| Float instead of venue-precision string | 5 |
| Scientific notation back to string surgery | 2 |
| minQty never cached | 1 |
| The env's own bare-int floor | 1 |
| `['symbols']` read outside the try | 1 |

**The first version of these tests was the bug itself, again.** They exercised `_round_quantity` directly and nothing else — reverting all five call sites to `round(quantity, 3)` left every one of them green. That is exactly the shape of the unread `LOT_SIZE` mock this PR is fixing: a fixture for wiring that was never asserted. There is now a cell checking what an order actually carries, including the TP/SL legs, which are separately-rounded call sites.

`pytest tests/` — **2528 passed, 19 skipped**.

Worth noting: `pytest tests/envs/binance` was green while the **full suite had 6 failures** — mocks in `tests/envs/test_live_trade_state.py` didn't model the widened trader interface. A scoped run would have missed it.

## Scope and one unverified claim

`get_lot_size()` is still not added for parity with bybit/bitget/okx — `round_quantity()` is the capability the env actually needs, and an unused accessor is dead code. Cross-exchange parity gaps are tracked in #289.

**Unverified:** that binance rejects `"7.0"` with `-1111` on a `quantityPrecision: 0` symbol. I could not reach the live API. The string formatting is defensible regardless — `str(7.0)` carries a decimal such a symbol does not define, and okx formats the same way with an explicit comment about it — but the rejection itself is reasoned, not observed. One testnet order on a step-1 symbol would settle it.

### DIFF
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -68,7 +68,7 @@ class TestBinanceFuturesOrderClass:
                 "symbol": "BTCUSDT",
                 "filters": [
                     {"filterType": "PRICE_FILTER", "tickSize": "0.10"},
-                    {"filterType": "LOT_SIZE", "stepSize": "0.001"},
+                    {"filterType": "LOT_SIZE", "stepSize": "0.001", "minQty": "0.001"},
                 ],
             }]
         })
@@ -447,12 +447,12 @@ class TestBinanceLotSizeRounding:
 
         filters = [{"filterType": "PRICE_FILTER", "tickSize": "0.10"}]
         if with_lot_size:
-            filters.append({"filterType": "LOT_SIZE", "stepSize": step})
+            filters.append({"filterType": "LOT_SIZE", "stepSize": step, "minQty": step})
         symbols = [{"symbol": symbol, "filters": filters}]
         for other_symbol, other_step in extra_symbols:
             symbols.append({
                 "symbol": other_symbol,
-                "filters": [{"filterType": "LOT_SIZE", "stepSize": other_step}],
+                "filters": [{"filterType": "LOT_SIZE", "stepSize": other_step, "minQty": other_step}],
             })
 
         client = MagicMock()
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps please review. The judgement call I'd most like checked: quantities are **floored** to the step rather than rounded to nearest (matching bybit), on the reasoning that rounding up can exceed available margin and get the whole order rejected. The cost is that a close can leave a sub-step residual — though a quantity read back from the venue is already a valid multiple, so in practice the floor is exact there. Is floor the right default for binance specifically?

BY571:
@CharlieHelps please review. The judgement call I'd most like checked: quantities are floored to the step rather than rounded to nearest (matching bybit), on the reasoning that rounding up can exceed available margin and get the whole order rejected. The cost is that a close can leave a sub-step residual — though a quantity read back from the venue is already a valid multiple, so in practice the floor is exact there. Is floor the right default for binance specifically?

Issue #349, opened 2026-08-12
Remaining dust-rule and fail-open reads in live close() and price fallbacks


BY571:
Leftovers catalogued while fixing #277 in #342, all pre-existing and none worth expanding that PR for.

1. **`close()` compares `qty != 0`** in `bybit/base.py`, `okx/base.py`, `bitget/base.py`, so a dust residual triggers a spurious "closing with an open position" warning. The dust rule (`position_direction_from_status` / `position_qty_from_status`, `core/state.py`) is the one place that comparison belongs — this is the same hand-rolled `qty != 0` that #283 removed from the trading paths, surviving in the shutdown path.

2. **`alpaca/base.py::_get_current_price` still fails open.** Its `current_price <= 0` test is NaN-transparent, so a NaN position price skips the entire fallback chain and is returned to the caller. #342 made every caller guard the value, so it is caught now — but the function itself still hands back the thing it exists to avoid.

3. **`polymarket/env.py:464` `_fetch_clob_midpoint`** returns `float(mid)` unchecked. `+inf` satisfies `yes_mid >= 0.99`, so a garbage midpoint can declare an outcome resolved and produce a real payoff into `self.cash`. #342 guarded the scanner's parsed prices; this is the other price source.

4. **SLTP `_step` reads `position_status.qty` raw** into `history.record_step` — four futures `env_sltp.py`, plus `alpaca/env_sltp.py:146` and `alpaca/env.py:124`. Latent only: the shipped `log_return_reward` reads just `portfolio_values`, which #342 guarded. A reward function that used quantities would inherit the hole.

## Related

- #342 (the fail-open sweep these fell outside), #347 (the boundary validation that would close 3 and 4)

Issue #348, opened 2026-08-12
A non-positive portfolio liquidates the position instead of refusing to trade


BY571:
Found while fixing #277 in #342. Pre-existing; #342 rewrote the guard line but not the behaviour behind it.

## The bug

`alpaca/env.py:227-229` — when the portfolio value is non-positive, `_calculate_fractional_position` logs a warning and returns `(0.0, 0.0)`. The caller reads that as *target position = nothing* and executes it.

Demonstrated: cash `-5000` (margin debit), 100 units at $30, `action_value = 1.0` — the agent asking for **maximum long** — submits `{'side': 'sell', 'amount': 100.0}`. A full liquidation, from an action that meant the opposite.

The four futures envs have the same shape.

## Why it is not a #342 fix

#342 changed that line from `total_portfolio <= 0` to a finiteness check, which strictly improved it: a NaN used to take this same route and now raises. What it did not do is change what happens on a genuine non-positive portfolio, because "refuse to size" and "size to zero" are indistinguishable in the `(value, qty)` return contract — fixing it means changing that contract or raising, which is a behavioural decision beyond a fail-open fix.

## Suggested fix

Return a sentinel the caller can distinguish from a legitimate flat target, or raise. "I cannot size this" must not be executable as "go flat".

## Also here

`alpaca/env.py:209` and `:251` call `get_status()` twice in the same sizing decision, so `current_qty` and `current_position_value` can come from different snapshots. #342 flagged the same second-independent-fetch shape for `_get_portfolio_value` and fixed it there by guarding both reads; consolidating to one fetch is the better fix and applies here.

## Related

- #342, #347

BY571:
Fixed in #356 (merged as 6c5b8eb).

Alpaca now raises rather than returning `(0.0, 0.0)` when it cannot size. The old behaviour was indistinguishable from "go flat" at the call site, and since alpaca has no `target_qty == 0` guard, a zero target against an open position became `delta = 0 - current` and SOLD — from an action that meant maximum long.

Covered by `test_alpaca_refuses_to_size_rather_than_returning_a_flat_target` (empty/negative equity).

Issue #345, opened 2026-08-12
Hoist the live bankruptcy baseline into the shared futures base


BY571:
Noted while fixing #277 in #342.

All four futures envs assign the bankruptcy baseline with the same lines inside their own `_reset` scaffolding:

```python
balance = self.trader.get_account_balance()
self.initial_portfolio_value = balance["total_margin_balance"]
if not (self.initial_portfolio_value > 0):
    raise ValueError(...)
```

Four copies of one rule, in the exact shape this codebase has already paid for twice (the three SLTP action-map implementations, the three dust-rule accessors). #342 had to fix the same line in four places twice in one PR — once to index the key, once to reject a non-positive baseline — and cover it with a structural guard rather than a behavioural test, because the assignment is only reachable through a live client.

`TorchTradeFuturesLiveEnv` already owns `_get_observation` and `_get_portfolio_value`. The baseline belongs beside them, as something like `_capture_bankruptcy_baseline()` called from each `_reset`, which would also make it directly testable and let the structural guard in `tests/envs/test_live_env_base.py` be replaced with a real one.

Not done in #342 because it means touching all four `_reset` methods in a PR that is already about something else.

BY571:
Fixed in #356 (merged as 6c5b8eb).

`_capture_bankruptcy_baseline` now lives once on `TorchTradeLiveEnv` and all five live envs delegate to it, replacing five per-exchange copies. It refuses to start an episode on non-finite or non-positive equity — the `isfinite` half matters because `is_bankrupt` is `current < threshold * initial`, so a NaN baseline makes that comparison False forever and disables the stop for the entire session.

Guard verified against every value class (NaN, ±inf, 0, negative, positive) and mutation-tested. `tests/envs/test_live_env_base.py::test_every_live_env_delegates_its_baseline` fails per-exchange if any base re-forks the read.

Issue #341, opened 2026-08-11
Bitget position parsing: a nullable field reports a healthy position as POSITION_UNKNOWN, and a missing side signs it short


BY571:
Found while tracing #283/#340. Both are pre-existing and in the code that PR now depends on.

## 1. A nullable CCXT field turns a healthy position into an outage

`torchtrade/envs/live/bitget/order_executor.py:566-573,580` calls bare `float()` on nullable CCXT fields. bybit and okx defend the same reads with `or "0"`.

A cross-margin position can return `liquidationPrice: None`. `float(None)` raises `TypeError`, which the broad `except Exception` at :587 swallows, and the position is reported as `POSITION_UNKNOWN`.

That is the worst possible failure shape here: `POSITION_UNKNOWN` is deliberately fail-closed, so a perfectly healthy position stops the env every bar — and #270s whole point is that an env must not trade through an unknown status. The account would be frozen, not mis-traded, but frozen on a lie.

## 2. A missing side signs a long as a short

`bitget/order_executor.py:561` (and `bybit/order_executor.py:335`) default the side such that **anything that is not `"long"` / `"Buy"` signs as a short**. A `None` side therefore silently flips a long position to negative qty.

Everything downstream reads that sign: `position_direction_from_status`, the account_state the policy sees, and now `position_qty_from_status` on the trade path. A flipped sign is a wrong-direction trade, not a warning.

## Proposed

Mirror bybit/okxs `or "0"` guard on the nullable reads, and make the side mapping explicit — long / short / neither, with "neither" raising or reporting `POSITION_UNKNOWN` rather than defaulting to a direction. `CLAUDE.md` invariant 1 is about exactly this: never hand-roll a direction.

BY571:
Fixed in #356 (merged as 6c5b8eb), including the half the first pass missed.

`get_status` in bitget/bybit/okx now returns `POSITION_UNKNOWN` for an unusable side rather than signing a direction. OKX was the worst case: it reports hedge-mode size as a positive `pos` with direction only in `posSide`, so an unrecognised value fell through to the net-mode branch keeping the positive sign — a short read as a long, silently, straight into the trade path.

[...]

PR #356, opened 2026-08-12
fix(live): validate at the boundary, and stop guessing (#347 #345 #348 #341)

### LINKED ISSUE #347: Live futures sizing and SLTP brackets read venue prices with no validation
Found while fixing #277 in #342, and deliberately left out of that PR — the fix belongs at the adapter boundary, not bolted onto an already-large change.

## The pattern

#342 fixed this on alpaca's sizing path. The four futures envs have the identical shape and got nothing.

**Non-SLTP sizing divides by an unguarded mark price** — `binance/env.py:432`, `bitget/env.py:326`, `bybit/env.py:254`, `okx/env.py:117`:

`get_mark_price()` returns a bare `float()` of a venue string. A NaN gives `notional / nan` → `delta_qty` NaN → `nan > 0` is `False` → **`side = "sell"`**. That is the same inversion #342 documents for alpaca: a garbage tick becomes a sell, and in some trade modes a sell is a full close.

The choke point is `torchtrade/envs/utils/fractional_sizing.py:90`, which validates nothing it is handed.

**SLTP bracket prices come from an unchecked observer close** — all five `env_sltp.py`. Guarded by a NaN-transparent `if current_price <= 0` in the four futures ones and by nothing in `alpaca/env_sltp.py:230`. Binance is the worst case: `trade()` submits the main order first and treats TP/SL leg failures as non-fatal (`return True` regardless), so `trade_mode="quantity"` plus a NaN close can **open a real leveraged position with a valid size and no stop**, then record `active_stop_loss = 0.0`.

**And `PositionStatus` fields + `get_mark_price()` parse with a bare `float()`** in all four futures order executors.

## The fix

Validate at the boundary (CLAUDE.md invariant #4). Guarding `get_mark_price()` and the `PositionStatus` construction in the four adapters closes the sizing path, the bracket path, and the history-recording path at a single point, rather than adding a check at each of the ~12 consumers.

`not (x > 0)` is not sufficient — NaN passes it, and so does `+inf`. Use `math.isfinite`, as #342 settled on after getting this wrong twice.

## Related

- #342 — the same fix for alpaca sizing, the shared observation path, and the bankruptcy baselines
- #343 — no live env has a fail-safe when the observation path raises
- #344 — cross-margin liquidation estimate ignores other positions' maintenance requirements

BY571:
Closes #347, #345, #348, #341.

Four issues in one PR because they are the same defect in the same files — and combining is *cheaper*: **+212 / −116**, because hoisting the bankruptcy baseline deletes four copies of what it replaces.

## What each was

**#347 — the sizing paths divided by an unvalidated mark price.** 13 call sites across 8 files read `self.trader.get_mark_price()` raw. A NaN makes `delta_qty` NaN; `nan > 0` is `False`, so the trade falls to the **sell** branch. The `<= 0` guards beside them cannot see NaN.

**#345 — four byte-identical copies of the bankruptcy baseline.** `is_bankrupt` is `current < threshold * initial`, so a baseline of `0` reduces to `current < 0`: it never fires above zero equity and a wiped account trades on. The guard now also rejects `+inf`, which `not (x > 0)` passed. Hoisting turned the structural grep I wrote in #342 into a behavioural test.

**#348 — "I cannot size this" returned the same value as "go flat".** I checked the impact rather than repeating the issue, and it differs by env:

| | guard at the caller | consequence |
|---|---|---|
| alpaca | **none** | a zero target against an open position sells — **a liquidation from an action meaning maximum long** |
| binance/bitget/bybit/okx | `if target_qty == 0.0: not executed` | silent no-op — the agent's action dropped every bar with only a warning |

Both are wrong for the same reason, but only alpaca liquidates.

**#341 — bitget turned a healthy position into an outage, and signed longs short.** Bare `float()` on nullable CCXT fields: a cross position returns `liquidationPrice: None`, `float(None)` raises `TypeError` into the broad `except`, and the position reports `POSITION_UNKNOWN` — which is deliberately fail-closed, so the env **freezes every bar on a lie**. Separately, `contracts if side == 'long' else -contracts` signs a long as a **short** for any unexpected value, and every consumer reads that sign: `position_direction_from_status`, the `account_state` the policy sees, and the trade path. bybit had the same shape. Neither infers a direction now.

Also folded in from #349: three `close()` paths hand-rolled `qty != 0`, bypassing the dust rule, so a 1e-12 residual triggered a spurious "closing with an open position" warning.

## Verification

| Mutant | Tests failed |
|---|---|
| Drop the finiteness check | 5 |
| One env reads the raw mark price | 1 |
| One env re-forks the baseline | 1 |
| alpaca returns a zero target again | 1 |
| bitget's nullable read back to two-arg `get` | 1 |
| bybit's side defaulted again | 1 |

`pytest tests/` — **2654 passed, 19 skipped**.

## Note on scope

This is deliberately one PR for four issues. The backlog went 19-closed/17-opened over this session — flat — because I had been filing adjacent findings instead of fixing them in the PR already touching those files. Clustering by file rather than by issue is the correction.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -795,11 +795,21 @@ size, exposure and liquidation distance the env has just admitted it cannot veri
 So the env **halts**: it raises `LiveObservationHalt` and produces no transition.
 
 **Which reads this covers.** Only the post-bar state read — the portfolio value and
-observation taken after the bar closes. Two other venue reads per step are *not* routed
-through it and still raise their original exception with no latch and no emergency
-flatten: the position read at the top of `_step()` (used for the position sync and the
-duplicate-action guard) and the initial read inside `_reset()`. Catch
-`PositionUnknownError` as well if you need to cover those. Closing that gap is #355.
+observation taken after the bar closes. Every *other* venue read raises its original
+exception with no latch and no emergency flatten:
+
+| read | raises |
+|---|---|
+| position read at the top of `_step()` (position sync, duplicate-action guard) | `PositionUnknownError` |
+| the initial read inside `_reset()` | `ValueError` |
+| `_current_mark_price()`, called at the top of every futures `_step()` | `ValueError` |
+| the sizing balance / portfolio checks | `ValueError` |
+| the candle close that prices SL/TP brackets | `ValueError` |
+
+So `except LiveObservationHalt` alone does **not** give you a policy-driven emergency
+close on an unreadable price or balance — it catches the post-bar read and nothing else.
+Catch `ValueError` and `PositionUnknownError` too if you need to cover those. Alpaca and
+Polymarket do not route through this path at all. Closing the gap is #355.
 
 ```python
 from torchtrade.envs.core.live import LiveObservationHalt
--- a/tests/envs/alpaca/test_torch_env_sltp.py
+++ b/tests/envs/alpaca/test_torch_env_sltp.py
@@ -8,6 +8,7 @@ pricing is not covered here -- that belongs to the order executor and the offlin
 from unittest.mock import patch
 
 import pytest
+import numpy as np
 import torch
 from torchrl.envs.utils import check_env_specs
 from tensordict import TensorDict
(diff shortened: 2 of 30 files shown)


BY571:
@CharlieHelps please review. Four issues in one PR (#347 #345 #348 #341) because they are the same defect class in the same files.

Two judgement calls I'd like checked:

1. **A venue-reported mark price of 0 now raises rather than aborting the trade gracefully.** Two existing tests pinned `success: False` for that case; I changed them, on the reasoning that a zero price is unusable data and `success: False` is indistinguishable from a legitimate refusal. Is raising right, or should a zero price stay a soft abort?
2. **bitget/bybit/okx now set `POSITION_UNKNOWN` explicitly when the venue reports an unusable `side`/`posSide`**, rather than guessing a direction. That fails closed — the env stops rather than trading — which is correct but does mean one malformed field halts the episode.

Round 1 found that okx had the same side bug unfixed and worse (it kept a *positive* sign for an unrecognised `posSide`, so a short read as a long straight into the trade path), and that a negative mark price passed my own finiteness guard while `fractional_sizing` divides by it. Both fixed.

BY571:
@CharlieHelps please review.

This closes #347, #345, #348 and #341 in one sweep across all five live exchanges (alpaca, binance, bitget, bybit, okx). Three rounds of review agents have run; the round-3 findings were almost entirely defects in my own earlier fix commits, so the highest-value thing you can do is be adversarial about the guards themselves rather than the issues they close.

Specific things worth your scepticism:

1. `_capture_bankruptcy_baseline` in `torchtrade/envs/core/live.py` — I hoisted this from five per-exchange copies and dropped the `isfinite` half in the process, which passes NaN and disables the bankruptcy stop for the whole session. Fixed, but please check the hoist did not weaken anything else the five copies were carrying.

2. `futures_live_base._get_observation` — the flat branch now writes `current_price = 0.0` as a sentinel. That made a previously near-dead `current_price == 0` clause load-bearing, so a venue mark of exactly 0 on a held levered position read as maximally safe. The flat check now keys on `position_direction` (the dust rule). Is there another reader of that sentinel I have missed?

3. `_current_mark_price(position_status)` now serves both the fetched mark and an open position mark, replacing 8 call sites. Please confirm the flat/open branches are still distinguishable where callers need them to be.

4. `close_position` in bybit/okx now refuses an unrecognised side rather than defaulting to Buy/net. This makes `FLATTEN` return `flatten_accepted=False` in cases that previously sent a (wrong-direction) order. I believe an explicit refusal beats a venue rejection, but that is a judgement call worth a second opinion.

Known and deliberately deferred to #355: `FLATTEN` does not fire for the most common outage shape, because every adapter re-wraps its `ValueError` as `RuntimeError` and `_acquire_post_bar_state` only catches `(PositionUnknownError, ValueError)` — while it *does* fire for metadata-only failures where the position itself was reported correctly. That needs a policy decision, not a guard.

Full suite: 2755 passed, 19 skipped.

Issue #339, opened 2026-08-11
POSITION_TOLERANCE_ABS is denominated in base-asset units


BY571:
Split out of #283, whose dust-rule half is fixed separately. This half is a semantic change to sizing rather than a dust-rule violation, so it is worth its own pass.

`torchtrade/envs/utils/fractional_sizing.py:22`:

```python
POSITION_TOLERANCE_ABS = 0.001
```

That is **0.001 of the base asset**. At BTC $100k it is $100 of position left unclosed on a $10k account — 1% of the account treated as "close enough to flat". At an asset priced in cents the same constant is meaninglessly small.

A tolerance for "is this position effectively closed" should be denominated in **notional**, or as a fraction of the account, so it means the same thing across assets and price regimes.

Needs care: it is a sizing threshold on the live trade paths, so changing its denomination changes which orders are placed. Worth measuring what the current constant corresponds to across the assets in use before picking the replacement.

PR #357, opened 2026-08-12
fix(live): direction from the resulting position, and the last raw qty reads (#276, #349)

### LINKED ISSUE #276: Live envs: partial position reduction recorded as direction flip — wipes hold counter, defeats duplicate-action guard
## Bug

Direction after a trade is inferred from the **order side**, not the resulting position: \`binance/env.py:173-180\`, \`bitget/env.py:173-180\`, \`bybit/env.py:138-145\`, \`okx/env.py:141-148\`, \`alpaca/env.py:125\`. With fractional sizing, a SELL that merely trims a long is recorded as \`current_position = -1\`.

Traced (bitget, \`action_levels=[-1,-0.5,0,0.5,1]\`, long 1.0 for 20 bars, agent picks 0.5):
1. \`delta_qty < 0\` → \`side="sell"\`, \`closed_position=False\` → cache says short, exchange holds a half-size long
2. next step's sync detects the (self-inflicted) mismatch → \`hold_counter = 0\`, \`current_action_level = NaN\`
3. \`account_state[3]\` reports holding_time=1 for a 22-bar-old position; the duplicate-action guard is permanently defeated (every repeat of 0.5 re-trades)

## Fix

Set direction from the resulting position (target level sign / post-trade qty via \`position_direction_from_status\`), not the order side. Regression test: partial reduction must keep direction and hold counter.

### LINKED ISSUE #349: Remaining dust-rule and fail-open reads in live close() and price fallbacks

BY571:
Closes #276. Closes #349.

## #276 — direction inferred from the order side

Five live envs set `current_position` from `trade_info["side"]`. Under fractional sizing a SELL that trims a long from 1.0 to 0.5 **leaves a long**, but was recorded as `-1`.

The damage is the chain, not the field: the next bar's `_sync_position_from_exchange` finds a mismatch the env inflicted on itself, discards `hold_counter` and NaNs `current_action_level`. So `account_state[3]` reports `holding_time=1` for a 20-bar-old position, and the duplicate-action guard is permanently defeated — every repeat of `0.5` re-trades.

The five copies had drifted into four spellings. Binance's was worse than the rest: uppercase `"BUY"`/`"SELL"` with its `closed_position` branch **last**, behind an `elif` that always matched first — so binance could never record a close at all.

Now one `_record_position_after_trade` on `TorchTradeLiveEnv`, taking the sign of the target level, plus a structural guard so no exchange re-forks it.

**Nothing failed when I changed this across five envs.** The behaviour had no test. It has one now, end to end.

## #349 — the leftovers

| item | what it did |
|---|---|
| `alpaca/base._get_current_price` | advanced its fallback chain on `<= 0`, which NaN passes — so a NaN skipped **both** fallbacks and was returned by the function that exists to avoid exactly that. `inf` failed the other way: satisfied `> 0`, skipped the fallbacks, left nothing. Now advances on *usable*, and an exhausted chain returns its documented `0.0`. |
| `polymarket._fetch_clob_midpoint` | returned `float(mid)` unchecked while the caller decides an outcome is RESOLVED from it and pays into `self.cash` — and `inf >= 0.99` is True. Outside `[0, 1]` now reads as unavailable. |
| raw `position_status.qty` | bypassed the dust rule. The issue catalogued five sites; **the structural guard found five more the moment it existed** — including one in okx whose own comment quoted the forbidden pattern while doing the right thing. |

The qty item is latent today (`log_return_reward` reads portfolio values, not quantities). That is precisely when a re-fork goes unnoticed, which is why it is a guard rather than a behavioural test.

## Not in this PR

**#339** turns out to touch the **offline** envs, not live — only `sequential.py` and `vectorized_sequential.py` consume `POSITION_TOLERANCE_ABS`. Changing a sizing threshold there is a behavioural change to the highest-priority environments and a scalar/vectorized parity surface. It gets its own PR rather than riding along in a live-env change.

## Verification

Every fix mutation-tested individually — including the one that revealed `#349.4` had zero coverage before the guard existed.

Full suite: **2788 passed, 19 skipped**. Diff: 15 files, +203/−49.

### DIFF
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -303,9 +303,13 @@ class TestBinanceFuturesTorchTradingEnv:
             env.step(TensorDict({"action": torch.tensor(long_idx)}, batch_size=()))
             mock_trader.trade.assert_called()
 
-            # 2. Exchange confirms the position. Re-commanding the SAME level is redundant.
+            # 2. Exchange confirms the position AT THE SIZE THAT WAS ORDERED. Re-commanding
+            # the SAME level is then redundant. Read from the call rather than hardcoded: a
+            # fixture reporting a different size is a partial fill, which the sync now
+            # detects and deliberately releases the guard for (#276 follow-up).
+            filled = mock_trader.trade.call_args.kwargs["quantity"]
             mock_trader.get_status = MagicMock(return_value={"position_status": PositionStatus(
-                qty=0.01, notional_value=500.0, entry_price=50000.0, unrealized_pnl=0.0,
+                qty=filled, notional_value=500.0, entry_price=50000.0, unrealized_pnl=0.0,
                 unrealized_pnl_pct=0.0, mark_price=50000.0, leverage=5,
                 margin_type="isolated", liquidation_price=45000.0,
             )})
--- a/tests/envs/bitget/test_torch_env_futures.py
+++ b/tests/envs/bitget/test_torch_env_futures.py
@@ -341,9 +341,13 @@ class TestBitgetFuturesTorchTradingEnv:
             env.step(TensorDict({"action": torch.tensor(long_idx)}, batch_size=()))
             mock_trader.trade.assert_called()
 
-            # 2. Exchange confirms the position. Re-commanding the SAME level is redundant.
+            # 2. Exchange confirms the position AT THE SIZE THAT WAS ORDERED. Re-commanding
+            # the SAME level is then redundant. Read from the call rather than hardcoded:
+            # a fixture reporting a different size is a partial fill, which the sync now
+            # detects and deliberately releases the guard for (#276 follow-up).
+            filled = mock_trader.trade.call_args.kwargs["quantity"]
             mock_trader.get_status = MagicMock(return_value={"position_status": PositionStatus(
-                qty=0.01, notional_value=500.0, entry_price=50000.0, unrealized_pnl=0.0,
+                qty=filled, notional_value=500.0, entry_price=50000.0, unrealized_pnl=0.0,
                 unrealized_pnl_pct=0.0, mark_price=50000.0, leverage=5,
                 margin_mode="isolated", liquidation_price=45000.0,
             )})
--- a/tests/envs/okx/test_torch_env_futures.py
+++ b/tests/envs/okx/test_torch_env_futures.py
@@ -303,6 +303,50 @@ class TestOKXFuturesTorchTradingEnv:
             f"a position opened after a liquidation inherited the dead position's age ({aged})"
         )
 
+    def test_a_partial_reduction_does_not_age_or_flip_the_position(self, env, mock_env_trader):
+        """#276 end to end: the harm was never the direction field, it was the chain.
+
+        Trimming a long 1.0 -> 0.5 sends a SELL, which was recorded as current_position
+        = -1 while the venue still held a half-size long. The NEXT bar's sync then found
+        a mismatch the env had inflicted on itself, discarded hold_counter and NaN'd
+        current_action_level -- so a 20-bar-old position reported holding_time=1 and the
+        duplicate-action guard never fired again.
+        """
+        from torchtrade.envs.live.okx.order_executor import PositionStatus
+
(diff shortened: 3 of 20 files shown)


BY571:
@CharlieHelps please review.

Closes #276 and #349. Round 1 of five review agents is done and its findings are applied; the most useful thing you can do is be adversarial about the new reconciliation, because round 1 showed the fix itself introduced a regression I did not anticipate.

Context on that, since it is the crux of the PR:

#276 was that five live envs inferred position direction from the **order side**. Under fractional sizing a SELL that trims a long from 1.0 to 0.5 leaves a long but was recorded as -1. Fixing it exposed something: the old bug was *accidentally* recovering from partial fills, because a wrong direction forced a sync mismatch every bar which NaN'd the action level and re-traded. Recording the direction correctly removes that recovery, and `_sync_position_from_exchange` reconciles direction only — its own docstring says so and calls it pre-existing.

So `_sync_position_from_exchange` now reconciles **size** as well. The env records the quantity the accepted action asked for (`PositionState.target_qty`), and a venue position that disagrees by more than `FILL_REL_TOL` releases the duplicate-action guard and logs it.

Please be sceptical about:

1. **`FILL_REL_TOL = 0.01` and the relative comparison.** I chose relative deliberately so the rule means the same thing on an asset priced in cents and one at $100k — that is exactly the denomination bug #339 is about, and I did not want to inherit it. But is 1% right? A venue that rounds to a coarse lot step on a small position could exceed 1% legitimately and release the guard every bar, which would be churn. Should this key off the venue lot step instead?

2. **`target_qty` plumbing.** It is attached to `trade_info` at the executing return of each `_execute_fractional_action`, and `_handle_close_action` reports `target_qty=0.0` so a partial *close* is caught by the same rule. Any path that trades without setting it leaves `target_qty=None`, which disables the check silently — that is the failure mode I am least sure I have covered.

3. **The alpaca SLTP asymmetry.** Its action map drops the side (long-only spot), so I derive direction from `action_tuple[0] is not None` rather than the `{"long": 1, "short": -1}` mapping the four futures SLTP envs use. Two different rules in files that otherwise mirror each other.

4. **Whether releasing the guard is the right response at all** versus halting. Right now a divergence logs a warning and lets the agent re-trade. An alternative reading is that a venue holding something other than what was ordered is an integrity failure that should stop the episode.

Round 1 also found that I had fixed all five `env.py` and left the identical bug in all five `env_sltp.py` — and that my own structural guard parametrized `env.py` only, so it passed vacuously over exactly the files I had missed. Both fixed. If you see more of that shape, it is the most valuable thing you can flag.

Every new guard is mutation-proven (revert it, confirm the rows die). Full suite: 2813 passed, 19 skipped.

Issue #314, opened 2026-08-10
Liquidation books at the liquidation price even when the bar gapped straight through it


BY571:
Follow-up from #311 (which fixed the same class of bug for stop-losses only).

## The bug

`_execute_liquidation` (`torchtrade/envs/offline/sequential.py:906`) and the vectorized `_apply_liquidation` both realise the loss at `self.liquidation_price`, regardless of where the bar actually opened. A bar that gaps straight through the liquidation price is booked as if it stopped there.

Measured on a long, entry 100, SL -2.5%, single gap bar:

| leverage | gap to | exit | balance | error |
|---|---|---|---|---|
| 2 | 85.0 | `sltp_sl` | 7000.00 | 0 (fixed by #311) |
| 5 | 85.0 | `sltp_sl` | 2500.00 | 0 (fixed by #311) |
| 20 | 96.0 | `sltp_sl` | 2000.00 | 0 (fixed by #311) |
| 20 | 90.0 | `liquidation` | 800.00 | books at liq_price (~95.4), not 90 |
| 20 | 85.0 | `liquidation` | 800.00 | same |

Once the gap is deep enough to reach the liquidation price, liquidation fires first (correct priority, per #299) and books at `liq_price`. At 20x a 10% gap wipes the account in reality; the env hands the agent back 800 and tells it that it survived. **The benefit of #311 shrinks as leverage rises**, which is the opposite of what you want.

## Why it was deliberately excluded from #311

Not an oversight — it is a genuine modelling question that a PR scoped to SL/TP should not answer by accident.

An exchange liquidation is not a resting order. On real venues an isolated-margin position's loss is **capped at its posted margin**, with the excess socialised (insurance fund / auto-deleveraging). So naively applying the #311 rule — fill at the open — could make the simulation lose *more than the margin*, i.e. more wrong, not less.

## What needs deciding

1. Does the sim model isolated margin (loss capped at margin) or cross (loss can exceed)? The config has a single `leverage` and no margin-mode switch.
2. If capped: the fix is `max(loss_at_open, -posted_margin)`, not `stop_fill_price`.
3. Whichever is chosen must be applied to **both** engines and pinned by the equivalence harness, or scalar and vectorized diverge.

## Scope

- `torchtrade/envs/offline/sequential.py::_execute_liquidation`
- `torchtrade/envs/offline/vectorized_sequential.py::_apply_liquidation`
[...]

PR #359, opened 2026-08-12
fix(offline,replay): incomplete bars, and replay's missing liquidation (#353, #269)

### LINKED ISSUE #269: Replay order executor: no liquidation; liquidation_price=0 pins distance_to_liquidation at 1.0
## Bug (two coupled defects)

1. \`torchtrade/envs/replay/order_executor.py:64-101\` — \`advance_bar\` checks only SL/TP. No counterpart to \`sequential.py\` liquidation (\`:279-296\`, \`:888-922\`) and no balance floor. Reproduced at 5x: equity goes to **-1000** and then fully recovers when price comes back. Offline env on identical input liquidates at 40200 and terminates. Any leveraged replay evaluation is unboundedly optimistic.
2. \`replay/order_executor.py:222\` hardcodes \`liquidation_price=0.0\`, so \`futures_live_base.py:95-96\` fails open and \`account_state[5]\` (distance_to_liquidation) reads **1.0 for every position** — a 5x levered position is presented to the policy as maximally far from liquidation. This is CLAUDE.md invariant #3 verbatim.

## Fix

Compute a real liquidation price with the same formula the offline env uses, check it in \`advance_bar\` before SL/TP, floor the balance, and return the price in \`get_status()\`. Regression tests: leveraged replay position must liquidate; \`account_state[5]\` must move.

### LINKED ISSUE #353: Sampler gap detection is close-only; a NaN open/high/low is silently forward-filled into the execution base
Found while reviewing #332. Pre-existing, and larger than that PR.

## The bug

`torchtrade/envs/offline/infrastructure/sampler.py`:

- `:228` — the DATA GAPS warning inspects **`close` only**: `nan_mask = execute_base_raw["close"].isna()`
- `:271` — the execution base is then forward-filled wholesale: `execute_base_filled = execute_base_raw.ffill()`

So a bar with a valid `close` but a missing `open`, `high` or `low` is:
1. **Not warned about** — the gap detector never looks at those fields
2. **Dropped from observations** (the resampled frames drop NaN rows)
3. **Kept in `execute_base_features_df`**, with the missing field carrying the **previous bar's** value

Step 3 is the damaging one. The execution base is what prices fills and answers level-touch questions, so a synthetic `open` and synthetic `high`/`low` are used as if they were real market data:

- `stop_fill_price()` (#280) prices a gapped stop at `open` — a forward-filled `open` produces a fill at a price the market never traded
- `_check_sltp_trigger` and `_check_liquidation` compare against `high`/`low` — forward-filled extremes answer "was this level touched?" with the previous bar's range

This is the same open-vs-extreme disagreement #326 was about, arriving through the NaN door rather than the malformed-OHLC door. The malformed path raises at ingestion; this path is silent.

## Why the malformed-OHLC validation doesn't catch it

The validation added for #326 tests `high < max(open, close)` and `low > min(open, close)`. NaN comparisons are False, so a row with a NaN field is never flagged as malformed — correctly, it isn't malformed, it's incomplete. Incompleteness has no check of its own.

## Suggested fix

Detect on any OHLC field rather than `close` alone, and decide explicitly what an incomplete bar means for the execution base — dropping it, or refusing to forward-fill price fields, rather than silently inheriting the previous bar's. Forward-filling volume is defensible; forward-filling a price that fills orders is not.

Whatever the choice, it should be stated in `docs/guides/sampler.md`, which currently implies incomplete rows are handled by gap detection.

## Related

#326 (malformed OHLC validation at ingestion), #280 (gap-aware stop fills), #332 (bounded diagnostics — where this surfaced), #278 (replay fidelity)

BY571:
Closes #353. Closes #269.

Both are fidelity bugs in the path that produces backtest numbers — the highest-priority surface per CLAUDE.md, since these train policies that trade real money.

## #269 — the replay executor had no liquidation at all

At 5x, equity ran to **−1000 and then fully recovered** when price came back. The offline env on identical input liquidates and terminates. **Any leveraged replay evaluation was unboundedly optimistic** — the one number a backtest exists to produce.

Two things kept it invisible:

- `advance_bar` returned early when `sl_price` and `tp_price` were both zero — **the default** — so a position with no bracket could never liquidate.
- `liquidation_price` was hardcoded `0.0`, which makes `futures_live_base` fail open and report `account_state[5]` as `1.0` for **every** position: a levered position handed to the policy as maximally far from liquidation. That is CLAUDE.md invariant 3 verbatim.

Liquidation is now checked **before** SL/TP, matching the offline env's priority (#299) — the venue closes the position whatever the agent's bracket says — and the balance is floored, because isolated margin caps the loss at the margin posted. The price comes from the shared `isolated_liquidation_price`, so replay and offline cannot answer the same question differently.

## #353 — gap detection was close-only, then the base was forward-filled wholesale

A bar with a good `close` and a missing `open`, `high` or `low` was never warned about and silently carried the **previous bar's** value. That base prices fills and answers level-touch questions, so:

- `stop_fill_price` could fill a gapped stop at a price the market never traded
- `high`/`low` answered "was this level touched?" with a range from one bar earlier

The empty-bar case was already handled by collapsing the range to that bar's close. An incomplete bar now gets the same treatment, for the same reason: forward-filling volume is defensible, forward-filling a price that fills orders is not, and its own close is the one price we actually know.

## Not in this PR

**#314** (liquidation books at the liquidation price even when the bar gapped through it) needs the isolated-vs-cross modelling decision its own issue sets out, and it has to land in **both** engines pinned by the equivalence harness. Answering it as a side effect of a sampler fix is exactly what that issue warns against.

## Verification

Every fix mutation-tested — reverting each turns the named rows red.

Full suite: **3099 passed, 19 skipped**.

### DIFF
--- a/tests/envs/offline/test_sampler_incomplete_bars.py
+++ b/tests/envs/offline/test_sampler_incomplete_bars.py
@@ -81,3 +81,28 @@ def test_a_stale_close_cannot_land_outside_its_own_bar():
     assert bar["low"] <= bar["close"] <= bar["high"], (
         f"close {bar['close']} sits outside its own bar [{bar['low']}, {bar['high']}]"
     )
+
+
+@pytest.mark.parametrize("missing,other", [("high", "low"), ("low", "high")])
+def test_the_collapse_cannot_reintroduce_a_malformed_bar(missing, other):
+    """The fix broke the invariant it was protecting.
+
+    A bar whose `open` is real and whose `high` is missing takes close as its high, which
+    can land BELOW open -- exactly the malformed shape the ingestion validator rejects on
+    raw input, recreated from the inside where no validator runs.
+    """
+    df = _frame()
+    df.loc[201, ["open", "high", "low", "close"]] = [210.0, 220.0, 195.0, 200.0]
+    df.loc[201, missing] = np.nan
+    when = df.loc[201, "timestamp"]
+
+    with pytest.warns(UserWarning, match="DATA GAPS"):
+        sampler = MarketDataObservationSampler(
+            df, time_frames=[TimeFrame(1, TimeFrameUnit.Minute)], window_sizes=[10],
+            execute_on=TimeFrame(1, TimeFrameUnit.Minute),
+        )
+
+    bar = sampler.execute_base_features_df.loc[when]
+    assert bar["high"] >= max(bar["open"], bar["close"]), "high below a real price"
+    assert bar["low"] <= min(bar["open"], bar["close"]), "low above a real price"
+    assert bar[other] == df.loc[201, other], "the extreme that WAS present was discarded"
--- a/tests/envs/offline/test_sampler_incomplete_bars.py
+++ b/tests/envs/offline/test_sampler_incomplete_bars.py
@@ -1,0 +1,35 @@
+import numpy as np, pandas as pd, pytest
+from torchtrade.envs.offline.infrastructure.sampler import MarketDataObservationSampler
+from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
+
+
+def _frame(n=400):
+    idx = pd.date_range("2024-01-01", periods=n, freq="1min")
+    return pd.DataFrame({"timestamp": idx, "open": 100.0, "high": 101.0,
+                         "low": 99.0, "close": 100.0, "volume": 10.0})
+
+
+@pytest.mark.parametrize("field", ["open", "high", "low"])
+def test_an_incomplete_bar_does_not_inherit_the_previous_bar_s_price(field):
+    """#353: gap detection was close-only, then the base was forward-filled wholesale.
+
+    A bar with a good close and one missing price field was never warned about and
+    silently carried the PREVIOUS bar's value -- so stop_fill_price could fill a gapped
+    stop at a price the market never traded, and high/low answered "was this level
+    touched?" with a range from one bar earlier.
+    """
+    df = _frame()
+    df.loc[200, ["open", "high", "low", "close"]] = [200.0, 205.0, 195.0, 200.0]
+    df.loc[201, field] = np.nan
+    stale = df.loc[200, field]
+    when = df.loc[201, "timestamp"]
+
+    with pytest.warns(UserWarning, match="DATA GAPS"):
+        sampler = MarketDataObservationSampler(
+            df, time_frames=[TimeFrame(1, TimeFrameUnit.Minute)], window_sizes=[10],
+            execute_on=TimeFrame(1, TimeFrameUnit.Minute),
+        )
+
+    bar = sampler.execute_base_features_df.loc[when]
+    assert bar[field] != stale, f"{field} inherited the previous bar's price ({stale})"
+    assert bar[field] == bar["close"], "an incomplete bar should collapse to its own close"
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps please review.

Closes #353 and #269 — both fidelity bugs in the path that produces backtest numbers.

Worth your scepticism:

1. **The isolated-margin assumption in replay.** I floor the balance at 0 on liquidation, on the grounds that isolated margin caps loss at posted margin. The config has a single `leverage` and no margin-mode switch, and `SequentialTradingEnvConfig.margin_type` is stored but never read — so "isolated" is an inference from behaviour, not a stated contract. If cross is ever modelled, this floor is wrong.

2. **Liquidation before SL/TP in replay.** I matched the offline env (#299). Please confirm that ordering is right for replay specifically, where the bracket is simulated rather than resting at a venue.

3. **The incomplete-bar rule.** I collapse `open`/`high`/`low` to that bar's own `close`. The alternative is dropping the bar entirely. Collapsing keeps the timeline dense but reports a zero-range bar the market did not trade; dropping loses an execution opportunity. I chose collapsing to match the existing empty-bar treatment — but it is a judgement call and a second opinion is worth having.

4. **Whether #314 should have been in here.** I deliberately left it out: it needs the isolated-vs-cross decision and has to land in both engines under the equivalence harness. If you think the three belong together, say so.

Every fix is mutation-tested. Full suite: 3099 passed.

BY571:
@CharlieHelps two review rounds are applied — re-requesting your look, since every round found a defect in the *previous* round's fix rather than in the original issues.

What changed since you were first asked:

1. **I cited a superseded rule.** Replay liquidated unconditionally "matching #299"; **#300 replaced that** — a stop on the same side of entry as liquidation is crossed first, so it wins. On one bar at 10x: replay left 400 where offline leaves 5000. The rule now lives in `liquidation.py` as `stop_precedes_liquidation` and both engines call it.

2. **I booked liquidations at the wrong price.** `stop_fill_price` is right for a stop and wrong for a liquidation, which the venue books at the bankruptcy price with its insurance fund absorbing the gap. Filling at the gapped open lost 4x the posted margin. The `max(balance, 0)` clamp I had added was a symptom and went with it.

3. **My sampler fix was a regression, twice.** First a row mask discarded the fields that were *present* (a bar missing only its open became flat — a backtest that skips losing exits). Then repairing the OHLC invariant by widening with the *stale* close fabricated prices the bar never printed, firing stops on bars that never traded there. It now widens from `open` only, a real print of that bar. Zero malformed shapes across all 15 NaN combinations.

Specific things I would value your scepticism on:

- **The widening choice.** A reviewer proposed collapsing the whole bar when `close` is missing. I tried it and `test_bar_without_a_usable_close_keeps_its_real_range` failed — an existing deliberate contract. Widening from `open` satisfies both, but if you think the contract itself is wrong, that is a better conversation than my compromise.
- **`stop_precedes_liquidation` keeps `not (open <= liq)` rather than `open > liq`.** They differ only on a NaN open (90 divergences over a 2662-combo grid). I kept the offline spelling so the extraction is pure — but that means a NaN open is treated as "did not gap", which is the optimistic answer. Unreachable today; still a decision someone should make deliberately.
- **The balance clamp removal** is only safe while `transaction_fee <= mmr x entry/liq`. Outside that, replay reports a negative balance where offline clamps and logs "likely an accounting bug". I have not fixed that asymmetry — flagging rather than hiding it.

Full suite: 3113 passed, 19 skipped. Every fix mutation-tested.

Issue #343, opened 2026-08-12
Live envs have no fail-safe when _get_observation raises while a position is open


BY571:
Raised by @CharlieHelps reviewing #342, but the problem predates it.

## The gap

`_get_observation()` can raise, by design, rather than fabricate an account state:

- `PositionUnknownError` when the exchange's position status is unknown (`core/state.py`) — pre-existing
- `ValueError` when a position is held against non-positive equity (#342)
- `ValueError`/`KeyError` when a venue omits leverage, entry price or the equity key (#342)

Every futures `_step()` calls `_get_observation()` after waiting for the next bar, and none of them catch it — e.g. `torchtrade/envs/live/binance/env.py:182-187`. So the raise propagates out of `step()` **before** history, reward and termination handling, and the process is left holding a live position with no built-in close or halt path.

Raising is the right call at the observation layer: the alternative is reporting a held position as flat, which is invariant #3 and the bug #277 was about. The missing piece is at the caller.

## What's needed

A fail-safe around the live loop, roughly:

1. Catch observation failures at the env or runner boundary.
2. Distinguish transient (retry with backoff) from terminal (equity gone, venue liquidating).
3. On terminal: flatten the position or halt deliberately, and emit a real `terminated` so the episode ends through the normal path rather than by exception.
4. Make the behaviour configurable — an operator running paper may want to halt where a live operator wants to flatten.

Related: #295 (stale-but-marked observation so a transient blip need not end the episode) covers the retry half of step 2. This issue is the other half — what to do when it is *not* transient.

## Why not fixed in #342

#342 narrowed the reachability (adapters that coerced a missing equity field to `0` now raise at the boundary, so the observation-layer raise fires on real states rather than partial responses), but a close/halt path is a live-runner feature affecting all four exchanges and every existing raise — not something to bolt onto a fail-open fix.

PR #360, opened 2026-08-12
fix(live): the halt covered the read after the trade, not the one before it (#355)

### LINKED ISSUE #355: Live halt covers only the post-bar read; the pre-trade and reset reads bypass it
Found reviewing #346, which introduces the halt. Not a regression — the gap is in coverage, and #346's docs now state the boundary rather than implying it is universal.

## The gap

`_acquire_post_bar_state()` wraps the portfolio-value and observation reads taken *after* the bar closes, converting a terminal venue failure into `LiveObservationHalt` after an optional emergency `close_position()`.

Each `_step()` makes **two** other venue reads that are not wrapped:

1. **The pre-trade read at the top of `_step()`** — `status = self.trader.get_status()`, used for `_sync_position_from_exchange()` and the duplicate-action guard. Reproduced: a `POSITION_UNKNOWN` between bars raises a bare `PositionUnknownError` here, so a caller doing `except LiveObservationHalt` (as the docs and the DQN example show) does not catch it, and **no emergency flatten runs even under `observation_failure_policy=FLATTEN`**. This read happens *before* the env trades.

2. **The initial read inside `_reset()`** — `self._get_observation(advance_hold=False)` in all four `base.py`. Same result: bare exception, no latch, no flatten. This is the resume-against-an-existing-position case, which is the scenario CLAUDE.md invariants #2 and #3 care about most.

## Also

- **`ParallelEnv` loses the exception.** A halt inside a worker kills it; the parent re-raises a generic `RuntimeError("Cannot proceed, worker N dead")`, so both the type and `flatten_accepted` are lost. Only matters if live envs are ever batched multi-process, but the docs describe a universal contract.
- **Alpaca and Polymarket do not route through the halt at all** — 2 of 10 live envs, plus polymarket.
- **`advance_hold_counter()` runs before the finiteness checks** inside `_get_observation`, so a halt on an open position still bumps `hold_counter`. Harmless while "never step again after a halt" holds, since `_reset` zeroes it — latent only if that contract is broken.

## Suggested shape

Route the pre-trade and reset reads through the same helper rather than adding three more wrapped call sites per env — the point of #343 is that *any* failure to read account state while a position is open must halt, not just the post-bar one.

## Related

#343 (the original issue), #346 (the post-bar halt), #295 (transient vs terminal classification, which this depends on for retry)

BY571:
Closes #355.

#346 wrapped the post-bar read. Two other venue reads per step were not wrapped, and **the pre-trade one is the more consequential**.

## The gap

A `POSITION_UNKNOWN` between bars raised a bare `PositionUnknownError` from the read at the top of `_step()`. So:

- a caller doing `except LiveObservationHalt` — **what the docs and the shipped DQN example both show** — did not catch it
- **no emergency flatten ran, even under `observation_failure_policy=FLATTEN`**

That read happens *before* the env trades, which is precisely when a flatten still has options.

The initial read in `_reset()` had the same gap. That is the resume-against-an-existing-position case — the one CLAUDE.md invariants 2 and 3 care about most.

## Shape

One shared `_halting(read)` on the futures base, rather than three wrapped call sites per env. #343's rule is that **any** failure to read account state while a position is open must halt, so the policy belongs on the reading, not on one of its callers.

The contract test's expectation changes with it: an outage now surfaces as `LiveObservationHalt` rather than the bare exception. That is the point of the issue, not a side effect.

## Not in this PR

**#295** adds an observation key (`status_unknown`) and truncates on a prolonged outage. Its own text says it *"invalidates policies trained before it and needs its own review"*, and its item 3 — whether `SyncDataCollector`'s `auto_reset` handles `truncated`-without-`terminated` for these stateful non-vectorized envs — is explicitly unverified. That needs a decision and a verification pass, not a side effect of a plumbing fix.

## Verification

Structural guard across all 8 futures envs plus a behavioural test that the configured policy actually runs on a failed read. Mutation-proven: unwrapping one exchange turns exactly that exchange red.

Full suite: **3123 passed, 19 skipped**.

### DIFF
--- a/tests/envs/okx/test_torch_env_futures.py
+++ b/tests/envs/okx/test_torch_env_futures.py
@@ -378,7 +378,10 @@ class TestOKXFuturesTorchTradingEnv:
             env.reset()
             mock_env_trader.get_status = MagicMock(return_value=status(mark_price))
             td = TensorDict({"action": torch.tensor(len(env.action_levels) - 1)}, [])
-            with pytest.raises(ValueError, match="mark.?price"):
+            # LiveObservationHalt now, not a bare ValueError: the pre-trade read runs
+            # under the halt policy (#355), so an unusable mark surfaces the same way an
+            # unreadable position does. Both are RuntimeError subclasses.
+            with pytest.raises(RuntimeError, match="mark.?price"):
                 env.step(td)
 
         mock_env_trader.trade.assert_not_called()
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -27,7 +27,11 @@ import torch
 from tensordict import TensorDict
 
 import torchtrade.envs  # noqa: F401  -- registers every live env as a subclass
-from torchtrade.envs.core.live import TorchTradeLiveEnv
+from torchtrade.envs.core.live import (
+    LiveObservationHalt,
+    ObservationFailurePolicy,
+    TorchTradeLiveEnv,
+)
 from torchtrade.envs.live.shared.futures_live_base import TorchTradeFuturesLiveEnv
 from torchtrade.envs.utils.sltp_mixin import SLTPMixin
 from torchtrade.envs.utils.liquidation import (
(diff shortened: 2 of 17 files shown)


BY571:
@CharlieHelps please review.

Closes #355 — the halt introduced in #346 wrapped the read *after* the bar; the read at the top of `_step()`, which happens **before** the env trades, was unwrapped.

Worth your scepticism:

1. **The shape.** I put the policy on one shared `_halting(read)` on the futures base rather than wrapping three call sites per env. My reasoning: #343's rule is "any failure to read account state while a position is open must halt", and expressed as "wrap these particular calls" that is only true while every future author remembers. But it does mean an exception type changes at eight call sites at once — please check none of those callers relied on catching `PositionUnknownError` specifically.

2. **The contract test's expectation changed.** `test_an_outage_stops_the_step_before_it_can_trade` now accepts `LiveObservationHalt` as well as `PositionUnknownError`. That is the intended behaviour change, but it is exactly the kind of test edit that can hide a regression — please confirm it still fails for the right reason.

3. **Alpaca and Polymarket still do not route through the halt at all.** #355 names this (2 of 10 live envs). I did not extend it there because alpaca has no `observation_failure_policy` on its config, so adding one is a config-surface change rather than plumbing. If you think that should have been in scope, say so.

4. **`_reset`'s read now halts too.** That is the resume-against-an-existing-position case. Worth checking the halt is safe *during* reset specifically — a flatten there acts on a position the env has not yet recorded.

Not in this PR: **#295**, which adds an observation key and truncates on prolonged outages. Its own text says it invalidates policies trained before it, and its item 3 (whether `SyncDataCollector`'s `auto_reset` handles `truncated`-without-`terminated` for these stateful non-vectorized envs) is explicitly unverified.

Full suite: 3123 passed, 19 skipped. Mutation-proven — unwrapping one exchange turns exactly that exchange red.

Issue #336, opened 2026-08-11
Two #290 decisions left open: unread OrderStatus fields, and test-only public API


BY571:
Split out of #290 so it could close honestly — #335 fixed the four defects on that list; these two are decisions rather than defects.

## 1. `OrderStatus.order_id` / `order_type` are constructed but never read

`torchtrade/envs/core/common_types.py` defines them; bitget populates them at `order_executor.py:536`. Nothing reads either field anywhere in the repo.

Decide: keep as a deliberate public API for downstream users (and say so in the docstring), or drop them and the construction cost.

## 2. Public API referenced only by its own test

Each of these is public and has exactly one caller — the test that covers it:

- `MarketDataObservationSampler.get_sequential_observation_with_ohlcv`
- `MarketDataObservationSampler.get_max_steps`
- `CoverageTracker.get_coverage_distribution`
- `compute_periods_per_year_crypto`
- alpaca `get_clock`
- the futures executors' `set_leverage` / `set_margin_mode` public wrappers

Same decision per item: intended API surface, or dead code with a test keeping it alive? The latter is the shape the repo has been cleaning up (a test that exists only to give an unused function a caller pins nothing).

Worth doing as one pass rather than piecemeal, since the answer is likely the same for most of them.

PR #361, opened 2026-08-12
fix(offline): denominate the flat-enough threshold in dollars, not base units (#339)

### LINKED ISSUE #339: POSITION_TOLERANCE_ABS is denominated in base-asset units

BY571:
Closes #339.

`POSITION_TOLERANCE_ABS = 0.001` meant 0.001 of the **base asset**, so one constant was worth wildly different things:

| asset | price | worth of unclosed position |
|---|---|---|
| BTC | $100,000 | **$100.00** — 1% of a $10k account, read as flat |
| ETH | $3,500 | $3.50 |
| SOL | $200 | $0.20 |
| DOGE | $0.35 | $0.0003 |
| a cent asset | $0.01 | $0.0000 |

A factor of ~300,000 between the ends, from a constant that reads like a small number.

It is `POSITION_TOLERANCE_NOTIONAL = 1.0` now, converted at the comparison site. The 2% relative term still dominates on any position large enough to matter, so this only moves behaviour where the old constant was meaningless.

Applied to **both** engines — the equivalence harness pins them together, and a threshold that differs between scalar and vectorized is worse than one that is wrong in both.

## The test was vacuous first, and mutation caught it

My first version computed `POSITION_TOLERANCE_NOTIONAL / price` itself and asserted the round-trip. That is true of *any* constant, so restoring the base-unit floor left it green. It drives the real engine now: a residual worth fifty cents is flat on BTC and on DOGE; one worth fifty dollars is flat on neither. The mutant dies.

Full suite: **3131 passed, 19 skipped**.

### DIFF
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -427,26 +427,6 @@ class TestPositionStatusDataclass:
         assert pos.leverage == 10
 
 
-class TestOrderStatusDataclass:
-    """Tests for OrderStatus dataclass."""
-
-    def test_order_status_creation(self):
-        """Test creating OrderStatus."""
-        from torchtrade.envs.live.binance.order_executor import OrderStatus
-
-        order = OrderStatus(
-            is_open=False,
-            order_id="12345",
-            filled_qty=0.001,
-            filled_avg_price=50000.0,
-            status="FILLED",
-            side="BUY",
-            order_type="MARKET",
-        )
-
-        assert order.is_open is False
-        assert order.filled_qty == 0.001
-
 
 class TestBinanceLotSizeRounding:
     """#271: binance parsed only PRICE_FILTER, so every quantity went out as
--- a/tests/envs/offline/test_position_tolerance.py
+++ b/tests/envs/offline/test_position_tolerance.py
@@ -143,3 +143,64 @@ def test_the_vectorized_floor_holds_a_sub_dollar_resize(gap_usd, expect_hold):
     assert held is expect_hold, (
         f"a ${gap_usd} gap was {'not ' if expect_hold else ''}held by the floor"
     )
+
+
+@pytest.mark.parametrize("price", [100_000.0, 100.0, 0.35, 0.01],
+                         ids=["btc", "eth-ish", "doge", "cent"])
+def test_both_engines_open_a_sub_dollar_position_from_flat(price):
+    """An open from flat is not a resize, and the floor must not refuse it.
+
+    The scalar gate was unconditional while the vectorized gates on `has_position`, so
+    from flat the floor refused the scalar open and let the vectorized one through -- the
+    policy saw a long in one engine and nothing in the other. Worse on the scalar side it
+    was absorbing: with the portfolio under $1 the agent could never re-enter at all.
+
+    Every fixture in the equivalence harness holds a position before stepping, so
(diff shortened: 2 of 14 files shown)


BY571:
@CharlieHelps please review.

Closes #339 and #336.

Worth your scepticism:

1. **#339 changes which orders get placed.** The tolerance floor moved from 0.001 base units to $1 notional. On BTC that tightens it ~100x (from $100 to $1); on a cent asset it loosens it enormously. The 2% relative term still dominates on large positions, so I believe this only moves behaviour where the old constant was meaningless — but it is a live sizing threshold and the offline envs train policies, so a second opinion on the direction is worth having.

2. **Both engines had to change together.** The equivalence harness pins scalar to vectorized; a threshold that differs between them is worse than one wrong in both. Please confirm the vectorized form (`torch.maximum` with `execution_prices.clamp(min=1e-12)`) is genuinely equivalent to the scalar `max(...)`, including at price 0.

3. **#336's deletions rest on greps.** I have shown the counts in the PR body, but a caller I cannot see — a downstream user reading `OrderStatus.order_id`, or a dynamic lookup — would make these wrong. The `order_id` case is the one to check: a bare grep says 25 hits and the truth is zero, which is exactly the shape that fools a careful reader.

4. **I deleted 16 tests.** They existed only to call the three removed methods. If any of them was actually covering something else incidentally, that coverage is gone.

Full suite: 3113 passed, 19 skipped.

BY571:
Closes #336 as well as #339 — this branch now carries both.

**#336, both decisions resolved as deletions after measuring:**

`OrderStatus.order_id` / `order_type` — constructed at three sites (binance, bitget, alpaca), read **nowhere**. Deleted with their construction.

Worth recording how nearly this went wrong: a bare grep for `order_id` returns **25 production hits**, which reads as heavily used. They are all other things — `trade()`'s `order_type` parameter, `last_order_id`, ccxt payload keys. Only a grep scoped to the dataclass and its construction sites gives the real answer, which is zero.

Public methods whose only caller is their own test — the issue assumed one answer for the list; measuring split it in two:

| symbol | production callers | verdict |
|---|---|---|
| `get_sequential_observation_with_ohlcv` | 0 | deleted |
| `get_max_steps` | 0 | deleted |
| `get_coverage_distribution` | 0 | deleted |
| `compute_periods_per_year_crypto` | 2 | **kept** — never dead |
| `get_clock` | 2 | **kept** |
| `set_leverage` / `set_margin_mode` | real callers | **kept** |

The 16 tests that existed to give the three a caller went with them, plus a constructor test asserting the dataclass assigns what it was handed.

Issue #288, opened 2026-08-07
Deduplicate the four futures exchange stacks (~2,000 LOC, 70-96% copy-paste) into the shared futures base


BY571:
Measured identical lines after exchange-name substitution: okx vs bybit — env.py 253/278 (91%), env_sltp.py 321/339 (95%), base.py 216/224 (96%); binance vs bitget env_sltp.py 388/436 (89%). This is the pattern that produced the historical triple-SLTP-action-map bug, and this audit found three fresh instances of a fix landing on some exchanges but not others (lot size #271, full_done_spec #272, hedge-mode surface).

Plan (each step extends the structural-guard tests in \`tests/envs/test_live_env_base.py\` so re-forking fails CI):
1. Fold the 4 per-exchange \`base.py\` files into \`TorchTradeFuturesLiveEnv\` via \`OBSERVER_CLS\`/\`TRADER_CLS\` class attributes — the pattern \`futures_live_base.py\` already proves (~600-700 LOC)
2. Share \`env.py\` trade-path helpers: 4 byte-identical \`_create_trade_info\`, near-identical \`_step\`/\`_execute_market_order\`/\`_handle_close_action\`/\`_execute_fractional_action\`; per-exchange facts reduce to taker fee + lot-size rounding (~500 LOC)
3. Share \`env_sltp.py\` incl. the 70-line config dataclass repeated 4x (~900 LOC)
4. Consolidate 3 verbatim \`_calculate_unrealized_pnl_pct\` and 3 verbatim \`_round_price\` copies in the order executors
5. Mirror in tests: route more per-exchange test files through the existing shared \`tests/envs/base_exchange_tests.py\` (okx test file is 392/506 lines identical to bybit's)

Also fold in the small ceremony cleanups: 20-alias \`unit_to_timedelta\` map in \`core/live.py:127-153\` (store TimeFrameUnit instead of str), redundant abstract re-declarations, defensive action-index cascades guarding \`Categorical\`-spec output.

BY571:
## Step 5 measured — it is much smaller than the line count suggests, and mostly already done

The issue cites "okx test file is 392/506 lines identical to bybit's". That is a **line-level** count; at the function level the overlap is far smaller. AST-comparing every function after normalising exchange names:

```
okx fns: 29   bybit fns: 35   identical-after-rename: 6
  mock_observer, mock_observations          <- fixtures
  test_step_emits_the_whole_done_family     <- ALREADY delegates to the shared harness
[...]

PR #362, opened 2026-08-12
fix(live): make the venue confirm the leverage before sizing against it (#277)

### LINKED ISSUE #277: Live: leverage and distance_to_liquidation fail open (swallowed set_leverage failures; empty liqPrice → 1.0)

BY571:
Closes #277.

## The defect

All four futures exchanges wrapped their set-leverage call in `except Exception: logger.warning("may already be configured")`. `self.leverage` then stayed at the configured value whatever the venue did — wrong twice over:

- every position size is computed from leverage the account may not have (config 20x against an account still at 5x sizes 4x the available margin);
- `account_state[4]` reports the config value while flat but the venue's value once a position opens, so the element the policy conditions on flips between two numbers with no trade in between.

## The fix

Leverage raises where position mode and margin mode still warn. Those two are preferences; leverage sizes money.

The carve-out is what makes that shippable: every venue reports *"already at this leverage"* as an API error, and that is the case the blanket handler existed to absorb. Recognising it narrowly (`is_already_applied`) is what lets everything else raise.

Where the venue echoes the leverage it applied (binance, bitget, okx) the echo is checked too — accepting the call is not applying the request, and binance caps leverage per notional bracket.

## What this does NOT cover

A venue that accepts the call, applies something else, and echoes nothing. Catching that needs a read-back, and OKX and Bitget return no position row at all when flat — so it needs their leverage-info endpoints, which I cannot verify against a real API from here. Left explicit rather than guessed at on a money path.

**Item 2 of #277** (empty `liqPrice` defaulting `distance_to_liquidation` to 1.0) was already fixed in the shared futures base by an earlier PR; only the leverage half remained.

## Incidental

Removing bitget's outer blanket handler exposed a latent `NameError`: `margin_mode_ccxt` was assigned *inside* the `try` it was referenced from, so a failure in `to_ccxt()` raised from the error path — invisible only because the outer handler caught that too. Hoisted above the try.

Test doubles that omitted the leverage API (and mocks returning a constant `{"leverage": 10}` regardless of request) now echo the request, as a venue does.

## Evidence

16 cases across 5 parametrized functions, built on **real** order executors — the defect lived in each executor's `_setup_futures_account`, so a fake trader would prove nothing about it. Mutation-verified:

| mutant | cells killed |
|---|---|
| neuter the echo check | exactly the 3 echoing venues |
| swallow refusals again | all 6 refusal cells, all four exchanges |
| drop the already-applied carve-out | the 4 rerun cells |

Full suite: 3146 passed.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -42,6 +42,16 @@ Non-SLTP environments use `action_levels` for fractional position sizing — sam
 
 Live environments use a **query-first pattern**: they query the actual exchange position (source of truth), calculate the target based on the action, round to exchange constraints (lot size, min notional), and trade only the delta.
 
+!!! warning "Leverage is verified against the exchange at construction"
+    Futures environments refuse to start if the exchange will not apply the configured
+    `leverage` — a rejected request, or an applied value that differs from the one asked
+    for. Both would otherwise size every position against leverage the account does not
+    have, and make `account_state[4]` report the configured value while flat but the
+    exchange's value once a position is open.
+
+    An exchange that is *already* at the requested leverage reports that as an error;
+    that case is recognised and is not a failure.
+
 !!! warning "Timeframe Format - Critical for Model Compatibility"
     Always use canonical timeframe forms:
 
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -51,7 +51,9 @@ Live environments use a **query-first pattern**: they query the actual exchange 
 
     A successful construction guarantees **leverage only**. Margin mode is still
     best-effort, and Bybit's set-leverage response carries no leverage, so on Bybit an
-    accepted-but-clamped value is not detected.
+    accepted-but-clamped value is not detected. Passing your own `trader=` skips the
+    handshake entirely — the environment still sizes from `config.leverage`, so an
+    injected executor must already be at that leverage.
 
 !!! warning "Timeframe Format - Critical for Model Compatibility"
     Always use canonical timeframe forms:
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -43,14 +43,15 @@ Non-SLTP environments use `action_levels` for fractional position sizing — sam
 Live environments use a **query-first pattern**: they query the actual exchange position (source of truth), calculate the target based on the action, round to exchange constraints (lot size, min notional), and trade only the delta.
 
 !!! warning "Leverage is verified against the exchange at construction"
-    Futures environments refuse to start if the exchange will not apply the configured
-    `leverage` — a rejected request, or an applied value that differs from the one asked
-    for. Both would otherwise size every position against leverage the account does not
-    have, and make `account_state[4]` report the configured value while flat but the
-    exchange's value once a position is open.
+    Futures environments refuse to start if the exchange rejects the configured
+    `leverage`, or reports applying a different one. Either would otherwise size every
+    position against leverage the account does not have, and make `account_state[4]`
+    report the configured value while flat but the exchange's value once a position is
+    open.
 
-    An exchange that is *already* at the requested leverage reports that as an error;
-    that case is recognised and is not a failure.
+    A successful construction guarantees **leverage only**. Margin mode is still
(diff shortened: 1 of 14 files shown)


BY571:
@CharlieHelps please review this one.

Focus areas where I'd most value a second opinion:

1. **The fail-open I left deliberately.** `require_leverage_applied()` returns early when the venue echoes nothing. Catching "accepted the call, applied something else, echoed nothing" needs a read-back, and OKX/Bitget return no position row when flat — so it needs their leverage-info endpoints, which I can't verify against a real API from here. Is leaving that gap the right call, or should this PR add the read-back anyway?

2. **Substring matching on venue error text** (`_ALREADY_APPLIED`). Every venue reports "already at this leverage" as an API *error*, and that carve-out is what lets everything else raise. Is text matching too fragile here? Are the codes right — bybit 110043, binance -4046?

3. **Raising at construction.** A live env now refuses to start if the venue won't apply the configured leverage. Correct for a money path, or too strict — should it degrade to the venue's actual leverage instead?

4. Whether the four near-identical `_setup_futures_account` edits should share more now, or wait for #288 (the pending dedup of these four stacks).

BY571:
Rounds 1 and 2 applied — both found real defects in the fix itself. Summary for review:

**Round 1** (5 agents, 4 found defects)
- `bitget`'s check was **inert**: ccxt returns the applied leverage per side under `data`, with no top-level `leverage` key. It passed only because I had reshaped the *mock* to match my code rather than the venue.
- `-4046` is `NO_NEED_TO_CHANGE_MARGIN_TYPE` — a margin-type code. Binance's leverage endpoint is idempotent and needs no carve-out; the allowlist asserted a false venue semantic on a money path.
- The numeric markers were fail-open: `110043` is a substring of ms timestamps like `1762110043251`, which ccxt embeds in error text.
- The four public `set_leverage()` methods were dead **and** carried #277 verbatim. Deleted (−98).
- `isinstance(response, dict)` was fail-open at the boundary being hardened.

**Round 2**
- I had fixed the *instance, not the mechanism*: an **absent** echo field still returned happy, so a renamed key would silently stop verifying on three venues. The field is required now.
- **Ordering bug**: bitget stores leverage per margin mode, and `set_margin_mode` ran *after* the verified `set_leverage`. On bybit `switch_margin_mode` carries buyLeverage/sellLeverage and was tolerated with a warning, so it could silently undo the verified set. Margin mode now runs first on both.
- The **crossed-margin branch was executed by no test** — renaming `crossMarginLeverage` left 206 tests green. Covered now.

26 cells, every one mutation-verified. Full suite 3150 passed. Round 3 running.

@CharlieHelps the questions in my earlier comment still stand — particularly whether raising at construction is right for a money path, and whether the remaining bybit gap (its response carries no leverage, so an accepted-but-clamped value is undetected) should block.

PR #364, opened 2026-08-12
fix(offline): book a liquidation where the bar traded, capped at posted margin (#314)

### LINKED ISSUE #314: Liquidation books at the liquidation price even when the bar gapped straight through it

BY571:
Closes #314.

## The bug

`_execute_liquidation` and the vectorized `_apply_liquidation` realised the loss at `self.liquidation_price` regardless of where the bar opened. A bar gapping straight through it was booked as if it had stopped there.

At 20x a 10% gap wipes the account in reality; the env handed back 800 and told the agent it survived. **The benefit of #311's gapped-stop fix shrank as leverage rose** — the opposite of what you want, and it understated tail risk exactly where leverage makes it largest.

## The modelling decision

#314 asks whether the sim models isolated or cross margin. I answered **isolated**, because the rest of the repo already does:

- `MarginMode.ISOLATED` is the default across the live configs
- `account_state`'s `distance_to_liquidation` uses the isolated formula
- #359 fixed replay to book at the liquidation price *precisely so* a loss could not exceed posted margin

So the fix is not simply #311's rule: fill where the bar actually traded, **then cap the loss at the margin the position posted**. The venue closes at the bankruptcy price and the insurance fund absorbs the rest. Applying the gap rule without the cap would overshoot the other way and lose more than the account ever committed — more wrong, not less.

**The cap binds only on a gap.** At the liquidation price itself the loss is the margin less the maintenance buffer by construction, so every non-gapped case is unchanged.

## Scope

Reuses `stop_fill_price` rather than growing a second gap rule; the vectorized engine spells the same min/max with `torch.where`. All call sites updated — `sequential`, `onestep`, `sequential_sltp`, `vectorized_sequential(+SLTP)`. The base vectorized env now reads the open column it never needed before.

## Evidence

The **equivalence harness did its job**: patching the scalar alone failed `test_both_envs_agree_which_exit_price_reaches_first` on both directions, which is exactly the scalar/vectorized divergence #314 warned about.

`test_a_bar_that_gapped_past_liquidation_still_liquidates` had pre-registered this outcome — its `400` was labelled *"an #314 number … when that is fixed this expectation moves with it."* It moves to **0.0**: a 15% adverse gap at 10x is 1.5× the posted margin, so the account is gone.

Full suite: 3152 passed.

⚠️ **Review rounds not yet run** — opening now, will run all five agents for 3 rounds before merge.


BY571:
@CharlieHelps please review.

The one thing I'd most value your judgement on: **#314 explicitly left a modelling question open** and I answered it myself rather than blocking.

The issue asks whether the offline sim models **isolated** margin (loss capped at posted margin) or **cross** (loss can exceed it). The config has a single `leverage` and no margin-mode switch. I chose isolated, on three existing signals:

- `MarginMode.ISOLATED` is the default across the live exchange configs
- `account_state`'s `distance_to_liquidation` uses the isolated formula
- #359 deliberately fixed replay to book at the liquidation price *precisely so* a loss could not exceed posted margin

If that's wrong, the change is localised to two clamps (`max(loss, -margin)` scalar, `torch.maximum(pnl, -margin_return)` vectorized).

Other things worth your eye:

1. **The naive fix would have been wrong.** #311's rule ("a gapped bar fills at the open") applied directly to liquidation would lose *more than the margin ever posted* — an error in the opposite direction. The correct rule is the composition: fill where the bar traded, **then** clamp to margin. I'd like a second opinion that the composition is right and the clamp is against the right quantity (margin? margin minus fees? the bankruptcy price?).

2. **The blast radius is the tail.** This only changes gapped bars, and it makes them strictly worse for the agent — at 10x a 15% gap now wipes the account (balance 0.0) where it previously returned 400. Policies trained against the old behaviour were being told they survived events that would have ended them.

3. **No new tests yet.** The diff only moves one pre-registered expectation. My test-justifier is checking whether the margin cap has any killer at all; if it doesn't, I'll add parametrized cases before merging.

Full suite 3152 passed. Review rounds in progress.

BY571:
## Round 1 findings — NOT ready to merge

Three agents ran. Two found defects; both are recorded here in full because I ran out of working context before I could land the fixes. **Branch is at its green committed state** (796 offline / 3152 full suite passing); nothing half-applied.

### 1. The margin cap has ZERO test killers (test-justifier, mutation-verified in an isolated worktree)

Deleting `max(loss, -margin_to_return)` (scalar) or `torch.maximum(pnl, -margin_return)` (vectorized) — separately or together — leaves **all 796 offline tests green**.

Cause: every gapped-liquidation test uses `position_fraction=1.0`, where `_clamp_balance()` floors at 0 regardless of how far past `-margin` the uncapped loss overshoots. The two code paths are numerically indistinguishable *only* at 100% sizing. With `position_fraction=0.5`:

| | balance |
|---|---|
| with cap (current code) | 5000.0 |
| cap deleted (mutant) | 2500.0 |

So the headline half of this PR is unpinned. Also: `test_liquidation_returns_margin` (`fundamental/test_margin_accounting.py:387`) asserts **nothing** — a dead-letter test whose docstring describes the behaviour it never checks.

### 2. The cap is fee-blind (code-simplifier, reproduced)

With a nonzero `transaction_fee`, a full-margin gap books `-margin - fee + margin = -fee`:

```
WARNING: Negative balance detected: -84.158416. This likely indicates an accounting bug.
```

Hidden only because the test fixture sets `transaction_fee=0.0`. The isolated-margin claim in my docstring is therefore not true as written — the fee escapes the cap.

**I attempted the fix** (`max(loss - fee, -margin) + margin`, both engines) and it broke `test_liquidation_equivalence[long,short]`: scalar reward floors at `-10.0` while vectorized reaches `-27.7`. That is a **pre-existing asymmetry between `_clamp_balance()` and `clamp_(min=0.0)`** that folding the fee into the cap exposes in non-gapped liquidations too. It needs real investigation, not a quick patch — reverted rather than shipped half-done.

### 3. Coverage gaps beyond the cap

- **Short direction** has no absolute pin for gapped liquidation on either engine — only a *relative* equivalence check, which by construction cannot catch a regression applied to both engines in lockstep. This repo has a documented history of short-side sign bugs.
- **Non-SLTP engines** (`SequentialTradingEnv`, `VectorizedSequentialTradingEnv`) have **no gapped-liquidation coverage of any kind** — their equivalence test explicitly avoids gapped data. They're protected only by sharing a code path with the SLTP variants.
- Mutation (a) — reverting the fill rule on *both* engines — is caught by exactly **one** test (`test_a_bar_that_gapped_past_liquidation_still_liquidates`, long-only), because it is the only *absolute* pin. Every equivalence test goes green on a joint regression.

### Remaining work before merge

1. Fold the fee inside the cap **and** resolve the scalar/vectorized clamp asymmetry it exposes.
2. Add the parametrized matrix `{long, short} x {position_fraction 1.0, 0.5}` on **both** engines — absolute balances, not equivalence, since equivalence cannot see joint regressions.
3. Share the duplicated tensor fill rule between `vectorized_sequential.py` and its SLTP subclass (same shape as the three-copy SLTP action-map bug CLAUDE.md cites as a scar).
4. Trim ~10 comment lines.
5. Rounds 2 and 3.

@CharlieHelps the modelling question in my earlier comment still stands, and finding 2 sharpens it: if the loss cap is against posted margin, does the liquidation fee come out of the margin or on top of it? That is the crux of the equivalence break.

BY571:
## Round 1, third agent — CRITICAL, and it falsifies this PR's stated invariant

### F1 — the cap makes a deeper crash CHEAPER, and at realistic fees this PR is MORE OPTIMISTIC than main

The loss is capped at the bankruptcy price; the fee is still charged at `fill_price`. Once the cap binds, `fill_price` is no longer the price anything was booked at, so for a long it falls without bound while PnL is frozen — **the fee shrinks as the crash deepens.**

10x long, entry 100, `position_fraction=0.30`, `transaction_fee=0.01`:

| bar open | main (book at liq) | this PR | delta |
|---|---|---|---|
| 90.0 | 6862.55 | 6754.55 | −108.00 |
| 70.0 | 6862.55 | 6809.09 | −53.45 |
| 50.0 | 6862.55 | 6863.64 | **+1.09 MORE OPTIMISTIC** |
| 30.0 | 6862.55 | 6918.18 | **+55.64 MORE OPTIMISTIC** |
| 10.0 | 6862.55 | 6972.73 | **+110.18 MORE OPTIMISTIC** |

Crossover at `f*(liq - fill) > mmr*E`. `fee=0.01` is the value in this file's own `sltp_config_spot`/`sltp_config_futures` fixtures. Non-monotonic at every leverage tested (2/5/10/25/125). **I claimed this change could only ever be more pessimistic — that was false.**

**Shorts are the mirror image**: `fill_price` rises without bound so the fee does too — 10x short, entry 100, margin 1000, fee 0.01, bar opening at 400 gives a total debit of **1400 on a 1000 margin**. The docstring's "cannot lose more than it posted" is false in that direction.

**Fix:** clamp the *fill* to the bankruptcy price so both legs price at the same instant — which makes `max(loss, -margin)` redundant:

```python
bankruptcy = entry * (1 - 1/L) if is_long else entry * (1 + 1/L)
fill_price = stop_fill_price(liq, open_price, is_long)
fill_price = max(fill_price, bankruptcy) if is_long else min(fill_price, bankruptcy)
```

Verified over the full leverage × fee × direction × gap grid: **0 non-monotonic pairs and 0 optimistic-vs-main cases**, down from 2475 and 431.

Open design call: with the fill clamped the fee still sits on top of a full-margin loss. Capping the *total* fixes that but is then more optimistic than main at high fees — because **main already debits more than the margin** when `f*liq > mmr*E`. The two claims are incompatible at fee=0.01. Given this repo's pessimism-first philosophy the recommendation is to keep the fee on top and soften the docstring to "the *PnL* cannot exceed posted margin; the liquidation fee is charged on top, as on every other exit." **This is the same question I asked @CharlieHelps above** — it is now the crux.

### F4 — ReplayOrderExecutor was not updated, and its justifying comment is now false

`torchtrade/envs/replay/order_executor.py:133-136` says it books at the liquidation price "as the offline env does". It no longer does. Same position, same bar opening at 85: **replay 50036.384 vs offline 49996.600 — 39.78 divergence on a single liquidation.** Replay is the backtest→live bridge and #278 is literally about this divergence. `_close_at_price` already imports `stop_fill_price` for stops two branches down; liquidation is the branch it was not applied to. CLAUDE.md's "apply to ALL environments" rule puts this in the same PR — #314's own scope list names this file.

### F3 — confirms the cap is unpinned (independent second confirmation)

Deleting the cap on either engine: **796 passed**. The moved assertion passes with the cap deleted, because `_clamp_balance()` maps both −5000 and 0 to 0.0 at `position_fraction=1.0`.

### What this agent verified as SOLID

- **Sign correctness both directions, both engines** — no repeat of the historical short bugs
- **Scalar/vectorized parity is exact**: 432 end-to-end comparisons across leverage 2–125, both directions, fee 0–0.01, gaps 0–80% past liquidation — worst |diff| **0.0**. The specific risk #314 flagged did not materialise.
- **`_base_tensor[:, 0]` really is open**, and robust to input column order (`.agg(dict)` orders by dict key)
- `open` threaded from the correct bar at all four call sites
- `margin_to_return` genuinely equals posted margin, including after averaging in

### Revised remaining work

1. **Clamp the fill to the bankruptcy price** (replaces the cap) — fixes F1 on both engines
2. **Update `ReplayOrderExecutor`** — F4
3. Resolve the fee-on-top vs fee-inside question, and make the docstring match whichever is chosen
4. Add the parametrized `{long,short} × {position_fraction 1.0, 0.5}` matrix with **free cash remaining**, plus a **monotonicity property test** (a deeper gap must never leave more money) — that would have caught F1 directly
5. Rounds 2 and 3

Not merging this without the above.

BY571:
## Round 2 — two findings, one of them new and previously invisible

Round 1's three findings (F1 non-monotonic pricing, F3 unpinned clamp, F4 replay divergence) are **fixed, pushed and mutation-verified** — see the last three commits. Round 2 then found two more.

### R2-1 — the vectorized bankruptcy clamp is completely unpinned (confirmed by mutation)

Deleting the clamp from `vectorized_sequential.py::_apply_liquidation` **alone** leaves the whole suite green — 2995 passed, 0 failed. My two new tests instantiate `SequentialTradingEnvSLTP` only; they never touch the vectorized engine.

Cause: `_make_sltp_pair` defaults to `position_fraction=1.0`, so the equivalence suite's gap rows (`long-gapped`, `short-gapped`) inherit **the exact blind spot round 1 diagnosed on the scalar side** — the balance floors at 0 whether the fill was clamped or ran far past the margin. The round-1 fix (leave free cash) was applied to the scalar tests and not to the equivalence suite.

### R2-2 — a wrong-but-consistent bankruptcy constant survives everything

Mutating `entry * (1 - 1/leverage)` to `entry * (1 - 1.5/leverage)` — wrong value, still gap-depth-consistent — passes **all 841 offline+replay tests and both of my new tests.** My identity test only asserts self-consistency across deep gaps; it never checks the constant against anything independent. It needs one absolute number derived from the account rather than from the formula under test.

### NEW — the fix for those two exposed a real scalar/vectorized divergence

I wrote the combined test (free cash + absolute value + both engines). The absolute number is confirmed correct — isolated margin means a liquidated position costs exactly its posted margin, so `0.30 x 1000 = 300` lost, **700 remaining**:

```
[gapped-liquidation-margin-long] Step 10 balance: scalar=700.000000 vec=699.999928 diff=0.000072
[gapped-liquidation-margin-long] Step 10 portfolio_value: scalar=700.000000 vec=699.999928 diff=0.000072
```

**Scalar is exactly right; the vectorized engine is 7.2e-5 off at the liquidation boundary.** Every previous gapped test ran all-in, where both engines floor to 0.0 — so this divergence has been invisible, not absent. It is ~1e-7 relative, i.e. float32-epsilon territory, which points at a dtype in the vectorized liquidation path (`_entry_prices`?) rather than at the formula, given `MONEY_DTYPE = torch.float64`.

I have **not** committed that test — it fails, and I would rather leave the branch green with the finding recorded than land a red test or paper over it by loosening the tolerance. Branch is at its pushed state, 841 offline+replay passing, 3156 full suite.

### Remaining before merge

1. Diagnose the 7.2e-5 vectorized divergence (dtype in the liquidation path), then land the combined test — it closes R2-1 and R2-2 together and pins both engines with one assertion
2. Round 3
3. Follow-up, not this PR: `test_liquidation_returns_margin` (`fundamental/test_margin_accounting.py:387`) computes `balance_after_open` and `locked_margin`, uses neither, and ends with a comment where an assertion should be — it asserts nothing about margin

@CharlieHelps the fee question from my earlier comment is still open and still the one modelling call I'd most like a second opinion on.

BY571:
## Round 2 closed — all six findings fixed. Round 3 running.

@CharlieHelps a full review of the current state would be valuable. **The fee question I asked you earlier is now answered by evidence rather than preference**, and the answer was the opposite of what I had shipped:

Charging the liquidation fee *on top* of a full-margin loss produces **negative equity** in replay (−466.67 at 50x, where main gave +1436.19). That is not a modelling trade-off, it is an impossible account state. Real venues take the fee *out of* the margin — that is what the maintenance buffer is for — so the bankruptcy price is the fill where `loss + fee == margin` exactly: `entry*(1 ∓ 1/L)/(1 ∓ f)`.

### Every finding from rounds 1–2, all fixed and mutation-verified

| | Finding | Evidence it is fixed |
|---|---|---|
| F1 | A deeper crash was **cheaper**; at 1% fee this PR was **more optimistic than main** | monotonicity property test; 0 non-monotonic pairs, 0 optimistic rows over 3920 grid rows |
| F3 | The clamp had **zero killers** across all 796 offline tests | free-cash test; removing the clamp now kills it |
| F4 | Replay diverged **39.78** from offline on one liquidation | same fill rule; every fee=0 and pf<1 row now agrees to <1e-6 |
| A | Vectorized bankruptcy price computed in **float32** — parity went 1.82e-12 → **9.54e-3**, past the repo's own 1e-9 tolerance, sign varying by leverage | branches rebuilt as float64 tensor expressions; equivalence row added |
| B | Replay ended with **negative equity** | verified non-negative on every measured config |
| C | Offline logged **"This likely indicates an accounting bug"** on ordinary input, then floored to 0 | warning gone; balance lands on exactly 0 |

### Two things worth a human eye

1. **The bankruptcy expression now appears in three places** (scalar, vectorized, replay). That is the shape CLAUDE.md names as the scar behind the triple SLTP action-map bug. Round 3's simplicity reviewer is judging whether it should move into `envs/utils/liquidation.py` — I'd value your view too.
2. **Both regression tests needed specific sizing to work at all.** The free-cash test needs `position_fraction<1` (all-in floors the balance to 0 and hides the clamp); the replay test needs **all-in** (leftover cash absorbs the overdraft — my 0.97-sized first draft let the mutant survive). Same bug class, opposite sizing. Worth checking I have not left a third variant of it somewhere.

Full suite 3161 passed. Rounds 1–2 complete, round 3 in flight.

BY571:
## Round 3 — DO NOT MERGE. One blocker, and it is the mirror image of the bug this PR fixes.

Two of round 3's findings are fixed and pushed (shared `bankruptcy_price` helper; fee-net divisor now pinned on **both offline engines** — it had been protected only by a replay test plus code-sharing, and my previous commit message wrongly claimed otherwise. Correction recorded in `c79a4c7`).

### BLOCKER — at `transaction_fee > mmr` the clamp becomes a RAISE

`liq_long = entry*(1 − 1/L + mmr)` but `bankruptcy_long = entry*(1 − 1/L)/(1 − f)`. Bankruptcy stays below liq only while `f ≤ mmr/((1 − 1/L) + mmr)` — with the default `DEFAULT_MAINTENANCE_MARGIN_RATE = 0.004` that threshold is **≈0.004 at every leverage**. Above it, `max(fill, bankruptcy)` stops being a floor.

Measured by execution:

- **347/800** `(L, mmr, f, side)` cells put bankruptcy on the wrong side of liq
- **41/98** legal `(leverage, fee, direction)` combos book a fill **outside the bar's traded range**. At `L=125, fee=0.01`: bar `open=high=99.50`, fill = **100.20** — *above the entry price*, i.e. a liquidated 125x long booked at a profit on price
- **24/150 non-gapped** liquidation cells now end richer than main, by up to **+1437.33 on a 10,000 account**. All 24 at `fee ∈ {0.004, 0.01}`; zero at `fee ≤ 0.001`

**#314 is "don't book at a price the bar never offered." Above the fee threshold this books above the bar's high.** Same error, mirrored.

**And my own test cannot see it**: `test_a_deeper_gap_never_leaves_the_account_richer` runs at `fee=0.01, L=10` — inside the inverted regime, threshold 0.0044. It passes because past bankruptcy every gap costs exactly the margin, which the assertion checks and the inversion preserves.

### Also unvalidated

- `SequentialTradingEnvConfig` **does not validate `transaction_fee` at all** — accepts `−0.5`, `1.0`, `2.0`. At `f=2.0` the long bankruptcy price is **−90.0**, silently disabling the clamp
- `VectorizedSequentialTradingEnvConfig` allows `0 <= f <= 1`, so **`f=1.0` is legal and raises `ZeroDivisionError`** at the liquidation site
- `ReplayOrderExecutor` validates `maintenance_margin_rate` but not `transaction_fee`

### The fix, per CLAUDE.md invariant 4 ("validate at the boundary, never guard in the rule")

Raise in `__post_init__` (and `ReplayOrderExecutor.__init__`) when the fee is large enough to push bankruptcy past liquidation — `f * (1 + 1/L) >= mmr` as the conservative both-sides bound. **A fee exceeding the maintenance buffer is an internally inconsistent venue model**; absorbing it in the fill clamp is exactly the "guard that makes the nonsense silent" the invariant warns against. Then correct the `binds ONLY on a gap` sentence at `sequential.py:924`, which is now false for `f > ~0.004`, and re-point my monotonicity test at a legal fee.

Note this also removes a signal: on main these configs tripped `_clamp_balance()`'s "accounting bug" warning. The branch emits **zero** warnings across a 150-cell grid — correct for the gapped path, but it means an inconsistent fee/mmr pair now passes silently.

### What round 3 confirmed as SOLID

| Check | Result |
|---|---|
| Monotonicity | 150 cells × 7 gap depths — **0 violations >1e-9** |
| Optimistic vs main | 0, for `fee ≤ 0.001` |
| Scalar/vec parity | worst **2.27e-13** (was 9.54e-3 with the float32 bug) |
| Replay vs offline | 510 cells, **0 differing**, worst 4.26e-14 — the round-2 divergence is gone |

Full suite 3163 passed. Rounds 1–3 complete: **9 findings, 8 fixed**, 1 blocker open.

@CharlieHelps this blocker is a genuine design question and I would value your read: should an inconsistent fee/mmr pair be rejected at config time (my inclination, and what the invariant says), or should the model instead derive the maintenance rate from the fee so the two cannot disagree?

BY571:
## Round 3 blocker — attempted the fix, reverted, and here is why it is not mechanical

I implemented the recommended boundary validation (`f * (1 + 1/L) >= mmr` → raise in `__post_init__`, plus the missing `0 <= f < 1` range check) and **reverted it**. Branch stays at its green pushed state — 848 offline+replay, 3163 full suite.

**It breaks 12 existing tests**, across five files:

```
fundamental/test_margin_accounting.py::test_liquidation_price_calculation_long
fundamental/test_margin_accounting.py::test_liquidation_returns_margin
test_sequential.py::test_invalid_transaction_fee_raises
test_sequential_sltp.py::test_a_deeper_gap_never_leaves_the_account_richer
test_vec_scalar_equivalence.py::test_liquidation_equivalence
```

Two things that means, neither of which I should decide alone at this point:

1. **The repo's own fixtures routinely use fees above the maintenance buffer.** With the default `mmr = 0.004`, the bound rejects roughly `fee >= 0.0036`. That these fixtures exist and pass today says either the default maintenance rate is unrealistically small for the fees this codebase tests with, or the bound should be per-direction (long-only: `f >= mmr/((1 − 1/L) + mmr)`) rather than the conservative both-sides form.

2. **`transaction_fee` IS already validated somewhere** — `test_invalid_transaction_fee_raises` expects `ValueError` matching `"Transaction fee"`. So my "validated nowhere" note above was wrong about the range check; round 3's finding that `f=1.0` and `f=2.0` are accepted still needs reconciling against whatever that existing check permits. I did not have the context left to trace it properly, and I would rather say so than paper over it.

**The blocker itself stands exactly as reported**: at `f > ~mmr` the fill clamp becomes a raise, booking liquidations above the bar's high and above entry at 125x, on **347/800** measured cells. Nothing about this attempt weakens that evidence.

### Where #364 actually stands

Rounds 1–3 complete. **9 findings, 8 fixed and mutation-verified, 1 blocker open.** Everything pushed and green.

Confirmed solid by round 3: monotonicity 0 violations over 150 cells × 7 depths; scalar/vec parity **2.27e-13** (from 9.54e-3); replay/offline **0 differing** over 510 cells.

@CharlieHelps the design question is now sharper and is the thing I would most value your call on: **should the fee/mmr consistency bound reject configs the repo currently ships and tests with, or does the default `maintenance_margin_rate = 0.004` need to rise to match realistic taker fees?** Those are the two ways out and they have different blast radii — the first breaks existing user configs, the second changes every liquidation price in the codebase.

BY571:
**Correction, traced:** `transaction_fee` is validated at `torchtrade/envs/core/base.py:87` and `vectorized_sequential.py:88` — "Transaction fee must be between 0 and 1" — at **env construction**, not on the config dataclass. So round 3's probe finding `-0.5`/`1.0`/`2.0` accepted was exercising the config object in isolation, which no user path does.

That narrows the unvalidated surface to two real items rather than three:

1. **`f = 1.0` is legal** under a `0 <= f <= 1` inclusive bound and divides by zero in the bankruptcy price. A one-character fix (`<` for `<=`), and the same off-by-one I fixed for `slippage` in #361.
2. **`ReplayOrderExecutor` validates `maintenance_margin_rate` but not `transaction_fee` at all** — it has no equivalent boundary, so it is the one path where the config object's looseness is reachable.

The **fee-vs-mmr consistency bound remains the open design question**, unchanged by this: it is a relationship between two independently-valid values, which no single-field range check can express.

Sorry for the noise — I should have grepped for the existing validator before asserting it was absent.

BY571:
@CharlieHelps review

BY571:
@CharlieHelps your blocker is fixed at `d2e4a77` — ready for re-review.

**Your counter-example is now refused at construction.** Verified by execution:

| case | result |
|---|---|
| `L=125, mmr=0.004, fee=0.01` (yours) | **REFUSED** — "transaction_fee 0.01 does not fit inside maintenance_margin_rate 0.004" |
| `fee=1` (your divide-by-zero) | **REFUSED** — `transaction_fee` is now `[0, 1)`, was `[0, 1]` |
| `L=125, fee=0.0004` (realistic) | constructs |
| `L=1, fee=0.01` (spot) | constructs — no liquidation to misprice |

I took your first option — **reject at the public boundary** — rather than revising the pricing model, because CLAUDE.md's invariant 4 says the same thing ("validate at the boundary, never guard in the rule") and because absorbing it in the clamp is what made the inconsistency silent in the first place.

Two things worth flagging from doing it:

1. **I got the comparison wrong first.** `>=` refused `fee=0, mmr=0`, where the bankruptcy price *equals* the liquidation price and the clamp is a no-op rather than a violation. Strict `>` now.

2. **The 12 fixtures this broke were the evidence, not the cost.** They used `fee=0.01` — a 1% futures taker fee, which no venue charges. Real venues run the maintenance buffer an order of magnitude *above* the fee (Binance futures ~0.04% taker against ~0.4% maintenance), and that gap is exactly why the bankruptcy price normally lands inside the liquidation price. Moved to 0.0004; the one test whose subject genuinely *is* fee impact at high leverage moved to 0.003 — high, but coherent.

So the guard didn't invalidate 12 valid configs, it surfaced 12 encoding an impossible venue. I'd have missed that framing without your review forcing the question.

Full suite 3163. Rounds 1–3 complete: 10 findings, all fixed.

BY571:
@CharlieHelps both blocking findings fixed at `dd62fba`.

You were right that I fixed the instance rather than the class. The rule now lives **once** in `torchtrade/envs/utils/liquidation.py`, beside the formula it protects, and all three constructors call it:

| boundary | before | now |
|---|---|---|
| scalar env (`core/base.py`) | validated | validated via shared rule |
| `VectorizedSequentialTradingEnvConfig` | **unvalidated** | validated |
| `ReplayOrderExecutor` | **unvalidated** | validated |

Putting it in any single config is what allowed the gap — there is no one config-level place the other two paths must pass through, which is why it moved next to `bankruptcy_price` itself.

**Regression coverage** is the 3×2 matrix you asked for: each boundary against both your cases (`L=125, mmr=0.004, fee=0.01` and `fee=1`). Mutation-verified per boundary — removing the check from the vectorized config fails exactly its 2 cells, and from the replay executor exactly its 2. Neither had any coverage before, which is precisely why the first fix looked complete when it wasn't.

Full suite 3169.

One note on the shared rule: the comparison is strict (`>`), because at `fee=0` with `mmr=0` the bankruptcy price *equals* the liquidation price and the clamp is a no-op rather than a violation. My first attempt used `>=` and refused that legitimate config.

BY571:
@CharlieHelps fixed at `43cfdd8` — and thank you, the third review caught a real compatibility regression I would have shipped.

I re-derived your condition rather than take it on faith, and it checks out:
- long: `(1 − 1/L)/(1 − f) ≤ (1 − 1/L + mmr)` ⟺ `f·(1 − 1/L + mmr) ≤ mmr`
- short: `(1 + 1/L)/(1 + f) ≥ (1 + 1/L − mmr)` ⟺ `f·(1 + 1/L − mmr) ≤ mmr`

**Your example needed one more step, which I think sharpens the fix.** `L=2, mmr=0.004, fee=0.004` is a floor for longs (bankruptcy 50.2008 vs liquidation 50.4) — but it genuinely *inverts for shorts* (149.402 vs 149.6). The short constraint binds whenever `1/L > mmr`, so it is not that the check was wrong to consider shorts; it was wrong to consider them for a config that cannot open one.

So `allows_short` is now derived from `action_levels`, and both constraints are applied only for the directions the config can actually take:

| config | verdict |
|---|---|
| `L=2, mmr=.004, fee=.004`, long-only | **accepted** (was refused) |
| same, with shorts enabled | refused — the short side really does invert |
| `L=125, mmr=.004, fee=.01` (your original) | refused |
| exact equality | accepted — bankruptcy lands *on* liquidation, so the clamp is a no-op |

Equality is accepted deliberately, per your "including equality".

**Acceptance coverage added, not just rejection** — near-threshold configs that must be allowed, including exact equality and a hair past it. Mutation-verified: restoring the conservative bound fails 2 cells.

Full suite 3173.

BY571:
@CharlieHelps fixed at `dcb4b30`. You were right and it was the exact mirror of the bug I had just fixed — I gated the short constraint and left the long one unconditional, so a short-only config was measured against a long it can never open.

Both are gated on `action_levels` now.

**The test needed a specific regime to be meaningful, and my first attempt did not have it.** In every case I had, the *short* term is the binding one — it is, whenever `1/L > mmr` — so ungating the long constraint changed nothing and the mutant survived all 34 tests. The long term only binds when `mmr > 1/L`:

| L=2, mmr=0.6, fee=0.6 | |
|---|---|
| short side | 0.540 ≤ 0.6 → passes |
| long side | 0.660 > 0.6 → fails |

So a short-only config must be **accepted** there, and is refused the instant the long constraint applies unconditionally. That case now kills the mutant; without it the fix would have looked verified and been unpinned.

Also folded the repeated `action_levels` lookup into a helper, defaulting to **both** directions when a config has none — a missing attribute should widen the check, not skip it.

Full suite 3176.

---

On **#368**: your review flags the regex omitting kwargs and `KNOWN_BROKEN` not being checked for staleness. Both were fixed in the two commits after the one you reviewed — reach went 289→307 triples (the mkdocs-tab `SequentialTradingEnvConfig` examples, all Polymarket, and the `live/README.md` `api_key` case are visible now), skips 12→1, and `test_no_known_broken_entry_has_gone_stale` asserts every entry still appears in the docs. Worth a re-look at `7848b9f` if you have a moment.

BY571:
@CharlieHelps your 07:54 note said the current head was still `43cfdd8` — the direction-gating fix landed after that, at `dcb4b30`. So the symmetric gap you flagged is already closed and you have not seen it.

**"The validator still always applies the long-side bound, so a safe short-only config can be rejected"** — both bounds are gated on `action_levels` now.

The test needed a specific regime to mean anything, and my first attempt did not have it: everywhere my existing cases live, the **short** term is the binding one (it is, whenever `1/L > mmr`), so ungating the long constraint changed nothing and the mutant survived all 34 tests. The long term only binds when `mmr > 1/L`:

| L=2, mmr=0.6, fee=0.6 | |
|---|---|
| short side | 0.540 ≤ 0.6 → passes |
| long side | 0.660 > 0.6 → fails |

A short-only config must therefore be **accepted** there, and is refused the instant the long bound applies unconditionally. That case kills the mutant; without it the fix would have looked verified and been unpinned.

Full suite 3416. Ready for your fifth look — and thank you for the previous four, every one of them found something my own three review rounds had passed.

PR #366, opened 2026-08-12
fix(okx): send posSide where OKX requires it on set-leverage (#363)

### LINKED ISSUE #363: OKX: set-leverage omits posSide, which OKX requires under isolated + LONG_SHORT
Found while reviewing #362 (leverage verification), verified by the silent-failure-hunter against OKX's API contract.

`torchtrade/envs/live/okx/order_executor.py` `_setup_futures_account` calls:

```python
self.account_client.set_leverage(instId=..., lever=..., mgnMode=self.margin_mode.value)
```

OKX **requires** `posSide` on set-leverage when margin mode is isolated and position mode is `LONG_SHORT`. Without it OKX answers `code="1"` with `sMsg="Parameter posSide error"`.

**Before #362** that logged a warning and construction continued — the env then sized every position against a leverage the account never had, i.e. #277 exactly, on that config.

**After #362** it raises at construction with the real `sMsg` surfaced. So this is no longer a silent money bug; it is now a hard failure on a configuration that cannot currently work.

The default `PositionMode.NET` is unaffected, which is why no test caught it.

**Fix:** pass `posSide` when `margin_mode is MarginMode.ISOLATED and position_mode is PositionMode.LONG_SHORT` — OKX needs one call per side there, so this is two calls, and both should be verified by `require_leverage_applied` the way the single call is today.

Note `torchtrade/envs/utils/leverage.py` already checks every entry in the response's `data` list, so the verification side needs no change.

BY571:
Closes #363. Found while reviewing #362, filed rather than scope-creeping that PR.

## The bug

OKX **requires** `posSide` on set-leverage under isolated margin + `LONG_SHORT` position mode, and rejects the call without it (`code 1`, `sMsg "Parameter posSide error"`). Isolated leverage is stored per side there, so it takes one call each.

- **Before #362**: the rejection was swallowed with a warning and construction continued — the env then sized every position against leverage the account never had. That is #277 exactly, on this one configuration.
- **After #362**: it became a hard construction failure. Safe, but the config is simply unusable.

Now it works.

The verification added in #362 needs no change — it already loops every entry in the response's `data` list, so both sides get confirmed.

## Evidence

Tested in **both directions**, which is what makes this more than a shape check: sending `posSide` where OKX does not expect it is its own rejection, so cross and net must *not* get it.

| mutant | result |
|---|---|
| never send `posSide` (the shipped bug) | fails `isolated-long-short` |
| send it for `LONG_SHORT` regardless of margin mode | fails `cross-long-short` |

Full suite: 3163 passed.

⚠️ Review rounds not yet run.

### DIFF
--- a/tests/envs/okx/conftest.py
+++ b/tests/envs/okx/conftest.py
@@ -33,10 +33,12 @@ def mock_okx_account_client():
 
     # Mock account configuration
     client.set_position_mode = MagicMock(return_value={"code": "0", "msg": ""})
-    # `lever` present: an entry without it no longer confirms anything (#277).
+    # Echoes `lever` AND `posSide`, as OKX documents: an entry without lever confirms
+    # nothing (#277), and one echoing the wrong side confirmed the wrong leg (#363).
     client.set_leverage = MagicMock(
-        side_effect=lambda instId=None, lever=None, **k: {
-            "code": "0", "msg": "", "data": [{"lever": lever}]})
+        side_effect=lambda instId=None, lever=None, posSide=None, **k: {
+            "code": "0", "msg": "",
+            "data": [{"lever": lever, **({} if posSide is None else {"posSide": posSide})}]})
 
     # Mock position information
     client.get_positions = MagicMock(return_value={
--- a/tests/envs/test_leverage_boundary.py
+++ b/tests/envs/test_leverage_boundary.py
@@ -294,8 +294,11 @@ def test_okx_sets_leverage_per_side_only_where_okx_requires_it(
     seen = []
 
     def record(**kwargs):
-        seen.append(kwargs.get("posSide"))
-        return {"code": "0", "data": [{"lever": "20"}]}
+        pos_side = kwargs.get("posSide")
+        seen.append(pos_side)
+        # Echoes the side back, as OKX documents -- the executor checks it.
+        entry = {"lever": "20", **({} if pos_side is None else {"posSide": pos_side})}
+        return {"code": "0", "data": [entry]}
 
     OKXFuturesOrderClass(
         symbol="BTC-USDT-SWAP", trade_mode="quantity", demo=True, leverage=20,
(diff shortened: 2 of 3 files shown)


BY571:
@CharlieHelps please review. Small PR, but it sits on the leverage path #362 just hardened, so a second opinion matters.

**What it does:** OKX requires `posSide` on set-leverage under isolated margin + `LONG_SHORT`, and rejects the call without it (`code 1`, `sMsg "Parameter posSide error"`). Isolated leverage is stored per side there, so it takes one call each.

Before #362 that rejection was swallowed and the env sized against leverage the account never had. After #362 it became a hard construction failure — safe, but the config was unusable. This makes it work.

**Where I'd value your judgement:**

1. **Is the condition exactly right?** I gate on `ISOLATED and LONG_SHORT`. If OKX merely *ignores* `posSide` for cross or net rather than rejecting it, the narrow gate is still correct but for a weaker reason, and I'd rather know which.

2. **The half-configured hazard.** With two calls, the first can succeed and the second fail — leaving long at 20x and short at 5x while construction reports success. Sizing then uses one number for both sides. I believe the per-call verification catches it (each call raises on mismatch, so a failed second call aborts construction), but I have asked round 1 to confirm by execution rather than take my reading of it.

3. **Is asserting the exact `posSide` sequence over-specification?** The test pins `["long", "short"]` in order. Order shouldn't matter to OKX, so this may break on a harmless refactor.

**Evidence so far:** mutation-verified in both directions — never sending `posSide` fails the isolated cell; sending it for cross fails another. That two-sidedness is deliberate: sending `posSide` where OKX doesn't expect it is its own rejection, so the guard has to be pinned on both edges. Full suite 3163.

---

Also still open and worth your time when you have it: **#364's blocker** (rounds 1–3 done, 9 findings, 8 fixed). At `fee > ~mmr` the liquidation fill clamp inverts and books above the bar's high. The two exits are reject the config (12 existing tests fail) or raise the default `maintenance_margin_rate` (changes every liquidation price) — different blast radii, both on money, and genuinely a judgement call rather than a bug with a right answer.

PR #358, opened 2026-08-12
chore: three sweeps — dead code, crashing examples, unrunnable docs (#285, #286, #287)

BY571:
Closes #285. Closes #286. Closes #287 (partially — see "Not done" below).

**Net −234 lines.** Mostly deletion, and everything machine-verified rather than reviewed by eye.

## #286 — fourteen example branches crashed on NameError

`ruff F821` confirms all fourteen. dsac and iql reference the SLTP and OneStep envs they never import; ppo_chronos imports *only* the SLTP env while its own shipped `env/sequential.yaml` selects `SequentialTradingEnv`. So `env=sequential_sltp` — **a config file we ship** — died immediately for anyone who selected it.

Also `ppo_vectorized` read `cfg.time_frames` where the config nests under `cfg.env` — a latent crash behind `model.network_type=cnn`.

`ruff check examples/ --select F821` now passes.

## #285 — deleted, each verified to have zero callers first

| deleted | why |
|---|---|
| binance `_handle_long_action` / `_handle_short_action` | uncalled, and they reference a config field that does not exist |
| `combinatory_action_map` alias | nothing imports it — the tests alias the real name themselves |
| `position_penalty` on 3 live configs | advertised in the config, never read anywhere |
| sequential `_reset_history` override | only restated the base |
| `tests/mocks/{base,binance,bitget,conftest}.py` | 244 lines nothing imports; the live fixtures are re-implemented in `tests/envs/*/conftest.py` |

## #287 — a half-finished rename left ~40 documented imports raising

Every `from torchtrade... import X` in `README.md`, `docs/` and the in-package READMEs now resolves. **The count was 39 broken; it is 0.**

Where a README documented an API that was never written at all — `FractionalPositionSizing`, `calculate_metrics`, `RewardFunction`, `SeqLongOnlyEnv` — the block is replaced by the real one or removed. An authoritative-looking example that cannot run costs the reader more than no example.

## The durable part

`tests/test_documented_imports.py` — 163 cases that **execute** every documented import.

Prose cannot be verified by review, only by running it, which is how #287 accumulated in the first place. Mutation-proven: restoring one pre-rename path turns it red.

## Not done here

#287 also lists config kwargs that raise `TypeError` and observation keys missing their `_{window_size}` suffix. Those live in prose rather than in import lines, so the guard cannot reach them — they need a human pass, and I would rather leave the issue open for that than close it claiming more than was done.

Full suite: **2996 passed, 19 skipped**.

[...]

Issue #287, opened 2026-08-07
Docs sweep: half-finished envs.live path rename, ~10 documented config params that raise TypeError, wrong observation keys, phantom symbols in package READMEs


BY571:
Every claim below verified against source (import attempted or grep):

**Old module paths (pre-envs/live layout)**: \`docs/environments/online.md:61,77,102,120,165,196,241,269\`, \`docs/getting-started.md:196,220\`, \`docs/environments/offline-rl.md:77\`, \`README.md:359\`; nonexistent \`torchtrade.envs.bitget.futures_order_executor\` at \`online.md:166,197\`. OKX/Polymarket sections on the same pages use correct paths — a half-finished rename.

**Config params that raise TypeError if used**: \`margin_call_threshold\` (\`docs/index.md:66\`, \`offline.md:204,310\`), \`rollout_steps\` (\`offline.md:306\`), \`reward_function\`/\`feature_preprocessing_fn\` documented as config fields when they are constructor args (\`guides/reward-functions.md:30-70\`, \`guides/custom-features.md:77-190\`, \`README.md:424\`), \`trading_mode="spot"\` (\`README.md:96,343,423,442,524\`), \`intervals=\`/\`quantity_per_trade=\` on non-SLTP live configs (\`online.md:106-110,179\`, \`getting-started.md:227-228\`).

**Wrong observation keys**: docs omit the \`_{window_size}\` suffix (real: \`market_data_1Minute_10\`) — \`offline.md:114-117\`, \`getting-started.md:150-153\`, \`guides/sampler.md:56-86\`, \`custom-features.md:24-25\`; \`README.md:449-460\` uses \`market_data_1min\` which the code can never emit.

**Phantom symbols in in-package READMEs**: \`torchtrade/envs/utils/README.md\` imports 7 nonexistent names (\`discrete_action_map_long_only\`, \`calculate_sltp_prices\`, \`check_sltp_hit\`, \`FractionalPositionSizing\`, \`calculate_metrics\`, \`SeqLongOnlyEnv(+Config)\`); \`core/README.md\`: \`RewardFunction\`, \`TorchTradeEnvConfig\`; \`live/README.md\`: \`WandbLogger\`.

[...]

PR #367, opened 2026-08-13
docs: remove config kwargs that raise TypeError if a reader copies them (#287)

BY571:
First slice of #287. Scoped to the class of error that actually breaks a reader: **parameters that do not exist.**

## Verified by grep, not by reading

`margin_call_threshold`, `rollout_steps` and `trading_mode` have **zero hits** anywhere in `torchtrade/`, and appeared 8 times across `README.md`, `docs/index.md` and `docs/environments/offline.md`. Copying any of them gives a `TypeError`.

| documented | real |
|---|---|
| `margin_call_threshold=0.2` | `maintenance_margin_rate=0.004` |
| `rollout_steps=24` | `max_traj_length=24` |
| `trading_mode="spot"` | `leverage=1` (incl. the hydra yaml key) |
| `market_data_1min` | `market_data_1Minute_12` — the suffix is the window size |

Also: `feature_preprocessing_fn` was shown as a **config field** when it is a **constructor argument**, in an example that then never used the function it had just defined. It now gets passed where it goes.

## Why #358's test didn't catch these

`tests/test_documented_imports.py` executes every documented import — 257 cases, all green — and #287's first bullet (old `envs.live` module paths) is already fixed by it and needed nothing here.

But its own docstring says it: *"Config kwargs and observation keys live in prose."* It validates that `from torchtrade.envs.offline import SequentialTradingEnvConfig` resolves, not that `SequentialTradingEnvConfig(trading_mode="spot")` constructs. Executable documentation stops at whatever boundary you drew around "executable" — here, the import line.

## Deliberately NOT in this slice

Phantom symbols in the in-package READMEs, the observation-key suffix across four more docs pages, and #287's misc list. Left for follow-up slices rather than one unreviewable docs diff.

Full suite: 3165 passed.

⚠️ Review rounds not yet run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,11 @@ df['timestamp'] = pd.to_datetime(df['timestamp'])
 
 # Create environment (spot trading = long-only)
 config = SequentialTradingEnvConfig(
-    leverage=1,  # 1 = spot; >1 for leveraged futures
+    leverage=1,             # 1 = spot; >1 for leveraged futures
+    action_levels=[0, 1],   # long-only; the default [-1, 0, 1] warns under leverage=1
     time_frames=["1min", "5min", "15min"],
     window_sizes=[12, 8, 8],
-    execute_on=(5, "Minute"),
+    execute_on="5Min",
     initial_cash=1000
 )
 env = SequentialTradingEnv(df, config)
@@ -340,10 +341,11 @@ df['0'] = pd.to_datetime(df['0'])
 
 # Configure multi-timeframe environment
 config = SequentialTradingEnvConfig(
-    leverage=1,  # spot; action_levels controls long-only vs bidirectional
+    leverage=1,             # spot
+    action_levels=[0, 1],   # long-only; the default [-1, 0, 1] warns under leverage=1
     time_frames=["1min", "5min", "15min", "60min"],
     window_sizes=[12, 8, 8, 24],
-    execute_on=(5, "Minute"),
+    execute_on="5Min",
     initial_cash=[1000, 5000],  # Domain randomization
     transaction_fee=0.0025,
     slippage=0.001
@@ -417,7 +419,9 @@ def custom_preprocessing(df):
         df["close"], window=14
     ).rsi()
     df.fillna(0, inplace=True)
-    return df
+    # timestamp must come back as a COLUMN, not an index, or the sampler raises
+    # KeyError: ['timestamp'] not in index
+    return df.reset_index()
 
 config = SequentialTradingEnvConfig(
     leverage=1,
@@ -444,14 +448,14 @@ config = SequentialTradingEnvConfig(
     leverage=1,
     time_frames=["1min", "5min", "15min", "60min"],
     window_sizes=[12, 8, 8, 24],
-    execute_on=(5, "Minute")
+    execute_on="5Min"
 )
 
 # Results in observations:
 # - market_data_1Minute_12: [12, num_features] - Last 12 one-minute bars
-# - market_data_5min: [8, num_features] - Last 40 minutes
-# - market_data_15min: [8, num_features] - Last 120 minutes
-# - market_data_60min: [24, num_features] - Last 24 hours
+# - market_data_5Minute_8: [8, num_features] - Last 40 minutes
+# - market_data_15Minute_8: [8, num_features] - Last 120 minutes
+# - market_data_60Minute_24: [24, num_features] - Last 24 hours
 ```
 
 ### Observation Structure
@@ -459,7 +463,7 @@ config = SequentialTradingEnvConfig(
 ```python
 observation = {
     "market_data_1Minute_12": tensor([12, num_features]),
-    "market_data_5min": tensor([8, num_features]),
+    "market_data_5Minute_8": tensor([8, num_features]),
     "account_state": tensor([6]),  # Universal 6-element state
 }
 
--- a/README.md
+++ b/README.md
@@ -429,6 +429,7 @@ def custom_preprocessing(df):
 
 config = SequentialTradingEnvConfig(
     leverage=1,
+    action_levels=[0, 1],   # long-only; the default [-1, 0, 1] warns under leverage=1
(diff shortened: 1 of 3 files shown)


BY571:
@CharlieHelps please review.

Small docs PR, but with a specific hazard worth a second pair of eyes: **a docs fix that substitutes one wrong value for another is worse than the original**, because it now looks authoritative and verified.

The three name replacements are grep-verified (the old names have zero hits in `torchtrade/`). What I am less sure of is whether the **values** preserve the examples' intent:

1. `margin_call_threshold=0.2` carried the comment *"20% margin ratio triggers liquidation"*. I replaced it with `maintenance_margin_rate=0.004`, which is the real field — but 0.004 is not 0.2, so the example no longer demonstrates what its comment said. Should the comment change, or the value?

2. `market_data_1Minute_12` — I derived the `_12` suffix from the surrounding example's `window_sizes`, by reading rather than by constructing an env and printing `market_data_keys`. Round 1 is checking that; if the format differs I have replaced one wrong key with another.

3. The hydra `trading_mode: "spot"` -> `leverage: 1` change assumes the yaml uses the same key name as the dataclass. I did not verify against the actual env yamls under `examples/`.

Deliberately scoped: this is one slice of #287 (the parameters that raise `TypeError`). Phantom symbols in the in-package READMEs, the observation-key suffix on four more pages, and the misc list are follow-ups — one unreviewable docs diff seemed worse than several small ones.

Worth noting for the wider issue: #358's `tests/test_documented_imports.py` passes on all of this (257 cases). Its docstring says why — it executes import blocks, and *"config kwargs and observation keys live in prose"*. The real fix for this class of rot is extending that test to construct the configs, not just import them. I have not done that here, and it may be the better PR.

---

Also still open: **#364's blocker**, which is the one genuine judgement call left. One piece of context I under-weighted when I first raised it — real venues run `mmr` at roughly **10x the taker fee** (Binance futures: taker ~0.04%, maintenance ~0.4%). So the repo's `mmr=0.004` default is realistic and the **test fixtures using `fee=0.01` are not**. That points toward rejecting the config and fixing the 12 fixtures, rather than loosening the rule to accommodate a venue that does not exist.

BY571:
Round 1 applied at `6ad3dd1`; round 2 running.

Round 1's verdict was **partial sweep** — the names were all correct, but the PR left the blocks it opened still broken. Three findings, all fixed:

1. **Half-fixed was worse than uniformly wrong.** I renamed `market_data_1min` and left `5min`/`15min`/`60min` beside it — in a block whose entire subject is the key format, so it read as though `1Minute` were special-cased. All four now carry the real `{freq}_{window_size}` form, confirmed by constructing the block's own config and printing `market_data_keys`.

2. **A reader copying the FIXED blocks still got a TypeError.** `execute_on=(5, "Minute")` appears 10 times across the three files — several inside the blocks I had just edited. `normalize_timeframe_config` converts `str`, not tuples.

3. **The example I "fixed" still crashed.** A `feature_preprocessing_fn` must return `timestamp` as a *column*; the README's returned it as an index. What convinced me this was real rather than theoretical: while verifying Charlie's counter-example on #364, my own throwaway script hit the identical `KeyError: ['timestamp'] not in index` — I wrote the same bug the docs teach, independently, within minutes. Fixed, and then I **ran the whole example end to end** rather than claiming it fixed.

Plus `action_levels=[0, 1]` where the text says spot is long-only: `leverage=1` with the default `[-1, 0, 1]` emits a UserWarning telling the reader their own config is wrong.

Round 1 also confirmed the judgement call I was unsure of: `margin_call_threshold=0.2` → `maintenance_margin_rate=0.004` was right to change the *value*, not just the name. Keeping `0.2` under the new name gives `isolated_liquidation_price(entry=100, long, lev=10, mmr=0.2) = 110.0` — a long liquidated **above** its entry, i.e. instantly. That was exactly the "one wrong value for another" trap I flagged to @CharlieHelps, and it would have been invisible.

Named for the next slice, not this one: `docs/guides/custom-features.md` still teaches `feature_preprocessing_fn` as a config field on 5 lines — and it is the page this README block links to for "more examples", so a reader lands on the broken version. Plus the phantom symbols in the in-package READMEs.

BY571:
Round 2 applied at `9f7662c`; round 3 running.

**Round 2 executed all 22 python blocks rather than reading them, and found three blockers — including that the very first block a reader runs did not work.**

`README`'s Quick Start called `env.step(tensordict)` on a tensordict from `reset()`, which carries no action:
```
KeyError: 'key "action" not found in TensorDict with keys
          ["account_state", "done", "market_data_15Minute_8", ...]'
```
Fixing that exposed the next line — `step()` writes under `"next"`, so `tensordict['reward']` is a `KeyError` too. **The previous commit had edited lines 96-100 of that same block and left 105-108 broken.**

Also: the HuggingFace block used `df['0']` against a dataset whose real columns are `timestamp/open/high/low/close/volume`, with `pd` never imported. And `offline.md`'s Observation Space listing still carried the unsuffixed keys **that round 1 had just fixed in README** — the config directly above it emits `market_data_1Minute_12`.

That last one is the third time on this PR that fixing one instance left its twin standing: one of four observation keys; README's keys but not `offline.md`'s; this block's config but not its usage. The common cause is that I verify at the granularity I edit at — executing the whole block is the only check operating at the granularity a *reader* uses.

The Quick Start block now runs verbatim: constructs, steps, prints a reward.

---

**Worth noting for #287 as a whole**: `tests/test_documented_imports.py` has passed all 257 cases through every one of these rounds. It validates that imports resolve, not that examples run. Extending it to *execute* the blocks would have caught every defect found in rounds 1-3 mechanically, and would keep catching them. I still think that is the more valuable PR than this one, and I would rather it be built deliberately than bolted on here.

@CharlieHelps this is ready for your eye whenever convenient — the specific thing I would value is whether the remaining slices (`docs/guides/custom-features.md` teaching `feature_preprocessing_fn` as a config field, phantom symbols in the in-package READMEs) are worth doing by hand at all, or whether the executable-docs test should land first and drive them.

PR #368, opened 2026-08-13
test(docs): check documented config kwargs actually exist (#287)

BY571:
Second slice of #287, and the one I think matters more than the hand-sweeping.

## Why the existing test missed everything

`tests/test_documented_imports.py` passed **all 257 cases** through three review rounds on #367 while `margin_call_threshold`, `rollout_steps` and `trading_mode` sat in the docs with **zero occurrences anywhere in the package**.

Its own docstring said why: it executes *import* blocks, and "config kwargs and observation keys live in prose". It validates that a **name resolves**, not that a **call is constructible**. The boundary was the import line, and every defect lived one level below it.

## What this adds

Extracts every `*Config(...)` kwarg from tracked markdown and checks it against the dataclass fields. That finds **58 more** — mechanically deriving what has until now been a hand-written list in the issue body.

## Shipped as a ratchet, not 58 doc edits

Every entry in `KNOWN_BROKEN` raises `TypeError` today. Shrink the list by fixing the doc; never grow it.

**Verified in both directions**, which is what makes it a ratchet rather than a suppression list:

| | result |
|---|---|
| a **new** bad kwarg (not on the list) | fails immediately |
| a listed entry that gets **fixed** | fails until removed from the list |

So the baseline cannot drift away from the docs in either direction, and the class of rot that produced #287 cannot come back silently.

`test_the_kwarg_sweep_found_configs_to_check` guards the regex itself — one that silently matched nothing would make the whole check vacuous, which is a failure mode this repo has hit before.

## Result

477 passed, 58 xfailed. Full suite green.

The 58 are the remaining #287 scope, now enumerated by CI rather than by hand: `intervals`/`quantity_per_trade` on non-SLTP live configs, `feature_preprocessing_fn`/`reward_function` as config fields, and the in-package README phantoms.

⚠️ Review rounds not yet run.

### DIFF
--- a/tests/test_documented_imports.py
+++ b/tests/test_documented_imports.py
@@ -100,6 +100,12 @@ KWARG = re.compile(r"(?:^|[\s,(])([A-Za-z_]\w*)\s*=(?!=)", re.M)
 # Comments are stripped first: a value comment like `# 1 = spot` parses as a kwarg
 # named "1" otherwise, and an identifier cannot start with a digit anyway.
 COMMENT = re.compile(r"#[^\n]*")
+# Nested calls are blanked before kwargs are read: matching "at any depth" pulls the
+# INNER call's kwargs out as if they belonged to the config, so
+# `reward_function=partial(fn, alpha=0.5)` yields a phantom `alpha`. That either cries
+# wolf or -- worse -- passes silently when the inner name collides with a real field.
+# No doc does this today; the point is that the next one to try cannot break the check.
+NESTED = re.compile(r"\([^()]*\)")
 
 
 def _documented_config_kwargs():
@@ -107,7 +113,10 @@ def _documented_config_kwargs():
     for path in _doc_sources():
         for block in PY_BLOCK.findall(path.read_text()):
             for cls_name, body in CONFIG_CALL.findall(COMMENT.sub("", block)):
-                for kwarg in KWARG.findall(body):
+                flat = body
+                while NESTED.search(flat):
+                    flat = NESTED.sub(lambda m: " " * len(m.group(0)), flat)
+                for kwarg in KWARG.findall(flat):
                     yield str(path.relative_to(REPO)), cls_name, kwarg
 
 
--- a/tests/test_documented_imports.py
+++ b/tests/test_documented_imports.py
@@ -1,11 +1,11 @@
 """The docs are executed here, because prose cannot be verified by review.
 
 Scope: `from torchtrade... import X` (flat AND parenthesized), plus a syntax check on the
-python blocks in the in-package READMEs. Config kwargs and observation keys live in prose
-and are not covered.
+python blocks in the in-package READMEs. Config kwargs are covered below; observation keys still are not.
 """
 
 import ast
+import dataclasses
 import importlib
 import pathlib
 import re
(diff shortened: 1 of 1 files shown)


BY571:
@CharlieHelps please review — small PR, but it changes how a whole class of doc rot is caught, so the design is worth a second opinion more than the code is.

**The problem it addresses**: `tests/test_documented_imports.py` passed **all 257 cases** through three review rounds on #367 while `margin_call_threshold`, `rollout_steps` and `trading_mode` sat in the docs with **zero occurrences anywhere in the package**. Its docstring said why — it executes *import* blocks, so it validates that a **name resolves**, not that a **call is constructible**. Every defect lived one level below that boundary.

Checking documented `*Config(...)` kwargs against the dataclass fields finds **58 more**, which is the remainder of #287 that has until now been a hand-written list in the issue body.

**Where I'd value your judgement:**

1. **Is the ratchet the right shape?** Every entry in `KNOWN_BROKEN` fails today. I verified it in both directions — a *new* bad kwarg fails immediately, and a listed entry that gets *fixed* also fails until removed. That second half is what I think makes it a ratchet rather than a suppression list, but I'd rather you told me if 58 xfails is the wrong trade against just fixing the docs.

2. **Coverage honesty.** `_resolve_config` searches six modules and `pytest.skip`s anything it cannot resolve. Skipped entries are *unchecked docs that look checked* — the exact failure mode this PR exists to fix, reproduced one level up. I have asked round 1 to enumerate them; if the list is long, the skip should probably be a failure.

3. **Regex reach.** The extraction only matches multi-line `*Config(` calls with 4-space kwarg indentation. Single-line calls and other shapes are silently missed, and `test_the_kwarg_sweep_found_configs_to_check` only asserts >40 — a weak floor. If most docs use a shape it misses, the check is mostly decorative.

Full suite 3385 passed, 58 xfailed.

---

Also: **#364 is back with you** after your third review. I re-derived your side-specific condition and it holds — but your `L=2, mmr=0.004, fee=0.004` example needed one more step: it *is* a floor for longs (50.2008 vs 50.4) and genuinely inverts for **shorts** (149.402 vs 149.6). So the check now applies each direction's constraint only when `action_levels` can actually take that direction, which accepts your case for a long-only config and still refuses it when shorts are enabled.

BY571:
@CharlieHelps both findings are closed at `e616e33` — your review landed on the head before the two coverage commits. Verified by execution, not by reading:

**1. "The extractor misses documented kwargs (single-line `OneStepTradingEnvConfig(...)`, `MarketScannerConfig`)"**

```
OneStepTradingEnvConfig: 19 kwargs captured   (resolvable: True)
MarketScannerConfig:      5 kwargs captured   (resolvable: True)
total triples: 307   (was 289)
```

`CONFIG_CALL` now tolerates arbitrary indentation **and** single-line calls, and `_resolve_config` gained `torchtrade.envs.live.polymarket` — which is why `MarketScannerConfig` went from a silent `pytest.skip` to 5 checked kwargs. Skips went 12 → 1.

You were right that the `> 40` floor is far too weak to notice this. Rather than raise the number, `e616e33` asserts the **invariant**: every `*Config(` in a documented block must yield a parse. It cannot be satisfied by a pattern that silently drops calls, whatever the count. Confirmed it fires — injecting a doubly-nested call gives `README.md: 1 calls, 0 parsed`.

**2. "`KNOWN_BROKEN` can remain stale when a tuple is deleted *or renamed*"**

`test_no_known_broken_entry_has_gone_stale` asserts every entry still appears in `CONFIG_KWARGS`. Tested on the harder half of your case — a **rename**, not a deletion:

```
renaming intervals -> intervals_renamed in docs/environments/online.md
AssertionError: 1 KNOWN_BROKEN entries are no longer in the docs -- delete them:
  [('docs/environments/online.md', 'BinanceFuturesTradingEnvConfig', 'intervals')]
```

508 passed, 1 skipped, 58 xfailed. The one remaining skip is `logging.basicConfig(level=...)` from a doc snippet — `\b(\w*Config)\(` matches stdlib `basicConfig` too; it resolves to `None` and skips, which is correct if slightly noisy.

Ready for a re-look whenever convenient.

PR #373, opened 2026-08-17
docs: close #287 — the SL/TP example put a long's stop above its entry

### LINKED ISSUE #287: Docs sweep: half-finished envs.live path rename, ~10 documented config params that raise TypeError, wrong observation keys, phantom symbols in package READMEs

BY571:
Closes #287.

The last slice of #287: the in-package READMEs the earlier PRs (#370, #371, #372) did not reach.

## The one that matters

`utils/README.md` called two functions that have never existed. An earlier PR corrected the *import line* to the real `calculate_bracket_prices` and left the **body** calling `calculate_sltp_prices` and `check_sltp_hit`. That is precisely why it survived three doc PRs: the import test checks `from x import y` lines, so a block can import the right name and then call the wrong one indefinitely.

Rewriting it against the real function surfaced something worse than a rename. **`sl_pct` is signed, and the sign differs by side** — negative for a long, positive for a short. The example passed `sl_percent=0.02` (positive, and the wrong parameter name) for a LONG:

```
calculate_bracket_prices("long", 100.0, -0.02, 0.05) -> (98.0, 105.0)   correct
calculate_bracket_prices("long", 100.0,  0.02, 0.05) -> (102.0, 105.0)  stop ABOVE entry
```

The helper does not normalise the sign, so anyone copying the block got a bracket that exits on the way **up**. The convention is now stated explicitly, and the block also demonstrates `stop_fill_price` — which is what actually answers "did it hit", since a bar gapping through the level fills at the open, not at the level.

## The rest

- `core/README.md` documented four reward **classes** that do not exist (`LogReturnReward`, `PercentReturnReward`, `RealizedPnLReward`, `SharpeRatioReward`) and instantiated one: `reward_fn=SharpeRatioReward()`. Rewards here are plain **functions** (`log_return_reward`, `sharpe_ratio_reward`, `drawdown_penalty_reward`) and the kwarg is `reward_function`. Every noun in that example was wrong.
- `core/README.md` class tree named `SeqLongOnlyEnv`/`SeqFuturesEnv`; replaced with the four real env names.
- `envs/README.md` listed a `RewardFunction` abstraction that does not exist.
- Root README claimed "Python 3.8+" against `requires-python = ">=3.11"`.

## Guard

New test for the *class* of defect rather than these instances: a documented **call** to a name the block never binds, no earlier block in the same file defines, and torchtrade does not export. Blocks are read in order — as a reader reads them — so `MyEnv(...)` in a subclassing walkthrough stays legal while a call to nothing does not. Mutation-checked in both directions.

It required restoring `import ast`, removed in #372 alongside `_safe_walk`; that removal was correct then and this guard genuinely needs it back.

Full suite **3629 passed**, 0 xfailed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ See **[Examples Documentation](https://torchtrade.github.io/torchtrade/examples/
 
 ### Prerequisites
 
-- Python 3.8+
+- Python 3.11+
 - CUDA (optional, for GPU acceleration)
 - [UV](https://docs.astral.sh/uv/) - Fast Python package installer
 
--- a/tests/test_documented_imports.py
+++ b/tests/test_documented_imports.py
@@ -4,6 +4,10 @@ Scope: `from torchtrade... import X` (flat AND parenthesized), plus a syntax che
 python blocks in the in-package READMEs. Config kwargs are covered below; observation keys still are not.
 """
 
+import ast
+import builtins
+import sys
+import functools
 import dataclasses
 import importlib
 import pathlib
@@ -41,10 +45,16 @@ def _documented_imports():
 
 
 CASES = list(_documented_imports())
+_README_SOURCES = [p for p in _doc_sources()
+                   if p.name == "README.md" and p.is_relative_to(REPO / "torchtrade")]
 README_BLOCKS = [
     pytest.param(block, id=f"{p.relative_to(REPO)}::block{i}")
-    for p in _doc_sources() if p.name == "README.md" and p.is_relative_to(REPO / "torchtrade")
-    for i, block in enumerate(PY_BLOCK.findall(p.read_text()))
+    for p in _README_SOURCES for i, block in enumerate(PY_BLOCK.findall(p.read_text()))
+]
+# (file, index, block) for the guard that needs to see a block's predecessors.
+README_BLOCKS_IN_CONTEXT = [
+    pytest.param(p, i, block, id=f"{p.relative_to(REPO)}::block{i}")
+    for p in _README_SOURCES for i, block in enumerate(PY_BLOCK.findall(p.read_text()))
 ]
 
 
@@ -205,3 +215,71 @@ def test_no_readme_redefines_a_package_config_as_a_dataclass():
         f"{len(offenders)} README blocks redefine a real config as a dataclass instead "
         f"of importing it: {offenders}"
     )
+
+
+# Names a reader supplies, or that a block deliberately continues from an earlier one.
+@functools.lru_cache(maxsize=None)
+def _resolves_anywhere(name: str) -> bool:
+    """True if `name` is exported by any already-imported torchtrade module.
+
+    Only modules the doc sweep has already pulled in are searched -- importing the whole
+    package to answer this would make the test's cost unbounded and its failures depend
+    on import order.
+    """
+    for module in list(sys.modules.values()):
+        if getattr(module, "__name__", "").startswith("torchtrade") and hasattr(module, name):
+            return True
+    return False
+
+
+def _names_bound_by(block):
+    try:
+        tree = ast.parse(block)
+    except SyntaxError:
+        return set()
+    bound = {n.id for n in ast.walk(tree) if isinstance(n, ast.Name) and isinstance(n.ctx, ast.Store)}
+    bound |= {a.asname or a.name.split(".")[0] for n in ast.walk(tree)
+              if isinstance(n, (ast.Import, ast.ImportFrom)) for a in n.names}
+    bound |= {n.name for n in ast.walk(tree) if isinstance(n, (ast.FunctionDef, ast.ClassDef))}
+    bound |= {a.arg for n in ast.walk(tree) if isinstance(n, ast.arguments) for a in n.args}
+    return bound
+
+
+# Objects a reader is expected to bring; naming one is not a claim about the package.
+_READER_SUPPLIED = {"env", "config", "action", "agent", "df", "td", "transition", "obs",
+                    "your_dataframe", "policy", "model", "data", "trainer"}
+
+
+@pytest.mark.parametrize("path,index,block", README_BLOCKS_IN_CONTEXT)
+def test_a_documented_block_calls_nothing_that_does_not_exist(path, index, block):
+    """A CALL to a name the block never binds and the package never defines.
+
+    This is the gap the import test cannot cover: it checks `from x import y` lines, so
+    a block whose import line was corrected while its BODY kept calling the old name
+    passes it. That is exactly what happened -- utils/README.md imported the real
+    calculate_bracket_prices and then called calculate_sltp_prices and check_sltp_hit,
+    neither of which has ever existed, three doc PRs in a row.
+
+    Only flags names that look like package API (snake_case functions, CamelCase
+    classes) and resolve nowhere in torchtrade. Anything the reader defines is theirs.
+    """
+    try:
+        tree = ast.parse(block)
+    except SyntaxError:
+        pytest.skip("covered by the compile test")
+
+    # A reader meets the blocks in order, so anything an EARLIER block defined is in
+    # scope here -- that is what makes `MyEnv(...)` in a subclassing walkthrough legal
+    # while `calculate_sltp_prices(...)`, defined nowhere at all, is not.
+    bound = set(_names_bound_by(block))
+    for earlier in PY_BLOCK.findall(path.read_text())[:index]:
+        bound |= _names_bound_by(earlier)
+
+    called = {n.func.id for n in ast.walk(tree)
+              if isinstance(n, ast.Call) and isinstance(n.func, ast.Name)}
+    unknown = sorted(
+        name for name in called - bound - _READER_SUPPLIED
+        if not hasattr(builtins, name) and _resolve_config(name) is None
+        and not _resolves_anywhere(name)
+    )
+    assert not unknown, f"calls names that exist nowhere in torchtrade: {unknown}"
--- a/tests/test_documented_imports.py
+++ b/tests/test_documented_imports.py
@@ -264,7 +264,7 @@ def _resolves_anywhere(name: str) -> bool:
     return name in _package_names()
 
 
-def _names_bound_by(block, *, include_params=True, before_line=None):
+def _names_bound_by(block, *, before_line=None):
     """Names bound at MODULE level in `block`, optionally only those defined earlier.
 
     Module level, and line-ordered, because a reader executes the fence top to bottom.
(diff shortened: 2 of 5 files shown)


BY571:
@CharlieHelps review requested.

This is the last slice of #287 — the in-package READMEs the earlier PRs didn't reach.

The finding worth your attention is not a rename. `utils/README.md` imported the real `calculate_bracket_prices` (an earlier PR fixed the import line) while its **body** still called `calculate_sltp_prices` and `check_sltp_hit`, neither of which has ever existed. Rewriting it against the real function exposed a sign trap: `sl_pct` is signed and the sign differs by side, and the example passed `+0.02` for a **long**, which puts the stop 2% *above* entry — a bracket that exits on the way up. The helper does not normalise it.

`core/README.md` documented four reward *classes* that don't exist and instantiated one (`reward_fn=SharpeRatioReward()`); rewards here are plain functions and the kwarg is `reward_function`.

Given this PR's history I'd particularly value scepticism on the new guard: it flags a documented **call** to a name the block never binds, no earlier block in the same file defines, and torchtrade doesn't export. I made it read blocks in order so `MyEnv(...)` in a subclassing walkthrough stays legal. Please check whether that ordering rule can be gamed into false negatives — a phantom call in block 1 of a file whose block 0 happens to bind the same name would slip through, and I have not proven that case impossible.

Suite 3629 passed, 0 xfailed. `uv sync` at the repo root provisions `.venv` with NumPy if the devbox lacks it.

BY571:
@CharlieHelps rounds 1 and 2 applied — please review `5ef8433`. Both rounds found the fix defective in the same way, so this needs a sceptical read rather than a confirmatory one.

**Round 1.** The PR renamed `sl_percent=` → `sl_pct=` on `self._init_sltp(...)` — a method that has **never** existed, along with `_check_sltp_triggered`, `_execute_sltp_exit`, `_check_sltp` and `has_position`. `SLTPMixin` has four members. I polished a kwarg on a phantom, half-renamed it (left `tp_percent=`), and missed the second copy of the same phantom ~100 lines below in the same file.

The guard I added was also **unsound**: `_resolves_anywhere` scanned `sys.modules`, so it saw 71 torchtrade modules under `-k` and 102 after the import tests ran. `ReplayObserver` — real, exported — resolved `False` in one case and `True` in the other. Identical file, opposite verdicts, guaranteed flake under `-k` or xdist. It walks the package with `pkgutil` once now.

**Round 2, and this is the one I'd want you to weigh.** `core/README.md:255` still carried `next_obs, reward, done, info = env.step(action)` — **143 lines below the comment round 1 added forbidding exactly that**, in a commit whose message called it "the idiom #372 spent three rounds removing." And round 1's replacement was itself a phantom: `return super()._step(td)` resolves to torchrl's `EnvBase._step`, an abstractmethod that raises. The class couldn't be instantiated either — `TorchTradeOfflineEnv` leaves two abstract methods and the block overrode one.

Also removed, all in the two files this PR edits: `spaces.Box`/`spaces.Discrete` under headings claiming environments "must define" spaces (gym is not a dependency and appears nowhere in the package); four phantom hooks presented as the required interface (`_make_observation_space`, `_make_action_space`, `_initialize`, `_finalize`); `TimeFrameUnit.MINUTE`, which raises (`Day`/`Hour`/`Minute`); and `create_alpaca_sltp_action_map()` documented as returning `{0: "BUY", 1: "SELL", 2: "HOLD"}` when it takes two required arguments and returns a bracket map.

**Known open, deliberately not bodged:** the guard only inspects `ast.Name` calls, so attribute calls (`self.foo()`, `obj.bar()`) are unchecked — which is precisely how `_init_sltp` survived three rounds and why the guard passed round 1's `super()._step()`. The tractable slice is: for a `ClassDef` whose bases resolve to a real torchtrade class, check `self.X()` against `dir(base)`. I'd rather file that than rush it into this PR — tell me if you disagree and want it here.

Suite 3624 passed, 0 xfailed.

BY571:
@CharlieHelps round 3 applied at `d5e6814`. Three rounds done; each one found the previous round's fix defective, so please read this sceptically rather than confirmatorily.

**Round 3's finding is the pattern in miniature.** Round 2 replaced a phantom `super()._step()` call with `return self.cash` — and `cash` does not exist on the offline envs either (the attribute is `self.balance`; `cash` appears only in polymarket). A phantom method swapped for a phantom attribute, one line away. The override signature was also wrong: the base calls `_get_portfolio_value(initial_price)`, so the no-arg version raises on reset. And filling both abstract methods still isn't enough — the base `_reset` calls `self._get_observation()`, which the base neither defines nor declares abstract.

**The signature defect, fourth occurrence:** the `TimeFrameUnit` prose listed six members in the wrong case (SECOND, MINUTE, HOUR, DAY, WEEK, MONTH) **twenty lines above** the fence where round 2 had just corrected `TimeFrameUnit.MINUTE` → `.Minute`. The real enum is `Minute`/`Hour`/`Day`. Fixing code and leaving the prose describing it has now happened in every round of this PR.

Also: `timeframe_to_alpaca(tf)  # "1Day"` returns an alpaca `TimeFrame` **object**, not a string — sitting right next to `timeframe_to_binance(tf)  # "1d"`, which genuinely is one. A **dead assertion** (`assert transition["next","done"] is not None` on a tensor, which can never be None). And `market_data_1Min_10` in the spec example, where a real env emits `market_data_1Minute_10`.

**Guard fix worth your attention:** its completeness assert turned a missing optional extra into a false failure. Blocking `vllm` (an `[llm]` extra) made two modules unimportable, fired the assert, and failed a doc block containing no phantom at all. Walking and importing are now separated — a subpackage that cannot be *walked* is fatal (that silently truncates the name set, making real API read as a phantom), while a module that cannot be *imported* is tolerated. Verified: with vllm blocked the walk still yields 364 names and nothing fires. The `while True/next()` construct was also dead code (with `onerror` set nothing escapes the generator) and its except branch would have swallowed failures without recording them — the exact hole its comment claimed to close.

The attribute-call gap remains open and is now stated outright in the guard's docstring rather than papered over. Happy to add the `ClassDef`-scoped `self.X()` check here if you'd rather not have it as a follow-up.

CI green, suite 3624 passed, 0 xfailed. Ready for your verdict.

BY571:
@CharlieHelps both addressed at `a18857b`.

**Your guard blocker was right and it was the sharpest finding on this PR.** `_names_bound_by` walked the entire AST, so it accepted two bindings Python does not:

```python
phantom(); def phantom(): ...     # forward def — NameError when the fence actually runs
def helper(): phantom = 1         # a nested local masking a later top-level call
```

Both let a real phantom through while the block itself would raise — the guard was validating something weaker than "this code runs". Bindings now come from module-level statements only, and **each call is judged against what is in scope at its own line**, so a definition further down cannot retroactively legitimise a call above it. Parameters still count inside their own function, where they genuinely are in scope. Mutation-checked: your `phantom(); def phantom()` case fails the guard now and passed it before.

**On the SLTP walkthrough** — that was fixed in round 1, which landed after your review was posted, so you were reading the earlier tree. `_init_sltp`, `_check_sltp_triggered`, `_execute_sltp_exit` and `config.sl_pct` now appear **zero** times in the file. You were right that changing only the kwarg was the wrong move: I had renamed `sl_percent=` → `sl_pct=` on a call to a method that never existed, missed the `tp_percent=` on the next line, and left a second copy of the same phantom ~100 lines below. Both blocks are rewritten against the real four-member `SLTPMixin` surface and the duplicate is collapsed.

Rounds 2 and 3 (already pushed) found more of the same, including one worth flagging since it bears on your "copy-runnability" framing: round 2 replaced a phantom `super()._step()` with `return self.cash`, and `cash` does not exist on the offline envs either — the attribute is `self.balance`. A phantom method swapped for a phantom attribute, one line away. Both are invisible to the guard because they are **attribute** calls/accesses, which it does not cover; that gap is now stated in its docstring rather than papered over.

CI green, suite 3624 passed, 0 xfailed.

BY571:
@CharlieHelps all three fixed at `8f4449d`. Thank you for reviewing from scratch rather than trusting the round summaries — blocker 1 is the single best catch of this PR.

**1. The action-map example recreated the exact bug this PR exists to fix.** I used `[0.02]` — a positive stop for a long — in a different block of the file where I had just fixed that same sign error. The map preserves the sign, so the documented value flows straight through:

```
create_alpaca_sltp_action_map([0.02], [0.05])       -> {1: (0.02, 0.05)}
calculate_bracket_prices("long", 100.0, 0.02, 0.05) -> (102.0, 105.0)   stop ABOVE entry
```

Now `[-0.02]` → `(98.0, 105.0)`, with the non-normalising behaviour stated. This is my recurring defect in its purest form: fixing the instance I was looking at and writing the same error two sections away.

**2. You were right that my scope claim was still false.** The previous commit said each call is checked against names in scope at its line; parameters were still collected block-wide, so your case passed while raising `NameError`. Scope is now threaded down the tree — a parameter is visible inside the function that declares it and nowhere else. Mutation-checked **both** ways: your `def helper(phantom)` + top-level `phantom()` fails, and a legitimate `def helper(real_param): return real_param()` still passes.

**3. The testing walkthrough could never run.** It instantiated the deliberately-incomplete `MyCustomEnv`, which satisfies the ABC and still raises `AttributeError` on `reset()`. My warning comment described the hole accurately and left the example broken — describing a defect is not fixing it. It uses a concrete `SequentialTradingEnv` now, and I **executed the documented test body end to end** against real data rather than reasoning about it: `check_env_specs succeeded`, reset/step/done all work.

Your non-blocking prose point is fixed too — the cardinality claim ignored the short half; index 0 is flat, the long grid adds `len(sl) * len(tp)`, and `include_short_positions=True` adds that many again.

Known and stated, not papered over: attribute calls (`self.foo()`, `obj.bar()`) remain outside the guard, which is how `_init_sltp` survived three rounds. Closing it needs the callee's type; the tractable slice is checking `self.X()` against `dir(base)` for a `ClassDef` whose bases resolve to a real torchtrade class. Say the word if you want it in this PR rather than filed.

Suite 3626 passed, 0 xfailed.

PR #374, opened 2026-08-17
fix(replay): the recorded PV was the previous bar's equity, and bybit refused every open

BY571:
Partial fix for #278 — the two defects that corrupt numbers a policy is evaluated on. Defect 2 (fractional+fee) was already closed by #369.

## Every replayed reward belonged to the previous bar

`_acquire_post_bar_state` returned `(self._get_portfolio_value(), self._get_observation())`. A tuple evaluates left to right, and under a `ReplayObserver` the clock advances only **inside** `get_observations()` — so the portfolio value was sampled before the bar moved. Measured over an 8-step rollout:

```
recorded PV == equity at the DECISION bar (stale): 8/8
recorded PV == equity at the NEXT bar (correct)  : 0/8
```

Worse, within a *single* step the emitted `account_state[2]` was already at the new bar while the recorded PV was at the old one — observation and reward disagreeing about which bar it is. A take-profit bar recorded `reward=+0.000000`. After the swap the counts invert exactly: **0/8 stale, 8/8 correct**.

Three sites cover all ten live envs: the shared futures base (binance, bitget, bybit, okx) plus alpaca's `env` and `env_sltp`, which keep their own copies.

## Bybit's replay path refused 100% of opens

Not in the issue — found while scoping. It derived rounding precision by string-inspecting the step:

```python
len(str(qty_step).rstrip('0').split('.')[-1]) if '.' in str(qty_step) else 0
```

`str(1e-06)` is `'1e-06'` — no `'.'` — so decimals came out 0 and `round(0.977, 0)` sent **1 whole BTC**, needing 10025 against a 10000 balance. Refused with a `logger.warning` and no raise, so replay reported a flat strategy for a policy that was trading: the same symptom #369 fixed one code path over.

Five sites shared that rule across four venues. Only bybit's operates on a float, where `repr` goes scientific, so only it breaks today — the rest are latent. One `decimals_for_step()` now answers it from `Decimal`'s exponent, where notation cannot change the result.

Restoring the shared helper flushed out a fail-open worth naming: dropping a variable the cache still referenced raised `NameError` **inside a broad `except Exception`**, so the lot-size cache silently never populated and re-fetched every call. The test that caught it asserts `call_count`, not behaviour.

## Tests

The PV test pins **order** via source text, because the lag is invisible to a live-mocked value assertion — the two calls look symmetric and are not. Mutation-checked: reverting the swap fails the test *and* returns the rollout to 8/8 stale. `decimals_for_step` is parametrized over both notations, str and float inputs, integral steps (where a negative `ndigits` would raise), and junk.

#278's remaining items, to keep this reviewable: `reset()` never rewinding (episode 2 inherits balance, position and bankruptcy baseline), `base_features` 9/10 rows zero-filled, the spec/key mismatches, and `reduce_only` discarding `quantity` (zero call sites today). Each is scoped with a repro.

Full suite **3641 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/alpaca/test_torch_env_sltp.py
+++ b/tests/envs/alpaca/test_torch_env_sltp.py
@@ -392,10 +392,25 @@ def test_alpaca_history_records_price_and_position_on_a_flat_bar(sltp):
     )
 
     step(open_action)
-    step(hold_action)
-    # The recorded size is the one held ENTERING the bar, as on every other live env, so
-    # it appears on the bar AFTER the one that opened the position.
+    # POST-trade, so the size appears on the bar that opened the position -- not the one
+    # after. Recording the entering size against a post-bar price and portfolio value
+    # labelled the return with the exposure that did not produce it, and is what offline
+    # has always done (#278).
     assert env.history.positions[-1] != 0, (
-        "an open position must reach history.positions, not stay at the 0.0 default"
+        "the opening bar must record the position it opened, not the flat size it "
+        "entered with"
     )
+    # The next bar records where it ENDED, which differs by variant: on the plain env
+    # `hold_action` is index 0 of [0.0, 0.5, 1.0], a 0% allocation that exits, so the row
+    # reads flat. On the SLTP env index 0 is a no-op and the position exits only via
+    # SL/TP, so it is still open. Under the old pre-trade semantics the plain env's
+    # closing row claimed the exposure it had just given up.
+    step(hold_action)
+    if sltp:
+        assert env.history.positions[-1] != 0, "the position is still open on this bar"
+    else:
+        assert env.history.positions[-1] == 0, (
+            "the closing bar must record the flat size it ended at, not the one it "
+            "entered with"
+        )
     env.close()
--- a/tests/envs/replay/test_pv_ordering_278.py
+++ b/tests/envs/replay/test_pv_ordering_278.py
@@ -1,35 +1,102 @@
-"""The recorded portfolio value must belong to the bar the action was taken on (#278)."""
+"""The recorded portfolio value and price must belong to the bar the action moved to (#278)."""
 
-import inspect
+from unittest.mock import patch
 
+import numpy as np
+import pandas as pd
 import pytest
 
-from torchtrade.envs.live.alpaca import env as alpaca_env
-from torchtrade.envs.live.alpaca import env_sltp as alpaca_env_sltp
-from torchtrade.envs.live.shared import futures_live_base
+from torchtrade.envs.live.bybit.env_sltp import (
+    BybitFuturesSLTPTorchTradingEnv,
+    BybitFuturesSLTPTradingEnvConfig,
+)
+from torchtrade.envs.replay.observer import ReplayObserver
+from torchtrade.envs.replay.order_executor import ReplayOrderExecutor
 
 
-@pytest.mark.parametrize("module,label", [
-    (futures_live_base, "futures_live_base"),   # binance, bitget, bybit, okx
-    (alpaca_env, "alpaca.env"),
-    (alpaca_env_sltp, "alpaca.env_sltp"),
-])
-def test_the_observation_is_acquired_before_the_portfolio_value(module, label):
-    """Order, not a value, because the lag is invisible to a live-mocked assertion.
+def _df(n=80):
+    ts = pd.date_range("2024-01-01", periods=n, freq="1min")
+    prices = np.linspace(50000, 55000, n)  # steep, so consecutive bars differ visibly
+    return pd.DataFrame({"timestamp": ts, "open": prices, "high": prices + 10,
+                         "low": prices - 10, "close": prices, "volume": np.ones(n) * 100})
 
-    Under a ReplayObserver the clock advances only inside `get_observations()`, so
-    reading the portfolio value first records the PREVIOUS bar's equity against this
-    bar's action: measured 8/8 against the decision bar and 0/8 against the next one,
-    while `account_state` in the same step was already at the new bar. A tuple evaluates
-    left to right, so `(self._get_portfolio_value(), self._get_observation())` reads
-    them in the wrong order while looking symmetric -- which is why this pins the
-    source text rather than a number.
+
+def test_the_recorded_portfolio_value_is_the_bar_the_action_moved_to():
+    """Behavioural, not a source-text check.
+
+    The first version of this test asserted that `_get_observation()` appeared before
+    `_get_portfolio_value()` in the module source. That is one docstring away from
+    vacuous -- `str.find` takes the first occurrence anywhere, prose included, and this
+    fix's own comments name both methods. Drive the value through instead.
+
+    The bug: a tuple evaluates left to right, and under a ReplayObserver the clock
+    advances only inside `get_observations()`, so reading PV first recorded the PREVIOUS
+    bar's equity against this bar's action -- 8/8 against the decision bar, 0/8 against
+    the next.
     """
-    source = inspect.getsource(module)
-    obs_at = source.find("self._get_observation()")
-    pv_at = source.find("self._get_portfolio_value()")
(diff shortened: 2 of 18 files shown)


BY571:
@CharlieHelps review requested — this one changes numbers rather than prose, so it deserves a harder look than the docs PRs.

**The core claim is a call-ordering bug.** `_acquire_post_bar_state` returned `(self._get_portfolio_value(), self._get_observation())`. A tuple evaluates left to right, and under a `ReplayObserver` the clock advances only *inside* `get_observations()` — so every recorded portfolio value was the previous bar's equity. Measured 8/8 against the decision bar, 0/8 against the next; after the swap, exactly inverted.

The part I'd most like challenged: **is observation-before-PV right for the LIVE path too, or only for replay?** My reasoning is that live already waits for the bar in a separate step, so moving the PV read after the observation is at worst a no-op and at best removes the same staleness. But I've only proven it on replay, and this changes the order of two real broker calls on ten live environments.

**The test pins source order rather than a value**, which I want you to push back on if you think it's the wrong call. My justification: the lag is invisible to a live-mocked value assertion because both calls return whatever the mock says regardless of order — the bug is *which* bar the value came from, and a mock has no bars. Mutation-checked: reverting the swap fails the test and returns the rollout to 8/8 stale.

**Second defect, not in the issue** — bybit refused 100% of opens because it derived rounding precision from `str(qty_step)`, and `str(1e-06)` has no `'.'`, so decimals came out 0 and `round(0.977, 0)` sent 1 whole BTC. Five sites across four venues shared that rule; only bybit's operates on a float, so only it breaks today. Now one `decimals_for_step()` reading `Decimal`'s exponent.

Worth flagging honestly: consolidating those five sites, I dropped a variable the okx lot-size cache still referenced. The resulting `NameError` was swallowed by a broad `except Exception`, so the cache silently never populated. A test caught it only because it asserts `call_count`. That fail-open is still there and is arguably a defect of its own.

Suite 3641 passed, 0 xfailed.

BY571:
@CharlieHelps rounds 1 and 2 applied — please review `cbccb94`. Two of my own fixes were defective, and one was a real-money risk, so this needs the sceptical read.

**Round 1's fix could emergency-flatten a live position.** To align the recorded price with the recorded equity I called `_current_mark_price()` inside `_halting`. That method raises on a non-positive mark — so a blank post-bar ticker would raise `LiveObservationHalt` and, under `FLATTEN`, **market-close a real position**. bybit and okx go further and raise `RuntimeError` on a blank ticker, which `_halting` deliberately does not catch, aborting the episode uncaught. A price that only labels a history row is not worth a market order.

**And it read from the wrong moment regardless.** With no `position_status` it takes the `else` branch — a fourth venue round-trip per step — so the recorded price came from a *different* snapshot than the `unrealized_pnl_pct` and `exposure_pct` in the same row. That is precisely the two-facts-two-moments defect this method exists to remove, re-created on the live path while fixing it on replay. The mark now comes from the snapshot `_get_observation` already holds: no extra call, no new failure mode, same moment.

**The price test I added was vacuous** — it asserted only that the recorded price was within 1.0 of *some* close in the frame, which every bar satisfies. Reverting `price=` to the pre-trade mark left it **passing**. It now pins price and equity to the same post-step snapshot and fails under either mutation, with a fixed action instead of the unseeded `action_spec.rand()`.

**My round-1 commit message claimed bybit and okx were both made atomic — bybit was never touched**, nor was binance. Both fixed now.

Two more worth your eye: okx applied "both or neither" to the tick pair but not the lot pair in the same method (a blank `minSz` raising between them left the cache `None`, falling back to a 0.001 step — `_format_size(0.977065)` → `'1.0'`, the #278 rounding bug again); and making `decimals_for_step` total silently turned binance's LOT_SIZE guard from **fail-closed to fail-late** — a NaN `stepSize` used to abort construction and instead surfaced at order time. An explicit `isfinite` check restores the contract its own comment demands.

Still open and stated rather than hidden: `position=` in the history row is the **pre-trade** size while the price and PV are now post-bar, so the row labels a return with the exposure that did not produce it. Offline records the post-trade size. I'd rather you tell me whether that belongs in this PR or a follow-up than guess.

Suite 3640 passed, 0 xfailed.

BY571:
@CharlieHelps round 3 at `24f75cd`. Three rounds done, and each found the previous round's fix defective — please read sceptically.

**Round 2's fix regressed what round 1 fixed.** I claimed "flat rows carry no PnL either way" and that was wrong about the row that matters most: `_last_observed_mark` is set only on the open-position branch, so every row ending flat fell back to the pre-trade mark — **and the exit row is flat and carries the realized PnL**. Measured 12 of 14 rows with `price[t] == close[t-1]` against `portfolio_value[t] == equity[t]`; binance plain worst at 1/6 correct. `base_prices` feeds the buy-and-hold baseline the policy is scored against, so a shifted price series shifts the benchmark.

Flat rows fetch the mark now but **never fatally** — which was round 2's whole reason for not fetching. An unavailable or non-positive mark returns `None` and the caller records the pre-trade price, rather than raising inside `_halting` where `FLATTEN` market-closes a live position.

**My test could not see the regression.** It opened at step 0 and held, so with a 3% take profit on a gentle series the position never closed and no flat row ever existed — the regression passed 4/4. The TP is 0.001 now so the exit actually fires. Re-mutated: all four venues fail with round 2's behaviour restored, all four pass without it. It is also parametrized across bybit/binance/bitget/okx, since the change is byte-identical across eight files and covering one venue is how this survived a round.

**bybit still was not atomic.** Round 2's message said "bybit was not touched, nor was binance. Both now compute before assigning" — and bybit was not in that commit's stat at all. That is my second false claim about the same file, which is worth flagging as a pattern rather than an accident.

**Alpaca's guard was inert.** `_get_current_price()` *returns* 0.0 when all three sources fail rather than raising, so the try/except I added never fired on the path that actually degrades. It checks the value now.

Two things I am deliberately leaving open rather than guessing, and would like your call on:
1. `position=` is still the pre-trade size while price and PV are post-bar — offline records post-trade, so the row labels a return with the exposure that did not produce it.
2. `docs/environments/online.md`'s halt table is stale (pre-dates this branch) and this PR adds behaviour it does not describe.

CI green, suite 3642 passed, 0 xfailed.

BY571:
@CharlieHelps your blocker is fixed at `472c78e`, and you were right that it belonged in this PR rather than a follow-up.

All ten live history-write paths now record the **post-trade** signed size. The eight futures envs take it from the same observation snapshot as the mark (no extra call); alpaca's two re-read the status non-fatally.

**The regression assertion you asked for took two attempts to be worth anything, and the first attempt is the more useful thing to report.** I wrote it, ran the mutation, and it *passed* — because with the tight take-profit the price test needs, the position opens and closes **inside one bar**, so pre- and post-trade sizes are both `0` on every row and the assertion was comparing zeros to zeros. The two defects need opposite scenarios: a tight tp to produce the flat exit row where the price defect lives, a wide one to keep a position open across bars where the position defect lives. The builder takes `tp` as a dial now. Mutating `position=new_qty` back to `position=position_size` fails; before the split it passed.

Worth noting what this exposed in the existing alpaca contract test: it encoded the old semantics **in its own comment** — *"the recorded size is the one held ENTERING the bar ... so it appears on the bar AFTER the one that opened the position."* It asserts the opening bar now. Its second assertion had to be variant-split, because index 0 is a 0% allocation on the plain env (that bar exits and correctly reads flat) while on the SLTP env it is a no-op and the position is still open. The old pre-trade reading hid that difference by lagging both — which is a small illustration of your point that the lag was masking real behaviour rather than merely shifting it.

Also since your review, from round 3: the flat-row price regression (the exit row is flat and carries the realized PnL, so 12 of 14 rows were a bar behind), bybit's tick assignment which my round-2 message falsely claimed to have fixed, and alpaca's inert guard — `_get_current_price()` *returns* 0.0 rather than raising, so an except-only wrap never fired on the degraded path.

Suite 3646 passed, 0 xfailed. Ready for your verdict.

PR #375, opened 2026-08-17
fix(replay): every episode after the first continued mid-stream

BY571:
Second slice of #278, after #374. This is the `reset()` defect.

## The bug

`ReplayObserver.reset()` existed and **nothing but a test ever called it**. Under a collector that made every episode after the first invalid three ways at once:

- it continued **mid-stream** through the data
- it started from the **previous episode's balance**
- it carried an **inherited position** whose brackets had already been cancelled

## The fix

The five live `_reset`s call `self.observer.reset()` **first**, before the balance, position and price reads that follow — those must see the rewound state. `reset()` is declared as a no-op on the three observation classes rather than guarded by `hasattr` at the call site, which would be fail-open: an observer that renamed the method would silently stop rewinding.

The bankruptcy baseline is re-captured per episode too. `__init__` still sets it so the attribute always exists, but leaving it there alone measured episode 2 against episode 1's opening equity — so a live run already down 40% read as 40% down on its first step.

## Two tests, because one scenario cannot show both

Replay proves the rewind: index, balance, position and baseline all return to the start, and the test first asserts the episode *actually moved the balance* so the rewind isn't trivially satisfied.

But replay **cannot** show the baseline fix — rewinding restores the same number, so capturing once and capturing per episode are numerically identical there. I found this by mutating: the recapture mutant passed. That half is tested on a live mock whose account persists and drops 40% between episodes. Both are mutation-checked independently.

## Test changes worth flagging

The bankruptcy tests across all five venues mocked the collapsed balance **before** `reset()`, relying on the `__init__` capture. An episode that *starts* at 50 is a small episode, not a bankrupt one, so the collapse now happens inside the episode — which is what those tests meant to check. The two alpaca ones overrode `initial_portfolio_value` by hand before reset, where the recapture discards it.

Full suite **3648 passed**, 0 xfailed.

## Still open in #278

`base_features` 9/10 rows zero-filled, the spec/key mismatches (`market_data_1Minute` vs `market_data_1Minute_10`, `(3,0)` vs `(3,5)`), and `reduce_only` discarding `quantity` (zero call sites today). Each scoped with a repro.


BY571:
@CharlieHelps review requested. Second slice of #278 — the `reset()` defect.

`ReplayObserver.reset()` existed and **nothing but a test ever called it**, so under a collector every episode after the first continued mid-stream, started from the previous episode's balance, and carried an inherited position whose brackets had been cancelled.

Two things I'd particularly like challenged:

**1. I declared `reset()` as a no-op on the three observation classes rather than guarding the call site with `hasattr`.** My reasoning is that a `hasattr` guard is fail-open — an observer that renames the method silently stops rewinding and nothing says so. But it does mean any third-party observer must now implement it, and I'd rather you tell me if that trade is wrong.

**2. The bankruptcy baseline is now re-captured per episode**, and this changed the meaning of five existing tests. They mocked a collapsed balance *before* `reset()`, relying on the `__init__` capture. My reading is that an episode which *starts* at 50 is a small episode, not a bankrupt one — so the collapse has to happen within the episode — but that is a semantic call and it silently changes when `done_on_bankruptcy` fires on a live account. Worth your scrutiny.

**Worth reporting because it nearly shipped:** my first version of the regression test could not fail. Replay rewinds the balance to its starting value, so a baseline captured once in `__init__` and one re-captured per episode are numerically identical there — the assertion passed either way. I only caught it by mutating. The baseline half is now tested on a live mock whose account persists and drops 40% between episodes; both halves are mutation-checked independently.

Suite 3648 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 applied at `fb10dfc`. The second judgement call I flagged to you was the right thing to worry about, and it was worse than I framed it.

**Per-episode rebasing removed the drawdown circuit breaker outright** — not just for a slow bleed. An account **halving every episode** reached 0.39% of its starting equity without ever reporting bankrupt:

```
 ep   baseline  end equity  bankrupt?
  0    1000.00      500.00      False
  7       7.81        3.91      False
```

On main that run terminates at ep3. My "matches offline" justification was wrong: offline resets the **balance and the baseline together**, so its ratio starts at 1.0 because the account is reset too. A live account persists, so rebasing only the yardstick makes it chase the equity down. It also turned a fully-liquidated account from *reports bankrupt* into *raises ValueError at reset* — the `<= 0` guard was written as boundary validation for `__init__` and now ran every episode.

The baseline now follows the account: `observer.reset()` returns `True` only when it actually rewound something. `ReplayObserver` does (it restores the executor's balance); live no-ops return `False`. The halving run terminates at ep3 again.

**And a fail-open I nearly shipped on top of that fix:** `if self.observer.reset():` — a `MagicMock` observer returns a **truthy Mock**, so the recapture fired on every venue whose tests mock the observer. Same shape as the `float(MagicMock()) == 1.0` bug earlier in this issue. It is `is True` now.

That gate also made the five venues' bankruptcy-test restructuring unnecessary — live never re-captures, so their original form is correct. **All eight test files reverted to main**, which took 58 lines of churn out of the diff.

On your first question — requiring observers to implement `reset()` rather than `hasattr`-guarding the call — a mutation sweep found that replacing `self.observer.reset()` with `pass` left **four of the five call sites green**, since replay-backed env tests exist for bybit only. So the interface decision was load-bearing and untested. Two structural guards now assert every live `_reset` rewinds its observer and that every observer class declares `reset()` returning `False`; mutation-checked on all four previously-uncovered venues.

Suite 3650 passed, 0 xfailed. Round 2 is running.

BY571:
@CharlieHelps round 2 at `6a2a6a7`. Round 1's fix turned out to buy nothing, so most of it is now deleted rather than repaired.

**The bankruptcy re-capture was a numerical no-op in every shipped configuration.** Live observers returned `False` so the branch never ran; `ReplayObserver` returned `True`, but `ReplayOrderExecutor.reset()` restores `initial_balance` — so the re-captured value was always identical to the `__init__` one. Measured over four episodes of a *losing* replay run: `[10000.0, 10000.0, 10000.0, 10000.0]`, unchanged. My own test file admitted this in a docstring and I kept the branch anyway. So the branch, the `bool` return contract, the `is True` gate and the `ReplayObserver` override are all gone — roughly 100 lines for zero behaviour.

Deleting it removed two live defects it was carrying:

- `ReplayObserver.reset()` returned `True` even when `executor is None` — the constructor default, and a construction that already exists in the test suite. That rebased the baseline having rewound **no account at all**: round 1's own headline failure verbatim, with a halving account reaching 0.78% of its start without terminating.
- `is True` was shaped by test doubles rather than production. It silently skips a truthy non-`True` (numpy `bool_`, an `int` subclass), and the only reason it was needed is that eight venue fixtures use a bare `MagicMock` observer.

**And the structural guard I added was a substring check that a comment satisfies.** Replacing okx's live line with `# NOTE: self.observer.reset() handled by the caller now.` left the **entire suite green** while okx's second episode continued mid-stream. The `"super()._reset(" not in src` escape widened it further — any `_reset` whose docstring mentioned that string was exempt. It parses now, matching the four older AST guards in the same file, and is mutation-checked against exactly that comment.

Two corrections to my round 1 message, since I would rather flag them than let them stand: ten venue test files were reverted, not eight; and at the threshold the tests use, the halving run terminates at episode **5** on main, not 3.

What remains is the actual fix: five `_reset`s call `self.observer.reset()`, the observation classes declare it as a no-op, and two guards keep it that way. Suite 3650 passed, 0 xfailed.

BY571:
@CharlieHelps round 3 at `5dcead5` — three rounds complete. Flagging that this PR has had three @mentions and a formal review request without a review landing, in case the trigger is not firing on it; the earlier PRs picked up within the hour.

Round 3's findings were smaller than the first two, but one was my guard claiming more than it enforced:

**`ast.walk` accepts unreachable calls.** The guard's docstring said the body "really CALLS" `self.observer.reset()`. It correctly rejected a comment, a docstring and a string literal — but accepted `if False: self.observer.reset()`, a nested `def` that never runs, and a statement after `return`. It scans top-level statements now, in the three shapes a call takes there (bare, returned, assigned — the SLTP envs assign their `super()._reset(...)`), mutation-checked against both unreachable forms.

**One test was misfiled.** `test_a_live_account_keeps_its_run_level_bankruptcy_baseline` passes on main *and* with the fix mutated out — so it is a guard against re-introducing round 1's defect rather than a test of this PR. Worth keeping (round 1 made exactly that mistake and it deletes the cross-episode drawdown circuit breaker) but it exercises `AlpacaTorchTradingEnv` with mock observers and has nothing to do with replay. Moved to `test_live_env_base.py`.

Prose trimmed to match the fix: **production is +37/−2, of which 8 lines are executable.** The 3-line ordering comment is stated once and pointed at from the other four sites.

Two corrections carried from the review, listed rather than quietly fixed: this file has **five** older AST guards, not four (my round 2 message said four), and `fb10dfc`'s "3648 passed" was the previous commit's number carried forward — it was 3650. That is the third factual error in this PR's commit messages, which is a pattern worth me owning rather than explaining away.

One thing for anyone reading a red suite: `tests/actor/test_frontier_llm_actor.py::test_generate_batch_preserves_order` is a **pre-existing flake** — fails 2 of 3 full runs on main, passes in isolation. Not this branch.

CI green, suite 3650 passed, 0 xfailed.

PR #376, opened 2026-08-18
fix(replay): the observation was a lookalike, not the one offline emits

BY571:
Third slice of #278, after #374 and #375. Three ways the replay observation differed from the offline and live ones it exists to be compared against.

## The market-data key had no window suffix

The sampler's raw keys are bare timeframe names; offline composes `market_data_{timeframe}_{window}`. So replay emitted `market_data_1Minute` where offline and live both emit `market_data_1Minute_10` — **a policy trained offline could not be fed the replay env without rewriting its in_keys**, which defeats the point of replaying through the live pipeline.

Parity is asserted against a real `SequentialTradingEnv`, not a hardcoded string, so the two cannot drift apart silently.

## `base_features` was 90% zeros

Only `base_arr[-1]` was written, on the reasoning that live envs read `[-1, 3]`. But `include_base_features=True` declares a `(window, 4)` spec and puts the array in the **observation** — so a policy consuming that key read 9 of 10 rows as zero, on every step, forever.

It fills from the sampler's execution bars now. Rows before the start of the data stay zero deliberately: there is no bar to report, and inventing one is worse than an obvious gap.

## Features reported as zero without a preprocessing fn

`get_features()` filtered to names starting with `features_`, so with no `feature_preprocessing_fn` it returned an empty list while the sampler still emitted the five raw OHLCV columns. The declared spec came out `(window, 0)` against an emitted `(window, 5)` and `check_env_specs` failed on a size mismatch. Only that one config was affected, which is why it survived.

## Verification

All three mutation-checked — reverting each fix fails its test (2, 1 and 2 failures respectively). Full suite **3655 passed**, 0 xfailed.

## Remaining in #278

`reduce_only=True` discarding `quantity` (forces a full close). It has **zero call sites** today, so it is latent interface divergence rather than a live defect — worth fixing when something needs a partial reduce.

### DIFF
--- a/tests/envs/replay/test_observation_parity_278.py
+++ b/tests/envs/replay/test_observation_parity_278.py
@@ -6,6 +6,7 @@ import pytest
 
 from torchtrade.envs.offline import SequentialTradingEnv, SequentialTradingEnvConfig
 from torchtrade.envs.replay.observer import ReplayObserver
+from torchtrade.envs.replay.order_executor import ReplayOrderExecutor
 from torchtrade.envs.utils import TimeFrame, TimeFrameUnit
 
 TF = TimeFrame(1, TimeFrameUnit.Minute)
@@ -105,15 +106,28 @@ def test_base_features_is_the_first_timeframes_bars(time_frames, window_sizes, e
     count, so neither a zero-padding regime nor a wrong-timeframe window can be written
     down as expected.
     """
+    executor = ReplayOrderExecutor(initial_balance=10000.0, leverage=5)
     observer = ReplayObserver(df=_df(1200), time_frames=time_frames,
-                              window_sizes=window_sizes, execute_on=execute_on)
+                              window_sizes=window_sizes, execute_on=execute_on,
+                              executor=executor)
     for _ in range(steps):
         obs = observer.get_observations(return_base_ohlc=True)
     base, market = obs["base_features"], obs[observer.get_keys()[0]]
 
     assert base.shape == (window_sizes[0], 4)
     assert int((base == 0).all(axis=1).sum()) == int((market == 0).all(axis=1).sum())
-    assert np.allclose(base[:, 3], market[:, 3], atol=1e-3), (
-        f"base_features closes {base[:, 3]} are not {time_frames[0]}'s bars "
-        f"{market[:, 3]} -- wrong timeframe or shifted by the end-time relabel"
+    # All but the last row are completed tf0 bars and must match exactly.
+    assert np.allclose(base[:-1, 3], market[:-1, 3], atol=1e-3), (
+        f"base_features closes {base[:-1, 3]} are not {time_frames[0]}'s bars "
+        f"{market[:-1, 3]} -- wrong timeframe or shifted by the end-time relabel"
+    )
+    # The last row is the FORMING bar, as live's is: its close is the latest trade.
+    # `base_features[-1, 3]` prices entries and BOTH brackets in three SLTP envs
+    # (`*/env_sltp.py: current_price = float(obs["base_features"][-1, 3])`), so leaving
+    # it at the last CLOSED tf0 close sizes and brackets off a stale number -- measured
+    # at -1.94%/+3.07% against a configured -2%/+3%. Asserted against the executor,
+    # which is what the venue would actually fill at.
+    assert base[-1, 3] == pytest.approx(executor.current_price, abs=1e-3), (
+        f"base_features[-1, 3] is {base[-1, 3]}, not the execution price "
+        f"{executor.current_price} -- entries and brackets would price off a stale bar"
     )
--- a/tests/envs/replay/test_observation_parity_278.py
+++ b/tests/envs/replay/test_observation_parity_278.py
@@ -79,15 +79,21 @@ def test_features_are_reported_when_no_preprocessing_fn_is_given():
     )
 
 
-# The timeframe layout is the axis both base_features bugs lived on, and the original
-# test parametrized only `steps` on a single-timeframe fixture -- so it could not see
-# either. execute_on COARSER than time_frames[0] is deliberately absent: market_data
-# there extends past the window this builds, and asserting a relationship I have not
-# pinned down would be worse than saying so.
+# The timeframe layout is the axis every base_features bug in this PR lived on, and the
+# first version parametrized only `steps` on a single-timeframe fixture where
+# `execute_on == time_frames[0]` -- the one layout on which none of them can appear.
+# Every replay test in the repo uses that layout, which is why they survived.
+M5 = TimeFrame(5, TimeFrameUnit.Minute)
+M15 = TimeFrame(15, TimeFrameUnit.Minute)
 LAYOUTS = [
     pytest.param([TF], [10], TF, id="single-1Min"),
-    pytest.param([TimeFrame(5, TimeFrameUnit.Minute), TF], [6, 10], TF, id="5Min-first"),
-    pytest.param([TimeFrame(15, TimeFrameUnit.Minute), TF], [4, 10], TF, id="15Min-first"),
+    pytest.param([M5, TF], [6, 10], TF, id="5Min-first"),
+    pytest.param([M15, TF], [4, 10], TF, id="15Min-first"),
+    # execute_on COARSER than time_frames[0]: tf0 is labelled by bar START, so looking it
+    # up at the raw execution label returned a window shifted back by
+    # `exec_period/tf0_period - 1` bars. Stale, deterministically.
+    pytest.param([TF], [10], M5, id="exec-coarser"),
+    pytest.param([TF, M5], [10, 6], M5, id="1Min-first-exec-5Min"),
 ]
 
 
--- a/tests/envs/replay/test_observation_parity_278.py
+++ b/tests/envs/replay/test_observation_parity_278.py
@@ -78,28 +78,42 @@ def test_features_are_reported_when_no_preprocessing_fn_is_given():
     )
 
 
-@pytest.mark.parametrize("steps", [1, 2, 10, 14])
-def test_base_features_is_as_full_as_the_market_data_from_the_first_step(steps):
-    """`include_base_features=True` declares a (window, 4) spec and puts it in the
-    OBSERVATION, so a policy consuming that key read 9 of 10 rows as zero on every step.
-    Only the newest row was written.
+# The timeframe layout is the axis both base_features bugs lived on, and the original
+# test parametrized only `steps` on a single-timeframe fixture -- so it could not see
+# either. execute_on COARSER than time_frames[0] is deliberately absent: market_data
+# there extends past the window this builds, and asserting a relationship I have not
+# pinned down would be worse than saying so.
+LAYOUTS = [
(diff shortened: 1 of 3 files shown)


BY571:
@CharlieHelps review requested — third slice of #278.

Note: #375 never picked up a review despite three @mentions and a formal request, so I have used the formal request here too. If the trigger is misbehaving, that PR merged without your sign-off (three agent rounds, CI green) and is worth a retrospective look.

The finding I would most like challenged here is the `base_features` one. The old code filled only the last row, with a comment justifying it: *"live envs only access [-1, 3] (close price)"*. That is true of the envs — but `include_base_features=True` declares a `(window, 4)` spec and puts the array in the **observation**, so a policy consuming that key reads 9 of 10 rows as zero. My reading is that the comment described the envs' access pattern and quietly became a justification for the data being wrong. If you think the original was deliberate and I have broken an intended contract, say so.

Two smaller ones: the market-data key lacked the `_{window}` suffix that offline and live both append (so a policy trained offline could not be fed the replay env without rewriting its in_keys), and `get_features()` filtered to `features_`-prefixed names, reporting **zero** columns when no preprocessing fn was given while the sampler still emitted five — a `(window, 0)` spec against a `(window, 5)` emission.

Key parity is asserted against a real `SequentialTradingEnv` rather than a hardcoded string, so the two cannot drift apart silently. All three are mutation-checked.

Suite 3655 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 applied at `99ef4df`. Both of this PR's main claims were false as shipped, so the earlier description is not to be trusted.

**The prefix went on at the wrong layer.** The env bases build their spec key as `"market_data_" + observer.get_keys()`, and the real observers return the **unprefixed** `{timeframe}_{window}`. So the sampler's bare timeframe names gave the env `market_data_1Minute`, and adding the prefix in the observer to compensate gave it **twice**: `market_data_market_data_1Minute_10`.

My test could not see either version, because it compared the *observer's* keys against the *offline env's* keys — two different layers. Both the missing suffix and the doubled prefix satisfied it. It now builds a real replay-backed env and compares spec keys with an offline env, which is the only comparison that settles the contract; mutation-checked in both directions.

**And `base_features` was still 9/10 zeros on step 1** — while `market_data_*`, same timeframe and same window, was full of real bars from step 1. `execute_base_features_df` is truncated at `min_start_time`, so a window ending at the first tradable bar had nothing to look back at. Those bars exist; they were outside the frame. The sampler now also keeps the untruncated frame (additive, nothing else reads it).

Worth flagging: **my test had codified the padding as intended** — a `(1, 9)` case in a parametrize, with a docstring explaining why nine zero rows were correct. It now compares `base_features` against `market_data` rather than a hardcoded count, so no zero-padding regime can be written down as expected again.

**The fallback I claimed to have fixed was still there.** `get_features()` had an `except ValueError: feature_keys = []` beside the filtering I removed, producing the identical `(window, 0)`-vs-`(window, 2)` mismatch for per-timeframe feature functions — same function, sibling branch, untouched. It raises now: that is boundary validation, and declaring a zero-width spec is the fail-open version of it.

On your earlier point about `base_features` being deliberate — round 1 settled it against you rather than me: the old comment ("live envs only access `[-1, 3]`") is accurate about the envs, but the array is in the observation under a `(window, 4)` spec, and `market_data_*` proves full history was always available at that moment.

Suite 3656 passed, 0 xfailed. Round 2 running.

BY571:
@CharlieHelps you reviewed `a12068b`; both blockers are fixed at `99ef4df`, and you and my round 1 found them independently and identically — which is a useful signal that they were the real defects rather than review noise.

**The doubled prefix.** Exactly as you describe: the env bases prepend `market_data_` to `observer.get_keys()`, and the real observers return the unprefixed `{timeframe}_{window}`, so putting the full prefix in `ReplayObserver` crossed that boundary. Keys are unprefixed now, suffix only. The test that let it through compared the *observer's* keys to the *offline env's* keys — two different layers, which is precisely how both the missing suffix and the doubled prefix passed. It builds a real replay-backed env and compares spec keys with an offline env now, mutation-checked in both directions.

**The lost pre-`min_start_time` history.** Also fixed: `execute_base_features_df` is truncated at `min_start_time`, so a window ending at the first tradable bar had nothing to look back at even though `market_data_*` — same timeframe, same window — was full of real bars at that moment. The sampler keeps the untruncated frame alongside it now (additive; nothing else reads it), and `base_features` is as full as `market_data` from step 1.

My own test had **codified the padding as intended** — a `(1, 9)` case with a docstring explaining why nine zero rows were correct. It compares `base_features` against `market_data` rather than a hardcoded count now, so that regime cannot be written down as expected again.

Thank you for settling the `base_features` question directly rather than deferring to my framing — "policies are entitled to the declared `(window, 4)` observation, not just the row the env happens to read internally" is a cleaner statement of it than mine, and it is the reason the old justifying comment was wrong even though it was accurate about the envs.

One thing your review did not reach, found in round 1: `get_features()` also had an `except ValueError: feature_keys = []` beside the filtering, producing the identical `(window, 0)`-vs-`(window, 2)` mismatch for per-timeframe feature functions. Same function, sibling branch. It raises now.

Suite 3656 passed. Round 2 is running; I will report before merging.

BY571:
@CharlieHelps round 2 at `12a80ab`. Filling the `base_features` window — the fix you and round 1 both asked for — made the defect **worse** before it made it better, in two ways.

**It was filled from `execute_on`, not `time_frames[0]`.** Live gates on `timeframe == self.time_frames[0]` and every venue sizes the spec off `window_sizes[0]`, so the window had the right *shape* and entirely different bars. With `time_frames=[15Min, 1Min]` a replayed window spanned 4 minutes where live spans an hour. Before this PR only row `[-1]` was mistimeframed and the rest were visibly zero — a broken key you could see. Round 1 filled them all from the wrong frame, so it looks plausible and `check_env_specs` passes.

**And my first attempt at the fix introduced lookahead bias.** I captured the first timeframe's OHLC *before* the end-time relabel — coarse bars are indexed by their bin START until #320's relabel moves them to bin END, so a window ending "now" contained a bar that had not closed yet. In the observation. Caught only by comparing closes against `market_data_*` for the same timeframe, which were off by exactly one bar.

**Round 1's sampler fields are deleted.** `execute_base_full_df` was written and never read, and the two full tensors duplicated rows the truncated ones already held — **+26.6 MB per sampler** on a 1M-bar frame, charged to every offline env and multiplied per `ParallelEnv` worker, for a field only replay used.

The root cause of my missing all of this: **the test was parametrized on the wrong axis.** It varied `steps` over a single-timeframe fixture where `execute_on == time_frames[0]` — the one layout on which both bugs are invisible. It parametrizes the timeframe layout now.

Two things stated rather than papered over: `execute_on` **coarser** than `time_frames[0]` is not asserted, because `market_data` there extends past the window this builds and I have not pinned down the correct alignment. And **offline never emits `base_features` at all** — `include_base_features` is declared and never read in `offline/sequential.py` — so the parity reference for this key is the live observers, not offline, contrary to what my earlier messages in this PR claimed.

Suite 3658 passed. Round 3 running.

BY571:
@CharlieHelps round 3 at `3c9cfa5`, and it found a **money-path regression I introduced in round 2** — worth your eye before this merges.

`base_features[-1, 3]` is not just an observation row. Three SLTP envs read it as `current_price` (`binance/env_sltp.py:286`, `bitget/env_sltp.py:290`, `alpaca/env_sltp.py:268`), unconditionally — not gated on `include_base_features` — and it divides the notional sizing **and** prices both brackets.

Round 2 moved the window to `time_frames[0]`, correct for the other rows and wrong for that one: the last row became the last **completed** tf0 bar instead of the execution bar. Over 40 steps on `[15Min, 1Min]`: **0/40** matched the execution price against main's 40/40, max error 0.444 on a ~100 price. End to end through a real `BitgetFuturesSLTPTorchTradingEnv`, brackets landed at **−1.94%/+3.07%** of the entry against a configured −2%/+3%.

It also diverged from live, which this PR names as its parity reference: live's last row is the **forming** candle, whose close is the latest trade. So the last row is the forming bar now — close set to the current price, high and low extended to cover it — with the earlier rows staying completed tf0 bars. 40/40 on both layouts that were 0/40, mutation-checked.

The reason it survived two rounds: **every replay test in the repo uses `time_frames=["1m"], execute_on="1m"`**, where tf0 and the execution bar coincide and the bug cannot appear. That is the identical blind spot my own round 2 message diagnosed for round 1, one layer down — I named the pattern and then reproduced it.

Also this round: `base_ohlc_df` was retained on the sampler for a single read in `__init__`, costing 40 MB per sampler on a 1M-bar frame — so round 2's replacement cost **2.0×** what it deleted for costing too much (+56.0 MB vs +27.9). It is a local now.

One characterised gap: `execute_on` **coarser** than `time_frames[0]` yields a window shifted back by `exec_period/tf0_period − 1` bars — stale, not lookahead, deterministic. The reviewer's view is that reusing `sampler._search_ts` fixes it in one line rather than warranting a raise. I have not done that in this PR; say if you want it here.

CI green, suite 3658 passed, 0 xfailed.

BY571:
@CharlieHelps final state at `c1f7207`. Your two blockers were fixed at `99ef4df`, and three review rounds since then found four more defects — three of them mine, introduced while fixing yours. Summary so you can review the end state rather than the sequence:

**Your blockers.** Doubled prefix: fixed, keys unprefixed, and the test now compares a real replay-backed *env*'s spec keys against an offline env's (it previously compared observer keys to env keys, which is how both the missing suffix and the doubled prefix passed). Lost pre-`min_start_time` history: fixed, `base_features` is as full as `market_data` from step 1.

**What the rounds found in my fixes:**

1. **Wrong timeframe** — I filled the window from `execute_on`; live gates on `time_frames[0]`. Right shape, entirely different bars: `[15Min, 1Min]` spanned 4 minutes where live spans an hour. Worse than the original, which left mismatched rows visibly zero.
2. **Lookahead bias** — my first attempt captured the OHLC before #320's end-time relabel, putting a bar that had not closed into the observation.
3. **A money-path regression** — `base_features[-1, 3]` is `current_price` in three SLTP envs, dividing the notional sizing and pricing both brackets. Moving the window to tf0 made the last row the last *completed* bar: **0/40** matching the execution price against main's 40/40, and end-to-end brackets at −1.94%/+3.07% against a configured −2%/+3%. The last row is the forming bar now, as live's is.
4. **The gap I said I would leave documented, closed instead** — `execute_on` coarser than `time_frames[0]` returned a window shifted back by `exec_period/tf0_period − 1` bars. `sampler._search_ts` exists for precisely that rule and I had re-implemented the easy half. Both coarse-execute layouts go 0/30 → 30/30.

**The single cause of all four surviving review:** every replay test in the repo uses `time_frames=["1m"], execute_on="1m"` — the one layout on which none of them can appear. The test parametrizes five layouts now.

Memory: round 1 added +27.9 MB/sampler, round 2's replacement cost +56.0 MB (I deleted a field for costing too much and replaced it with one costing twice as much), round 3 dropped the retained frame.

CI green, suite 3662 passed, 0 xfailed. I would value your re-review, but if the trigger is not firing on this PR — as it did not on #375 — I will merge on green rather than block, and this comment is the record of what to look at retrospectively.

PR #377, opened 2026-08-18
fix(replay): reduce_only reduces BY a quantity, it does not always flatten

### LINKED ISSUE #278: Replay fidelity: reward lags one bar; fractional+fee rejects 100% of trades; reset never rewinds; obs keys don't match live/offline

BY571:
The last item in #278.

`if reduce_only: return self.close_position()` discarded the quantity, so `trade("SELL", 0.25, reduce_only=True)` on a 1.0 position closed **all** of it. All four live executors forward the quantity with a `reduceOnly` flag and honour a partial, so replay silently disagreed with every one of them.

**Latent rather than live:** `grep -rn "reduce_only" torchtrade/ --include=*.py` outside the executors returns nothing, so no env passes it today. It is interface divergence in the one component whose job is to behave like the venues.

## What changed

`_close_at_price` takes an optional quantity and books **pro rata** — PnL, fee and released margin all scale with the closed portion.

Two things deliberately do *not* change on a partial:
- **the entry price**, because the remainder was opened at it and its cost basis does not move because part of it was sold;
- **the brackets**, because a partially reduced position still has one. Clearing them would leave a live position with no stop.

An oversized reduce clamps to the position rather than flipping the side; a reduce on a flat account returns `False` rather than booking anything.

## Verification

Pro-rata is asserted by comparing a 0.25 reduce against the remaining 0.75 at the same moved price — three times the credit, rather than a recomputed number, so the test cannot agree with the implementation by sharing its arithmetic. Mutation-checked: restoring the full close fails 5 tests.

Full suite **3669 passed**, 0 xfailed.

Closes #278.

### DIFF
--- a/tests/envs/replay/test_reduce_only_278.py
+++ b/tests/envs/replay/test_reduce_only_278.py
@@ -5,10 +5,11 @@ import pytest
 from torchtrade.envs.replay.order_executor import ReplayOrderExecutor
 
 
-def _opened(qty=1.0, price=100.0, leverage=5):
-    executor = ReplayOrderExecutor(initial_balance=10000.0, leverage=leverage)
+def _opened(qty=1.0, price=100.0, leverage=5, fee=0.0, side="BUY"):
+    executor = ReplayOrderExecutor(initial_balance=10000.0, leverage=leverage,
+                                   transaction_fee=fee)
     executor.advance_bar({"open": price, "high": price + 1, "low": price - 1, "close": price})
-    executor.trade("BUY", qty)
+    executor.trade(side, qty)
     return executor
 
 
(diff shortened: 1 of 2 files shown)


BY571:
@CharlieHelps review requested — this closes #278.

`if reduce_only: return self.close_position()` discarded the quantity, so a 0.25 reduce on a 1.0 position closed all of it. All four live executors forward the quantity with a `reduceOnly` flag and honour a partial.

Two judgement calls I'd like checked rather than assumed:

**The entry price is unchanged on a partial.** My reasoning is that the remainder was opened at that price and its cost basis doesn't move because part of it was sold. That matches how the venues report `avgEntryPrice` after a partial reduce, but if any of them re-derive it differently I'd rather know.

**The brackets survive a partial and are cleared only on a full close.** A partially reduced position still has a stop, and clearing it would leave a live position unprotected. The opposite argument — that the bracket quantity no longer matches the position — is also coherent, and the venues differ on whether a reduceOnly fill resizes the attached orders. I picked the fail-safe side.

Pro-rata is asserted by comparing a 0.25 reduce against the remaining 0.75 at the same moved price — three times the credit, rather than a recomputed number, so the test can't agree with the implementation by sharing its arithmetic.

Latent, not live: no env passes `reduce_only` today, so this is interface divergence in the one component whose job is to behave like the venues.

Suite 3669 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `2bdaa70`. Both of the judgement calls I asked you about came back **correct**, but the review found I had fixed only half the divergence and written a test that could not fail.

**`side` was ignored entirely.** I fixed the *size* dimension of `reduce_only` and left the *sign* dimension diverging from the same four executors the change cites. Direction came from `copysign(..., position_qty)`, so `trade("BUY", 0.5, reduce_only=True)` on a **long** returned True and reduced it. On every venue `side` opposes the position — bybit routes a BUY reduceOnly to the short leg (`positionIdx = 2 if Buy else 1`), okx sets `posSide = "short" if buy else "long"`, binance and bybit both build a close side as the inverse of the held one — and in one-way mode the venue **rejects** a same-direction reduceOnly. It is rejected here now. Same defect class this branch exists to remove, one argument over.

**And my pro-rata test proved nothing.** It compared a 0.25 reduce against the remaining 0.75 and asserted only that the second was three times the first — pure linearity, which holds for *any* basis that scales with the closed quantity. Corrupting the fee basis to entry price **and** the margin basis to fill price left all seven tests passing. It also never set `transaction_fee`, which defaults to 0.0, so the word "fee" in its name was untested. I had described it in the PR body as designed so "the test cannot agree with the implementation by sharing its arithmetic" — it didn't share the arithmetic, it simply didn't check it.

It asserts the absolute ledger now (PnL on the closed portion, fee on the **fill** notional, margin released at **entry**) with a real fee, parametrized over long and short — every test in the file opened with BUY, and a short PnL sign is the classic failure here. All three mutations fail it.

On your side: both judgement calls confirmed against the source rather than assumed. Entry price — all four executors *read* it from the venue (`entryPrice`/`avgPrice`/`avgPx`) and never recompute, so a partial leaves it unchanged. Brackets — `sl_price`/`tp_price` are not quantity-scoped, and a later SL closes `position_qty`, i.e. the remainder, so there is no over-close; that matches venue-side `reduceOnly` capping a full-size stop at the actual remaining position.

Suite 3672 passed, 0 xfailed.

BY571:
@CharlieHelps round 2 at `8ec3d0c`. The headline is that **round 1's replacement test reproduced the defect it replaced.**

`test_a_partial_reduce_keeps_the_entry_price_and_the_brackets` never moved the bar, so `fill == entry` and the entry assertion was a tautology — re-basing the remainder to the fill on every partial passed all 10 tests in the file and all 75 in the package. That is round 1's own finding, in round 1's own replacement, one test over. It matters: `entry_price` feeds `liquidation_price`, the next reduce's margin release, and unrealized PnL.

Writing that fix immediately produced a **second** version of the same mistake — the bracket levels I set (tp=103) fire on the 110 advance, so the position was flat and the assertion ran against a closed position. Caught only because the test then failed loudly.

**The dust guard had zero coverage** — the one new import in the PR. Replacing `abs(qty) <= POSITION_DUST_EPS` with `== 0.0` passed every replay test, yet it is load-bearing: seven reduces of 1/7 against a 1.0 short leave `-2.2e-16`, and `get_status()` reports a phantom open short. That is CLAUDE.md invariant 1 exactly.

**And the same diff that imports `POSITION_DUST_EPS` hand-wrote `== 0` and `> 0`** for the two direction decisions in the new branch — the forms invariant 1 forbids. With a 1e-12 position, `trade("SELL", 0.5, reduce_only=True)` booked a trade against dust. Both go through the dust rule now.

Dead code removed: `"ASK"` was unreachable (`trade()` raises on anything but BUY/SELL 26 lines earlier), `if closing == 0: return` was unreachable, and `side.upper()` recomputed a `side_upper` already bound above. The 7-line comment is 3.

Two corrections to my round 1 message rather than leaving them: **"N partial closes leave exactly the balance of one full close" is false** — they differ by up to 5.46e-12 over a 60-case grid, immaterial but I stated exactness; and "All three mutations fail it" reads as though all three fail the pro-rata test, when the side mutation fails only the rejection tests.

CI green, suite 3673 passed, 0 xfailed. This closes #278 on merge.

PR #378, opened 2026-08-18
refactor(live): one unrealized-PnL helper instead of three

BY571:
First slice of #288 (step 4), deliberately the smallest — the issue's own history is of fixes landing on some exchanges and not others, so the dedup itself should arrive in reviewable pieces.

## What was duplicated

`_calculate_unrealized_pnl_pct` existed **three times** (bitget, bybit, okx) with identical bodies and differently sized docstrings. `_round_price` existed three times byte-identical (binance, bybit, okx). Both now live on an `ExecutorHelpersMixin`.

## What is deliberately NOT deduplicated

**bitget's `_round_price`** rounds through CCXT's `price_to_precision`; the others do tick-size arithmetic against a cached tick. Different mechanisms, different failure modes. Collapsing them to make the file shorter would be the same error as leaving four copies of one rule — there is a test pinning bitget's version so a later pass doesn't "finish the job".

## A defect the copies shared

All three branched on `qty > 0`, so a **1e-12 residual** left by a full close took the SHORT branch and reported **−10%** on a position that does not exist — into `account_state[2]`, which the policy reads every step and a reward function sees. That is CLAUDE.md invariant 1, in the observation. The shared version goes through `POSITION_DUST_EPS` in both directions.

## Guard

Per the issue's plan ("each step extends the structural-guard tests so re-forking fails CI"): a test asserts no executor defines its own copy and that each resolves to the mixin's. A copy that hasn't drifted *yet* is still a copy. Mutation-checked — re-forking okx fails 1 test, reverting the dust rule fails 3.

Net **−35 lines**. Full suite **3689 passed**, 0 xfailed.

## Remaining in #288

Steps 1–3 and 5: folding the four `base.py` files into `TorchTradeFuturesLiveEnv` (~600-700 LOC), the `env.py` trade-path helpers (~500), `env_sltp.py` including the 70-line config repeated 4× (~900), and routing per-exchange test files through the shared harness.

### DIFF
--- a/tests/envs/test_executor_helpers_288.py
+++ b/tests/envs/test_executor_helpers_288.py
@@ -33,10 +33,32 @@ def test_no_executor_re_forks_the_shared_pnl(cls):
                                  OKXFuturesOrderClass], ids=lambda c: c.__name__)
 def test_tick_rounding_goes_through_the_shared_helper(cls):
     """These three had byte-identical tick arithmetic. bitget is deliberately absent --
-    it rounds through CCXT's `price_to_precision`, a different mechanism with a different
+    it OVERRIDES with CCXT's `price_to_precision`, a different mechanism with a different
     failure mode, and collapsing it in to shorten the file would be the opposite error.
     """
-    assert "_round_price_by_tick" in inspect.getsource(cls._round_price)
+    assert cls._round_price is ExecutorHelpersMixin._round_price
+
+
+@pytest.mark.parametrize("cls", EXECUTORS, ids=lambda c: c.__name__)
+def test_no_executor_re_implements_the_pnl_rule_INLINE(cls):
+    """By RULE, not by name -- the name check alone certified the one that was wrong.
+
+    Binance never had a method called `_calculate_unrealized_pnl_pct`; it had the same
+    arithmetic inlined in `get_status`, still branching on `qty > 0`. So
+    `"_calculate_unrealized_pnl_pct" not in cls.__dict__` passed VACUOUSLY on the single
+    exchange still carrying the dust bug, and the dust parametrization below calls the
+    mixin directly and never touches an executor. The guard certified the miss.
+
+    A re-inlined copy is spelled `(mark_price - entry_price) / entry_price` or its short
+    counterpart; anything computing that outside the mixin is a fourth copy.
+    """
+    source = inspect.getsource(cls)
+    for expression in ("(mark_price - entry_price) / entry_price",
+                       "(entry_price - mark_price) / entry_price"):
+        assert expression not in source, (
+            f"{cls.__name__} recomputes the PnL rule inline instead of calling the "
+            f"shared helper -- a copy that is not a method is still a copy"
+        )
 
 
 def test_bitget_keeps_its_ccxt_rounding():
--- a/tests/envs/test_executor_helpers_288.py
+++ b/tests/envs/test_executor_helpers_288.py
@@ -1,0 +1,63 @@
+"""One copy of the shared executor helpers, and a guard against re-forking (#288)."""
+
+import inspect
+
+import pytest
+
+from torchtrade.envs.live.binance.order_executor import BinanceFuturesOrderClass
+from torchtrade.envs.live.bitget.order_executor import BitgetFuturesOrderClass
+from torchtrade.envs.live.bybit.order_executor import BybitFuturesOrderClass
+from torchtrade.envs.live.okx.order_executor import OKXFuturesOrderClass
+from torchtrade.envs.live.shared.executor_helpers import ExecutorHelpersMixin
+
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps review requested — first slice of #288, step 4, deliberately the smallest.

**The interesting part is what I did NOT deduplicate.** bitget's `_round_price` goes through CCXT's `price_to_precision`; binance/bybit/okx do tick-size arithmetic against a cached tick. Those are different mechanisms with different failure modes, so I left bitget alone and added a test pinning it, on the reasoning that folding it in to shorten the file is the same error as leaving four copies of one rule. If you think that is over-cautious and it should be one path, say so — the opposite argument (that a single rounding contract is the point of the dedup) is coherent.

**Deduping surfaced a real defect the three copies shared:** they branched on `qty > 0`, so a 1e-12 residual from a full close took the SHORT branch and reported **−10%** on a position that does not exist — into `account_state[2]`, which the policy reads every step and a reward function sees. That is invariant 1 in the observation. The shared version uses `POSITION_DUST_EPS` in both directions.

Worth a sceptical look: **that is a behaviour change riding inside a refactor PR.** I judged it in-scope because it is the exact defect class #288 exists to prevent and it would be perverse to canonicalise the bug into a shared helper — but it does mean this PR is not purely mechanical, and you may reasonably want it split.

Guard added per the issue's plan: a test asserts no executor defines its own copy and each resolves to the mixin's. Mutation-checked both ways.

Net −35 lines. Suite 3689 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `e229f04`. The review found this PR committing the exact defect it exists to prevent, which is worth stating plainly.

**There were four copies of the PnL rule, not three.** Binance never had a *method* called `_calculate_unrealized_pnl_pct` — it had the same arithmetic **inlined in `get_status`**, still branching on `qty > 0`. Measured against the real `get_status`: `qty=-1e-12 → pnl_pct=-0.1`, verbatim the behaviour this PR claimed to have removed. So I deduplicated three copies and left the fourth, which is #288's entire thesis.

**And my structural guard certified it clean.** `"_calculate_unrealized_pnl_pct" not in cls.__dict__` is a check by *name*, and binance's copy had no name; the dust parametrization called the mixin directly and never touched an executor. The one exchange still carrying the bug was the one the test passed. It checks for the **rule** now — the arithmetic appearing anywhere in the class — and the mutation that re-inlines it fails.

**I also overclaimed the justification, in the source file.** I wrote that dust PnL "puts a phantom position's number in the observation". It does not: `futures_live_base._get_account_state` already hard-sets it to 0.0 when the direction is dust, and there is an existing test asserting exactly that. The change is defence in depth — worth doing, because canonicalising `qty > 0` into a *shared* helper hands the next caller the bug pre-approved — but the docstring said something untrue and now says what is actually the case.

On your open question about the bitget carve-out, the review settled it empirically rather than by argument: **600,000 realistic bracket prices through CCXT `price_to_precision` vs the tick arithmetic diverge zero times.** They differ only on exact half-tick ties (banker's vs half-up) and sub-tick prices, neither reachable from computed bracket prices. So the carve-out pins no latent divergence. The mixin's method is named `_round_price` directly now and bitget *overrides* it, which deletes the three one-line forwarders and makes the divergence explicit rather than incidental.

Suite 3693 passed, 0 xfailed.

BY571:
@CharlieHelps thank you — and your non-blocking follow-up is already done at `e229f04`, which landed after the `7936cb3` you reviewed.

You and the review agent found the binance inline copy independently and identically. It now routes through the shared helper, and the structural guard checks for the **rule** rather than the name — the name check passed vacuously on binance precisely because its copy had no name, so the one exchange still carrying the bug was the one my test certified clean. Mutation-checked: re-inlining the arithmetic fails it.

Your two judgement calls both confirmed empirically rather than by argument:

- **The bitget split** — 600,000 realistic bracket prices through CCXT `price_to_precision` vs the tick arithmetic diverge **zero** times. They differ only on exact half-tick ties (banker's vs half-up) and sub-tick prices, neither reachable from a computed bracket. So keeping it separate pins no latent divergence, and bitget now *overrides* the shared method rather than sitting beside it.
- **The dust change in a refactor slice** — I've also corrected the justification I wrote in the source. I had claimed it reached the observation; it does not, exactly as you say: `_get_account_state` normalises to flat before `account_state[2]`. The docstring now says it is defence in depth, the real argument being that canonicalising `qty > 0` into a *shared* helper hands the next caller the bug pre-approved.

On your suggestion of exact `±POSITION_DUST_EPS` / just-outside cases: the review's grid covered 1e-12, 9.99e-10, 1e-9, 1.0000001e-9, 2e-9 and 1e-8 against main and confirmed the boundary is exactly equivalent to `position_direction_from_status`. I'll fold explicit boundary cases into the next slice's tests rather than re-open this one.

Suite 3693 passed, 0 xfailed. Merging.

PR #379, opened 2026-08-18
refactor(live): one wait duration instead of a 17-entry spelling table

BY571:
Second slice of #288 — the ceremony cleanups.

## The alias table

`_wait_for_next_timestamp` looked its time unit up in a **17-entry** map: `"TimeFrameUnit.Minute"`, `"Minute"`, `"Min"`, `"min"`, `"minute"`, `"h"`, `"H"`, `"D"`, `"d"`, `"seconds"`, `"hour"`, `"day"`…

The reason is that **five exchanges stringify the same enum four different ways**:

| exchange | assignment |
|---|---|
| bybit / okx / bitget | `str(config.execute_on.unit)` → `"TimeFrameUnit.Minute"` |
| alpaca | `str(config.execute_on.unit.value)` → `"min"` |
| binance | hardcoded `"seconds"` |

The table grew an entry per spelling instead of the spellings being fixed. A sixth exchange spelling it a fifth way would have raised **at the first bar** — in production, on a timer, after the env had already opened positions.

It derives from the TimeFrame now: `timedelta(seconds=timeframe_to_seconds(execute_on))`. Verified identical for 1Min/5Min/1H/4H/1D.

Also: bitget's `isinstance(config.execute_on, TimeFrame)` branch had **byte-identical arms**, one commented "should not happen after normalization" — three lines of ceremony for one assignment.

## The guard

It reads **literals out of the AST**, not the source text. My first version grepped the source and failed on its own explanatory comment, which quotes the old spellings — a check a comment can trip is a check a comment can also satisfy, and this PR series has been bitten by that in both directions. Mutation-checked: reintroducing a two-entry alias table fails it.

Full suite **3694 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -2724,3 +2724,32 @@ def test_a_live_account_keeps_its_run_level_bankruptcy_baseline():
     assert env._check_termination(62.5) is True, (
         "62.5 is below 10% of the 1000 this run started with and must terminate"
     )
+
+
+def test_the_bar_wait_derives_from_the_timeframe_not_a_string_alias():
+    """One duration rule, not an alias table that grows per spelling (#288).
+
+    `_wait_for_next_timestamp` looked its unit up in a 17-entry map --
+    "TimeFrameUnit.Minute", "Minute", "Min", "min", "minute", "h", "H", "D", "d",
+    "seconds" -- because five exchanges stringified the same enum four different ways
+    and the map grew an entry per spelling rather than the spellings being fixed. A
+    sixth exchange spelling it a fifth way would have raised at the first bar, in
+    production, on a timer.
+    """
+    from torchtrade.envs.core import live as live_module
+
+    tree = ast.parse(textwrap.dedent(
+        inspect.getsource(live_module.TorchTradeLiveEnv._wait_for_next_timestamp)
+    ))
+    called = {n.func.id for n in ast.walk(tree)
+              if isinstance(n, ast.Call) and isinstance(n.func, ast.Name)}
+    assert "timeframe_to_seconds" in called, "the wait must derive from the TimeFrame"
+
+    # Literals only -- the explanation above quotes the old spellings, and a check that
+    # a comment can trip is a check a comment can also satisfy.
+    literals = {n.value for n in ast.walk(tree)
+                if isinstance(n, ast.Constant) and isinstance(n.value, str)}
+    regrown = literals & {"TimeFrameUnit.Minute", "Minute", "Min", "min", "minute",
+                          "TimeFrameUnit.Hour", "Hour", "h", "H", "seconds",
+                          "TimeFrameUnit.Day", "Day", "D", "d", "hour", "day"}
+    assert not regrown, f"the unit alias table is regrowing in the wait path: {regrown}"
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -2726,6 +2726,62 @@ def test_a_live_account_keeps_its_run_level_bankruptcy_baseline():
     )
 
 
+@pytest.mark.parametrize("cls", LIVE_ENVS, ids=lambda c: c.__name__)
+def test_every_live_env_can_actually_run_the_bar_wait(cls):
+    """RUN it, do not inspect it. This is the test that was missing.
+
+    `_wait_for_next_timestamp` reads `self.execute_on`, and alpaca's base set only
+    `execute_on_value`/`execute_on_unit` -- so both alpaca envs raised AttributeError at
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps review requested — second slice of #288.

A 17-entry alias table existed because **five exchanges stringify the same enum four different ways** (`str(unit)`, `str(unit.value)`, a hardcoded `"seconds"`). The table grew an entry per spelling rather than the spellings being fixed. A sixth exchange spelling it a fifth way would have raised **at the first bar** — in production, on a timer, after positions were already open.

The bit I'd like checked hardest: **this changes live timing behaviour**, which no test can really exercise. The old path built `timedelta(minutes=execute_on_value)` from a parsed unit; the new one is `timedelta(seconds=timeframe_to_seconds(execute_on))`. I verified they agree for 1Min/5Min/1H/4H/1D, but binance previously stored `execute_on_value` in *seconds* with unit `"seconds"` while the others stored value+unit — so binance was already computing a different quantity through the same code path. If you can see a config where the two disagree, that's a real regression and I'd rather hear it now.

I've left `execute_on_unit` and `execute_on_value` in place rather than deleting them: `offline/infrastructure/utils.py` consumes the offline equivalents and I did not want a rename crossing that boundary in a slice about timing. That means the divergent spellings still exist, just no longer load-bearing — arguably worth finishing in the base.py slice.

One process note: my first version of the guard grepped source text and failed on **its own explanatory comment**, which quotes the old spellings. It reads literals out of the AST now. A check a comment can trip is a check a comment can also satisfy, and this series has been bitten in both directions.

Suite 3694 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `57a4052` — and the thing I asked you to look for was real. **This refactor shipped a production break.**

`_wait_for_next_timestamp` now reads `self.execute_on`. Four exchanges set it; **alpaca does not** — its base set only `execute_on_value`/`execute_on_unit`. So both alpaca envs raised `AttributeError` at the first bar wait, which happens **after** `_execute_trade_if_needed` and `_record_position_after_trade`: the order goes to the venue, the position is recorded, and then the step dies. 100% of alpaca live and paper trading, every timeframe. It worked on main because alpaca's `"min"`/`"h"`/`"D"` were in the alias table.

**834 exchange tests and 513 contract tests passed through it.** Every one of them stubs this method — `env._wait_for_next_timestamp = lambda: None` or `patch.object(...)`. Nothing in the repo had ever executed the body. That is the same fixture-monoculture failure as #376, in a different form: not one config held constant, but one *method* universally stubbed.

Two tests now exercise it: one asserts every exchange base assigns `self.execute_on`, one drives the real body with `time.sleep` patched across four timeframes.

Worth reporting: **the first version of that guard also failed.** It searched the whole MRO, and `TorchTradeLiveEnv` initialises `self.execute_on = None` — so it passed on a concrete class that never assigns it, i.e. exactly the alpaca case, certified clean. The mutation survived. It checks the *exchange* base now and the mutation fails 4.

With the wait fixed, `execute_on_unit`/`execute_on_value` have **zero readers repo-wide**, so the five assignments in four spellings are deleted — leaving them would preserve the exact hazard this slice removes.

The review confirmed the part that was right: all 72 futures cases (8 envs × 9 timeframes) produce byte-identical durations to main, binance included.

Suite 3710 passed, 0 xfailed.

BY571:
@CharlieHelps you and the review agent found this independently and identically — including the root cause, that CI misses it because the step tests replace `_wait_for_next_timestamp` with a lambda. Both of your requests are in `57a4052` (your review is at `722a147`).

**The Alpaca initialization:** its base now sets `self.execute_on = config.execute_on` — the canonical TimeFrame, as you suggested — replacing the two derived strings.

**The regression test that exercises the inherited wait path:** two, actually. One asserts every exchange base assigns `self.execute_on`; the other drives the real body with only `time.sleep` patched, across 1Min/5Min/4H/1D. Nothing in the repo had ever executed that body.

Worth flagging because it is the same failure one level up: **my first version of that guard also passed while Alpaca was broken.** It searched the whole MRO, and `TorchTradeLiveEnv` initialises `self.execute_on = None` — so a concrete class that never assigns it still matched. The mutation survived. It checks the *exchange* base now, and reverting Alpaca fails 4 tests.

With the wait fixed, `execute_on_unit`/`execute_on_value` have zero readers repo-wide, so the five assignments in four different spellings are deleted rather than left as decoration — leaving them would preserve exactly the hazard this slice removes.

Your confirmation on Binance is the one I most wanted, since it was the case I could not convince myself about: its `execute_on_value` was already in seconds, so old and new agree. The review's 72-case grid (8 futures envs × 9 timeframes) says the same.

Suite 3710 passed, 0 xfailed.

PR #380, opened 2026-08-18
refactor(live): one _format_price instead of two

BY571:
Third slice of #288, deliberately small.

`_format_price` was byte-identical on bybit and okx. It emits a **string** rather than a float because the venues parse the wire value and `repr` of a rounded float can carry more digits than the tick allows — which the venue then rejects or silently re-rounds. A drifted copy is a rejected order, not a cosmetic difference.

Guarded by identity against the mixin; mutation-checked (re-forking okx fails it).

## Deliberately not in this slice

`TorchTradeLiveEnv._get_portfolio_value` and `_step` are raise-only re-declarations of methods already abstract on `TorchTradeBaseEnv` and torchrl's `EnvBase`. Removing them changes `__abstractmethods__` and therefore which classes are instantiable — not something to bundle into a two-file dedup. The previous slice of this issue shipped an `AttributeError` into every Alpaca bar by making one shared method read one more attribute, so the bar for "small refactor" here is low.

`_format_size` exists only on okx, so there is nothing to share.

Full suite **3712 passed**, 0 xfailed.

## Remaining in #288

The three large ones: `base.py` consolidation (~600-700 LOC), `env.py` trade path (~500), `env_sltp.py` including the 70-line config repeated 4× (~900), plus routing per-exchange tests through the shared harness.

### DIFF
--- a/tests/envs/test_executor_helpers_288.py
+++ b/tests/envs/test_executor_helpers_288.py
@@ -8,7 +8,10 @@ from torchtrade.envs.live.binance.order_executor import BinanceFuturesOrderClass
 from torchtrade.envs.live.bitget.order_executor import BitgetFuturesOrderClass
 from torchtrade.envs.live.bybit.order_executor import BybitFuturesOrderClass
 from torchtrade.envs.live.okx.order_executor import OKXFuturesOrderClass
-from torchtrade.envs.live.shared.executor_helpers import ExecutorHelpersMixin
+from torchtrade.envs.live.shared.executor_helpers import (
+    ExecutorHelpersMixin,
+    TickSizeMixin,
+)
 
 EXECUTORS = [BinanceFuturesOrderClass, BitgetFuturesOrderClass,
              BybitFuturesOrderClass, OKXFuturesOrderClass]
@@ -36,7 +39,7 @@ def test_tick_rounding_goes_through_the_shared_helper(cls):
     it OVERRIDES with CCXT's `price_to_precision`, a different mechanism with a different
     failure mode, and collapsing it in to shorten the file would be the opposite error.
     """
-    assert cls._round_price is ExecutorHelpersMixin._round_price
+    assert cls._round_price is TickSizeMixin._round_price
 
 
 @pytest.mark.parametrize("cls", EXECUTORS, ids=lambda c: c.__name__)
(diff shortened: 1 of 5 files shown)


BY571:
@CharlieHelps review requested — third slice of #288, deliberately small.

`_format_price` was byte-identical on bybit and okx. Worth saying why it matters despite looking cosmetic: it returns a **string**, because the venues parse the wire value and `repr` of a rounded float can carry more digits than the tick permits — so a drifted copy is a rejected or silently re-rounded order.

**What I chose not to do, and would like your view on:** `TorchTradeLiveEnv._get_portfolio_value` and `_step` are raise-only re-declarations of methods already abstract on `TorchTradeBaseEnv` and torchrl's `EnvBase`. The issue lists them as ceremony worth removing, and mechanically they are — but removing them changes `__abstractmethods__`, and therefore which classes Python will let you instantiate. After the last slice shipped an `AttributeError` into every Alpaca bar by making one shared method read one more attribute, I did not want that riding along in a two-file dedup. If you think that is over-cautious for what is genuinely a no-op, say so and I will fold it in.

`_format_size` exists only on okx, so there is nothing to share there.

Suite 3712 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `8fabd44` — including a definite answer to the question I put to you, which went against me.

**You asked implicitly and the review answered explicitly: my deferral was unfounded.** `TorchTradeLiveEnv._get_portfolio_value` and `_step` are `@abstractmethod`, but both are *already* abstract on `TorchTradeBaseEnv` and torchrl's `EnvBase`, so `getattr` still finds them. Measured `__abstractmethods__` before and after removal — **identical set**, and all six concrete envs still have empty abstracts. Removed. I had reasoned from "this touches the class hierarchy" to "this is risky" without checking, which is the same shortcut that produced the Alpaca break one slice earlier, just in the cautious direction.

**Two real defects in the PR:**

The mixin handed bitget a `_format_price` that **always raises** — it needs `_tick_size`, which bitget has not got, and unlike `_round_price` there was no bitget override. My own docstring justified the unusable inheritance ("the override is what makes it work"), true of `_round_price` and false of the method I had just added. An override only covers the one you remember to write, so the tick-dependent helpers are now a separate `TickSizeMixin` that bitget does not mix in at all — not mixing in cannot be half-done.

And the guard was **defeatable by inlining**, which is the exact hole the name-based PnL guard had in the first slice, in a file carrying that lesson 26 lines above the new test. Identity catches a re-added *method* and is blind to `f"{rounded:.{self._tick_decimals}f}"` at a call site — mutation confirmed 20 passed with the drift in place. There is a rule-based check beside it now.

The dedup itself verified clean: 986-cell grid across tick sizes, decimals and prices (sub-tick, half-tick, 1e-300, 1.7e308, inf, nan) shows **0 differences** from main including exception behaviour, and an AST-identity scan over all 58 methods of the four executors finds no remaining duplicate pair.

Suite 3715 passed, 0 xfailed.

Issue #279, opened 2026-08-07
SLTP short actions carry the inverse risk/reward of the mirrored long action — intentional?


BY571:
## Observation

\`torchtrade/envs/utils/action_maps.py:78-82\`: shorts are built as \`("short", tp, sl)\` — the sign-swap is necessary, but the **magnitudes** are reused from the opposite list.

With \`stoploss_levels=(-0.025,-0.05,-0.1)\`, \`takeprofit_levels=(0.05,0.1,0.2)\`: long action k is (risk 2.5%, reward 5%), the mirrored short action is (risk 5%, reward 2.5%). There is **no short action with a 2.5% stop anywhere in the action space**. A policy that learned "tight stop, wide target" gets the opposite geometry when it goes short.

## Question

Is this deliberate (shorts inherently want inverted geometry?) or an artifact of the swap implementation? If unintentional, the fix is to mirror magnitudes: short SL from \`|stoploss_levels|\` above entry, short TP from \`|takeprofit_levels|\` below entry — which changes the action-space size contract for existing checkpoints, so it needs a decision, not just a patch.

BY571:
Measured it. **The asymmetry is real and larger than "inverted geometry" suggests** — it is not a mirror, it is a different and strictly worse action set.

`create_sltp_action_map([-0.025, -0.05, -0.1], [0.05, 0.1, 0.2], include_short_positions=True)`:

| | longs | shorts |
|---|---|---|
| stop magnitudes | 2.5%, 5%, 10% | **5%, 10%, 20%** |
| R:R available | 0.5 … **8.0** | 0.12 … **2.0** |

```
  idx  side   risk    reward   R:R          idx  side   risk     reward   R:R
    1  long   2.500%  5.000%   2.00          10  short  5.000%   2.500%   0.50
    2  long   2.500%  10.000%  4.00          11  short  10.000%  2.500%   0.25
    3  long   2.500%  20.000%  8.00          12  short  20.000%  2.500%   0.12
```

Three consequences worth separating:

1. **No short action has a 2.5% stop.** A policy that learned "tight stop" cannot express it short — the tightest available is 5%, double the risk.
2. **Shorts get a 20% stop that longs never see**, so the widest short risk is 2× the widest long risk.
[...]

PR #381, opened 2026-08-18
feat(sltp): say out loud that short actions are not the mirror of long ones

BY571:
Partial for #279 — the part that needs no decision.

#279 asks a design question (*is the inverted short geometry intentional?*) and says its own fix "needs a decision, not just a patch". That is right, and **this is not that patch.**

## The asymmetry

The sign swap in `create_sltp_action_map` is necessary — a short's stop sits *above* entry — but it reuses the opposite list's **magnitudes**. With `stoploss_levels=(-0.025,-0.05,-0.1)` and `takeprofit_levels=(0.05,0.1,0.2)`:

| | risk | reward |
|---|---|---|
| long action k | 2.5% | 5% |
| short action k | **5%** | **2.5%** |

And **no short action anywhere in the space has a 2.5% stop**. A policy that learned "tight stop, wide target" gets the opposite geometry the moment it goes short — silently.

## Why not just fix it

Mirroring the magnitudes keeps the action-space **size identical** while changing what every short index *means*. A trained checkpoint would load without complaint and trade a different strategy. Nothing would fail; the model would just be wrong. That is a call about existing models, not a patch — so it stays open for @BY571.

## What this does

A warning at construction, keyed on the magnitude **sets**, so it fires exactly when the geometry is genuinely asymmetric and stays silent when the swap really does mirror (`(-0.05,-0.1)` against `(0.05,0.1)`). A warning on every construction is noise; one that never fires is decoration. Both mutations fail the tests.

Tests also **pin the current behaviour** — long k is `(-0.025, 0.05)`, short k is `(0.05, -0.025)`, no short has a 2.5% stop — so whoever makes the decision has a baseline to diff against, and the action-space size is asserted unchanged.

Full suite **3721 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_sltp_short_asymmetry_279.py
+++ b/tests/envs/test_sltp_short_asymmetry_279.py
@@ -9,9 +9,16 @@ ENTRY = 100.0
 
 
 def _geometry(side, sl_pct, tp_pct):
-    """Effective (risk %, reward %) at a real entry, through the real bracket helper."""
+    """Effective (risk %, reward %) at a real entry, through the real bracket helper.
+
+    SIGNED against the position, not `abs()`: a bare magnitude cannot tell a stop below
+    entry from one above it, so a mutation putting a short's stop BELOW entry -- a
+    real-money inversion -- left every assertion here passing. Risk must come out
+    positive for a correctly placed bracket and negative for an inverted one.
+    """
     sl_price, tp_price = calculate_bracket_prices(side, ENTRY, sl_pct, tp_pct)
-    return (abs(sl_price - ENTRY) / ENTRY, abs(tp_price - ENTRY) / ENTRY)
+    direction = 1.0 if side == "long" else -1.0
+    return ((ENTRY - sl_price) * direction / ENTRY, (tp_price - ENTRY) * direction / ENTRY)
 
 
 @pytest.mark.parametrize("stoploss,takeprofit", [
@@ -21,7 +28,9 @@ def _geometry(side, sl_pct, tp_pct):
     # rather than pairing index against index. Long k=1 risks 5% to make 10%; short k=1
     # risks 10% to make 5%.
     ([-0.05, -0.1], [0.05, 0.1]),
-    ([-0.01, -0.02, -0.03], [0.01, 0.02, 0.03]),
+    # Asymmetric lengths: the relationship is an identity, not an artefact of equal-sized
+    # lists -- both halves consume the same product(), so the nth pair always matches.
+    ([-0.02, -0.05], [0.03, 0.06, 0.12]),
 ])
 def test_short_k_has_the_risk_and_reward_of_long_k_exchanged(stoploss, takeprofit):
     """Measured through calculate_bracket_prices, not by reading the tuple.
@@ -37,6 +46,10 @@ def test_short_k_has_the_risk_and_reward_of_long_k_exchanged(stoploss, takeprofi
     for k, (long_action, short_action) in enumerate(zip(longs, shorts)):
         long_risk, long_reward = _geometry(*long_action)
         short_risk, short_reward = _geometry(*short_action)
+        assert long_risk > 0 and short_risk > 0, (
+            f"action {k}: a stop on the wrong side of entry (long {long_risk:.4f}, "
+            f"short {short_risk:.4f}) -- the bracket is inverted, not merely exchanged"
+        )
         assert (short_risk, short_reward) == pytest.approx((long_reward, long_risk)), (
             f"action {k}: long is {long_risk:.4f}/{long_reward:.4f}, short is "
             f"{short_risk:.4f}/{short_reward:.4f} -- expected the exchange of the long's"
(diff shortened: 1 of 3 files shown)


BY571:
@CharlieHelps review requested — partial for #279, and the scoping is the thing I'd like challenged.

#279 asks whether the inverted short geometry is intentional, and says its own fix "needs a decision, not just a patch." I agree, so **this deliberately does not fix it.** The reason is specific: mirroring the magnitudes keeps the action-space **size identical** while changing what every short index *means* — so a trained checkpoint loads without complaint and trades a different strategy. Nothing fails; the model is just wrong. That is a call about @BY571's existing models.

What I've done instead is make the asymmetry visible: a warning keyed on the magnitude **sets**, firing exactly when the geometry is genuinely asymmetric and silent when the swap really does mirror ( against ).

**Push back if you think this is the wrong call.** Two coherent objections: (a) a warning is the classic way a known bug becomes permanent, and the honest move is to fix it and force a checkpoint migration; (b) if it *is* a bug, shipping a warning that users will filter out is worse than shipping nothing. I chose the interim because the silent-checkpoint-drift failure mode is the one I'd least want to cause, but I hold that loosely.

The tests also pin the current behaviour — long k , short k , no short with a 2.5% stop — so whoever makes the decision has a baseline to diff against.

Suite 3721 passed, 0 xfailed.

BY571:
@CharlieHelps (reposting — my previous comment was mangled by shell expansion.)

Review requested on this partial for #279, and **the scoping is the thing I would like challenged**.

#279 asks whether the inverted short geometry is intentional, and says its own fix "needs a decision, not just a patch." I agree, so this deliberately does **not** fix it. The reason is specific: mirroring the magnitudes keeps the action-space **size identical** while changing what every short index *means* — so a trained checkpoint loads without complaint and trades a different strategy. Nothing fails; the model is just wrong. That is a call about existing models.

What I did instead is make the asymmetry visible: a warning keyed on the magnitude **sets**, firing exactly when the geometry is genuinely asymmetric and silent when the swap really does mirror (`(-0.05, -0.1)` against `(0.05, 0.1)`).

**Push back if you think this is the wrong call.** Two coherent objections:

1. A warning is the classic way a known bug becomes permanent, and the honest move is to fix it and force a checkpoint migration.
2. If it *is* a bug, shipping a warning users will filter out is worse than shipping nothing.

I chose the interim because silent checkpoint drift is the failure mode I would least want to cause, but I hold that loosely.

The tests also pin the current behaviour — long k is `(-0.025, 0.05)`, short k is `(0.05, -0.025)`, and no short action has a 2.5% stop — so whoever makes the decision has a baseline to diff against. Both mutations fail: removing the warning, and making it fire unconditionally.

Suite 3721 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `1d24813` — the review dismantled my approach and it was right to. **The warning is gone entirely.**

**The trigger was the wrong test.** It compared magnitude *sets*, which only detects that the set of short geometries differs from the set of long ones — not the per-index inversion that is the actual defect. So `(-0.05, -0.1)` against `(0.05, 0.1)` was **silent**, while long k=1 risks 5% to make 10% and short k=1 risks 10% to make 5%. Fifteen existing short-enabled constructions hit that false negative, and my own test enshrined one of them as "mirrored → silent".

**Which exposes the real shape of the thing: it is not conditional.** Short k is the long's risk and reward exchanged whenever the two lists differ element-wise — a property of the construction, not of unlucky inputs. So a warning is the wrong vehicle entirely: it fired on **343 of 403** short-enabled constructions (85%), including every shipped default. That is not a signal, it is a log line on every SLTP env in the repo.

It also crashed a working config (`min()` on an empty `stoploss_levels`, which `__post_init__` permits) and warned when `include_short_positions=True` produced *zero* short actions.

So it is documented where a user reads it — the public docstring and `utils/README.md`, both with the geometry table and the explicit statement that symmetric-*looking* pairs are still inverted.

**On your objection (1), which I flagged as the risk of this approach:** the review found it already materialising. My warning survived a true-mirror mutation — i.e. it would have kept firing after #279 was fixed, becoming a false positive nobody would trust. The tests now **fail** when the bug is fixed (4 of 6), so they cannot outlive it: whoever makes the decision is forced to update them rather than leaving a stale guard behind.

The tests also measure through `calculate_bracket_prices` at a real entry rather than reading the tuple, since the claim is about the prices the venue receives. The review confirmed nothing downstream negates the swap — so the premise held even though my implementation of it did not.

Suite 3721 passed, 0 xfailed.

BY571:
@CharlieHelps round 2 at `d72c58f`. Documenting the defect went wrong in both of the ways the code had, which is worth recording.

**Every number in both tables was wrong**, because `k` was never defined. Both documents state that index 0 is HOLD, eleven lines above a table whose first row is `0 | 2.5% / 5%`. The `k` meant a position within each side's sublist — stated nowhere, and `create_sltp_action_map`'s own example (`action_map[3]  # Short`) primes the opposite reading. Under the real indexing `action_map[2]` is `('long', -0.025, 0.1)` = 2.5%/10%, not the 2.5%/20% I claimed, and there is no short action at index 2 at all. The tables now use real map indices (1 and 10, 3 and 12), each verified through `calculate_bracket_prices`.

**And my stated condition was the deleted bug, restated as prose.** I wrote "this holds whenever the two magnitude lists differ element-wise" and cited `(-0.05, -0.1)` against `(0.05, 0.1)` as an instance — but those lists do *not* differ element-wise, and are inverted anyway. That is exactly the set-equality error I removed the warning for, written straight back into the documentation one round later. The exchange is **unconditional**: both halves iterate the same `product()`, so the nth long and nth short always carry the same magnitude pair the other way round; it is only *invisible* when every magnitude is equal.

**The tests' `abs()` discarded which side of entry a bracket sits on**, so a mutation putting a short's stop *below* entry — a real-money inversion, not an exchange — left all six passing. The geometry is signed against the position now and that mutation fails 4. (Thirteen tests elsewhere already caught it, so this was an overclaiming docstring rather than a coverage hole.)

Two tests dropped: the action-space-size one restated the formula it tested and survived every mutation, and `test_sequential_sltp.py` already asserts it alongside `action_spec.n`. One redundant parametrize case replaced with asymmetric list lengths.

Suite 3720 passed, 0 xfailed. Both mutations verified: stop-on-wrong-side fails 4, and the true-mirror fix for #279 fails 4 — so these tests still cannot outlive the bug they pin.

PR #382, opened 2026-08-18
feat(alpaca): add the close-position escape hatches the futures exchanges have

BY571:
First independent slice of #289.

The issue sequences most of itself after the #288 dedup, so parity features land once rather than five times — but this one is **not** a dedup candidate, because alpaca is the only env that lacks it.

## The gap

Alpaca hardcoded `close_all_positions()` at construction and **never closed on reset**. So there was no way to keep a position across a restart, and no way to flatten a stale one at an episode boundary. The four futures exchanges have had both switches all along:

```python
close_position_on_init: bool = True
close_position_on_reset: bool = False
```

This matters more on alpaca than on futures: a restart there leaves an equity position that outlives the process.

## Compatibility

Defaults reproduce the previous behaviour **exactly** — flatten on construction, not on reset — so no existing config changes meaning. All 138 alpaca tests passed unchanged before the new ones were added.

## Verification

Tested on both variants through a trader that counts flatten calls, since the behaviour under test is whether the call happens at all. Mutation-checked: ignoring the init flag fails 6 tests, dropping the reset close fails 2.

Full suite **3730 passed**, 0 xfailed.

## Still sequenced behind #288

binance's 236-LOC parallel observation implementation, the `margin_type`/`margin_mode` naming split, and the vectorized envs' missing `reward_function`/`include_base_features`.

### DIFF
--- a/tests/envs/alpaca/test_close_position_flags_289.py
+++ b/tests/envs/alpaca/test_close_position_flags_289.py
@@ -60,7 +60,11 @@ def test_close_position_on_init_is_honoured(Env, Cfg, on_init, expected):
     """It was hardcoded, so there was no way to keep a position across a restart --
     the four futures exchanges have had this switch all along."""
     _, trader = _build(Env, Cfg, close_position_on_init=on_init)
-    assert trader.close_all_calls == expected
+    assert trader.close_symbol_calls == expected, (
+        "the init flatten must be SYMBOL-scoped too: this flag defaults to True, so an "
+        "account-wide call means merely constructing an env flattens unrelated holdings"
+    )
+    assert trader.close_all_calls == 0, "init must not touch the whole account"
 
 
 @pytest.mark.parametrize("Env,Cfg", VARIANTS)
@@ -133,3 +137,34 @@ def test_close_does_not_flatten_the_account(Env, Cfg):
         f"close() flattened the account: {trader.calls}"
     )
     assert "cancel_open_orders" in trader.calls, "close() must still cancel orders"
+
+
+@pytest.mark.parametrize("Env,Cfg", VARIANTS)
+def test_a_flat_account_does_not_warn_that_the_reset_flatten_failed(Env, Cfg, caplog):
+    """On alpaca, `close_position()` returns False for the state we WANTED.
+
+    It wraps a client call that raises when the symbol is already flat (code 40410000),
+    so a bare `if not closed` warns on every reset of a flat account -- training the
+    operator to ignore the one warning that is real. bybit returns True when flat, which
+    is why the copied pattern misfired. The check re-reads the status instead.
+    """
+    class _FlatTrader(_CountingTrader):
+        def close_position(self, *args, **kwargs):
+            super().close_position(*args, **kwargs)
+            return False  # alpaca's wrapper does this when the symbol is already flat
+
+    trader = _FlatTrader(initial_cash=10000.0)
+    env = Env(
+        config=Cfg(symbol="BTC/USD", window_sizes=[10],
+                   close_position_on_init=False, close_position_on_reset=True),
+        observer=MockObserver(window_sizes=[10]),
+        trader=trader,
+    )
+    env._wait_for_next_timestamp = lambda: None
+
+    with caplog.at_level("WARNING"):
+        env.reset()
+
+    assert not [r for r in caplog.records if "close_position_on_reset failed" in r.getMessage()], (
+        "warned that the flatten failed on an account that is already flat"
+    )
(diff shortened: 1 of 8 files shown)


BY571:
@CharlieHelps review requested — first independent slice of #289.

The issue sequences most of itself after the #288 dedup so parity features land once rather than five times. This one is **not** a dedup candidate: alpaca is the only env that lacks it, so there is nothing to consolidate.

Alpaca hardcoded `close_all_positions()` at construction and never closed on reset — no way to keep a position across a restart, and no way to flatten a stale one at an episode boundary. The four futures exchanges have had both switches all along. It matters more on alpaca than on futures: a restart there leaves an equity position that outlives the process.

Defaults reproduce the previous behaviour exactly (flatten on construction, not on reset), so no existing config changes meaning — all 138 alpaca tests passed unchanged before the new ones were added.

**What I would like checked:** the reset close is placed *after* `cancel_open_orders()`, matching the futures order. On alpaca specifically I am less sure that is right — cancelling first then flattening is correct if a resting order could otherwise re-open, but if `close_all_positions()` itself can leave a resting order behind, the order should be the other way round. I could not establish which from the mock, and the real API's behaviour on a partially-filled bracket is the case I cannot test here.

Mutation-checked: ignoring the init flag fails 6, dropping the reset close fails 2.

Suite 3730 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `cfd0b62`. My new code had a real-money defect, and the uncertainty I raised to you was backwards.

**`close_position_on_reset=True` used account-wide `close_all_positions()`.** That iterates `get_all_positions()` over the **whole account**, so opting in would market-close a second env's symbol, or a manual holding, at *every episode boundary*. The four futures envs are symbol-scoped. It uses `close_position()` now and checks the result — the account-wide call returns `{}` on exception and `{sym: False}` per-symbol, so the opt-in could also fail silently.

**The ordering question I asked you was answered in reverse.** I worried the reset's cancel-then-close might be wrong; it is right and matches all four futures envs. The hazard was the **init** path I left untouched: close-then-cancel, where `cancel_open_orders()` falls through to the account-wide cancel-all and revokes the market close submitted a line earlier. `close_all_positions()` does not block, and outside RTH the close cannot fill first — so the init flatten was silently reverted. Alpaca was using opposite orderings at its own two sites.

**`close()` still flattened unconditionally**, which made `close_position_on_init=False` pointless: the position it preserved was market-closed the moment the process exited cleanly, and `examples/online_rl/` calls `env.close()`. It cancels and warns now, matching the futures envs. Two existing tests asserted the old contract and are updated.

Worth reporting: **the ordering test needed two goes.** The first cleared the call log before `reset()`, so it only ever saw the reset ordering — and the init ordering was the broken one, so the mutation restoring it survived. It checks both sites now; all three mutations fail.

Docs corrected too: alpaca's README config block was missing both fields, and `live/README.md` listed them as futures-only.

Suite 3738 passed, 0 xfailed.

BY571:
@CharlieHelps round 2 at `13a7d56`. Round 1's fix removed one account-wide flatten and left the other.

**Init still called `close_all_positions()`** — and that flag defaults to `True`, so merely *constructing* an alpaca env flattened unrelated holdings. Round 1's own comment on the reset path argues account-wide is unsafe because it "would market-close a second env's symbol or a manual holding"; that applies verbatim at init, where the default makes it happen every time. Worse, **my test pinned the account-wide call**, locking the divergence in while the config comment claimed parity with the futures envs. Both sites are symbol-scoped now and both tests assert `close_all_calls == 0`.

**`if not closed` was the wrong failure test on alpaca.** `close_position()` wraps a client call that *raises* when the symbol is already flat (code 40410000, documented in this repo's own order_executor), so it returns `False` for the state we wanted. Every reset of a flat account would have logged "close_position_on_reset failed" — training the operator to ignore the one warning that is real. bybit returns `True` when flat, which is why the copied pattern misfired.

**`close()` used `is not None`** — the form invariant 1 forbids — so a 1e-12 residual warned "still holds a position", and `POSITION_UNKNOWN` (what `get_status()` returns on a 429/5xx) was reported as a held position rather than an unknown one. It goes through `position_direction_from_status` now, guarded, since it is a network read on the shutdown path.

Two corrections to my round 1 claims, listed rather than quietly fixed: **"the training loops in `examples/online_rl/` call `env.close()`" is false** — `ppo/live.py` calls `pbar.close()`, and the `*/train.py` scripts build offline envs. The real callers are `examples/rule_based/live.py`, `llm/local/live.py` and `llm/frontier/live.py`, now named in the docstring. And **"all four futures envs only warn here" is 3 of 4** — binance neither reads status nor warns.

Also: #284's unsettled-close fixture stubbed `close_all_positions`, which init no longer calls, so the close was silently settling — its own self-check caught it. Restubbed.

Suite 3740 passed, 0 xfailed.

PR #383, opened 2026-08-18
chore: pin the version at 0.1.0

BY571:
The sweep's final step. All 30 PRs from this run are merged.

`0.0.1` was hatchling's default and survived **1011 commits**, so nothing installed from this repo could be identified. This sweep merged 30 PRs closing 20 issues, most of them money-path defects in the live and replay pipelines — a reader needs to be able to say which side of that their install sits on.

## Why 0.1.0 and not 1.0.0

The four-exchange dedup (#288 steps 1–3, ~2,000 LOC) is unstarted, and **#279** and **#295** are open design decisions that change behaviour for already-trained checkpoints. That is not a 1.0 surface.

## `__version__`

`pyproject.toml` is what pip installs, but a user reporting a bug prints the attribute — and until now there was nothing to print. Two files holding one number is exactly the shape that drifts, so a test pins them together and separately refuses the `0.0.1` placeholder. Mutation-checked: bumping one file alone fails it.

Full suite **3742 passed**, 0 xfailed.

## Still open after this

| issue | state |
|---|---|
| #279 | Asymmetry documented and pinned; **the mirroring decision is yours** |
| #295 | Premise corrected (the `auto_reset` API doesn't exist in the pinned torchrl); **parts 1–2 are your call** |
| #288 | Steps 4 merged, step 5 measured as ~90% already done; **steps 1–3 remain** |
| #289 | Alpaca flags merged; the rest waits on #288 |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "torchtrade"
-version = "0.0.1"
+version = "0.1.0"
 description = "Reinforcement learning environments for trading applications"
 readme = "README.md"
 requires-python = ">=3.11"
--- a/tests/test_version_pin.py
+++ b/tests/test_version_pin.py
@@ -1,0 +1,33 @@
+"""The version lives in two files; keep them one number."""
+
+import pathlib
+import re
+
+import torchtrade
+
+REPO = pathlib.Path(__file__).resolve().parent.parent
+
+
+def test_the_package_version_matches_pyproject():
+    """Two files holding one number is exactly the shape that drifts.
+
+    `pyproject.toml` is what pip installs and `torchtrade.__version__` is what a user
+    prints when reporting a bug -- so a mismatch means the version in a bug report is
+    not the version that was installed, which is the one case the number exists for.
+    """
+    declared = re.search(
+        r'^version = "([^"]+)"', (REPO / "pyproject.toml").read_text(), re.M
+    )
+    assert declared, "pyproject.toml has no version"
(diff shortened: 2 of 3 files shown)


BY571:
@CharlieHelps review requested — the sweep's final step.

`0.0.1` was hatchling's default and survived **1011 commits**, so nothing installed from this repo could be identified. This run merged 30 PRs closing 20 issues, most of them money-path defects in the live and replay pipelines, and a reader needs to be able to say which side of that their install sits on.

**The judgement I'd like checked is the number.** I chose 0.1.0 rather than 1.0.0 because #288 steps 1–3 (~2,000 LOC of four-exchange dedup) are unstarted and #279/#295 are open design decisions that change behaviour for already-trained checkpoints — that is not a 1.0 surface. But 0.1.0 is also a large jump from 0.0.1 for what is, strictly, a bug-fix run; 0.0.2 is defensible. I went with 0.1.0 because the behaviour changes in this sweep are not all backward-compatible (the SLTP fee reserve, the bankruptcy baseline, the replay observation keys, alpaca's `close()` no longer flattening), and a patch bump would understate that.

`__version__` is new on the package: pyproject is what pip installs, but a user reporting a bug prints the attribute, and until now there was nothing to print. Two files holding one number is the shape that drifts, so a test pins them together and separately refuses the `0.0.1` placeholder — mutation-checked, bumping one file alone fails it.

Not tagged: that is a repo-owner action and I did not want to create a tag that might not match what @BY571 wants the release to be called.

Suite 3742 passed, 0 xfailed.

PR #384, opened 2026-08-18
refactor(live): hoist the two accessors every futures base copied

BY571:
Start of #288 step 1 — with the measurement first.

The issue cites *"base.py 216/224 (96%)"* identical between okx and bybit. That is a **line-level** count. AST-comparing every method across all four bases after normalising exchange names:

```
methods present in all 4 AND byte-identical after rename: 2
  get_account_state
  get_market_data_keys
```

Not 96% of the file. The rest is per-exchange client construction, spec building and observation assembly that only *looks* alike.

## What this does

`get_account_state` and `get_market_data_keys` — one-line returns, four copies each — move to `TorchTradeFuturesLiveEnv`. Harmless in themselves, but this repo has shipped a fix landing on some exchanges and not others three times (lot size #271, `full_done_spec` #272, hedge-mode surface), and a copy that hasn't drifted *yet* is still a copy.

## Two guards

Hoisting an accessor can hide a divergence in what it *returns*, so: one asserts no futures env re-forks either method; the other pins that all four still report the same 6-element `ACCOUNT_STATE`. The data stays per-exchange, only the accessor is shared. Both mutation-checked — re-forking okx fails 1, renaming one of okx's fields fails 1.

## Deliberately not in this slice

Step 1's remaining substance (the per-exchange `__init__` / `_init_trading_clients` / `_build_observation_specs` bodies, ~600 LOC) is untouched. Those genuinely differ per venue, and folding them behind `OBSERVER_CLS`/`TRADER_CLS` is a separate change with a real blast radius — the three small slices of this issue have each produced a live defect (a wrong-timeframe observation, an `AttributeError` in every Alpaca bar, an account-wide flatten).

Full suite **3744 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -2809,3 +2809,44 @@ def test_the_bar_wait_derives_from_the_timeframe_not_a_string_alias():
                           "TimeFrameUnit.Hour", "Hour", "h", "H", "seconds",
                           "TimeFrameUnit.Day", "Day", "D", "d", "hour", "day"}
     assert not regrown, f"the unit alias table is regrowing in the wait path: {regrown}"
+
+
+def test_no_futures_env_re_forks_the_shared_accessors():
+    """`get_account_state` / `get_market_data_keys` were four byte-identical copies.
+
+    One-line returns, so the copies were harmless in themselves -- but this repo has
+    shipped a fix landing on some exchanges and not others three times (lot size #271,
+    full_done_spec #272, hedge-mode surface), and a copy that has not drifted YET is
+    still a copy (#288).
+    """
+    from torchtrade.envs.live.shared.futures_live_base import TorchTradeFuturesLiveEnv
+
+    for name in ("get_account_state", "get_market_data_keys"):
+        offenders = [
+            c.__name__ for c in LIVE_ENVS
+            if issubclass(c, TorchTradeFuturesLiveEnv)
+            for base in c.__mro__
+            if base.__module__.startswith("torchtrade.envs.live.")
+            and not base.__module__.startswith("torchtrade.envs.live.shared")
+            and name in base.__dict__
+        ]
+        assert not offenders, f"{name} re-forked on {sorted(set(offenders))}"
+
+
+def test_every_futures_env_reports_the_same_account_state_layout():
+    """The accessor is shared; the DATA still lives per exchange, so pin it.
+
+    `account_state` is the observation vector a policy consumes, and offline/live parity
+    depends on all venues agreeing on its order. Hoisting the accessor would otherwise
+    hide a divergence in what it returns.
+    """
+    from torchtrade.envs.live.shared.futures_live_base import TorchTradeFuturesLiveEnv
+
+    layouts = {
+        c.__name__: tuple(c.ACCOUNT_STATE)
+        for c in LIVE_ENVS if issubclass(c, TorchTradeFuturesLiveEnv)
+        and hasattr(c, "ACCOUNT_STATE")
+    }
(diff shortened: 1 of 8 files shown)


BY571:
@CharlieHelps review requested — start of #288 step 1, with the measurement first.

The issue cites "base.py 216/224 (96%)" identical between okx and bybit. That is a **line-level** count. AST-comparing every method across all four bases after normalising exchange names gives **2** byte-identical methods, not 96% of the file. The rest is per-exchange client construction, spec building and observation assembly that only looks alike. Same pattern as step 5, which I measured at ~90% already done.

So this slice moves exactly those two one-line accessors. Harmless in themselves — but this repo has shipped a fix landing on some exchanges and not others three times, and a copy that has not drifted *yet* is still a copy.

**The part I would like challenged is the stopping point.** Step 1's stated goal is folding the four bases behind `OBSERVER_CLS`/`TRADER_CLS`, ~600 LOC. I have not done that, on the grounds that the three small slices of this issue have each produced a live defect — a wrong-timeframe observation, an `AttributeError` in every Alpaca bar, and an account-wide flatten that would have closed unrelated holdings. Each was caught in review, but each also passed 1300+ tests first. If you think the `OBSERVER_CLS` fold is safe enough to do in one go, say so; my read is that it wants a session with room to see it through, not the tail of one.

Two guards, because hoisting an accessor can hide a divergence in what it *returns*: one asserts no futures env re-forks either method, the other pins that all four still report the same 6-element `ACCOUNT_STATE`. Both mutation-checked.

Suite 3744 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `b1b877f`. The hoist stopped one level too early, and both guards were defeatable.

**Hoisting to the futures base left alpaca holding the only copies** — turning "five byte-identical copies" into two rather than one, with alpaca's byte-identical to the new shared pair. Both accessors are on `TorchTradeLiveEnv` now, which every live env inherits. Deliberately *not* folding in `core/offline_base.py`'s versions: those return `self.account_state`, an instance attribute — a different implementation, not the same duplication.

**The re-fork guard filtered on `torchtrade.envs.live.*`, and `SLTPMixin` lives in `envs.utils`** — which is the one place a re-fork actually wins. All four SLTP envs are `class X(SLTPMixin, XBase)`, so the mixin *precedes* the shared base in the MRO: a method defined there shadows the hoisted one on every venue at once, with the guard green. Verified NOT CAUGHT before, CAUGHT now. It asserts the MRO owner is `TorchTradeLiveEnv` rather than checking a module allowlist, which also closes the intermediate-class bypass.

**The layout guard was fail-open and checked the wrong property.** `hasattr(c, "ACCOUNT_STATE")` meant deleting the attribute *dropped the class from the comparison* instead of failing it — while `get_account_state()` would then raise on every step. And it asserted only that venues **agree**: renaming a field consistently across all of them passed, while breaking every consumer that indexes `account_state` by meaning. It pins the six names now, excludes the abstract bases *by identity* rather than by `hasattr`, and covers alpaca — which the `issubclass(..., FuturesLiveEnv)` filter had excluded entirely.

All three bypasses re-tested: SLTPMixin shadowing, a consistent rename across all venues, and alpaca alone diverging — CAUGHT, CAUGHT, CAUGHT.

Dead `List` imports removed from the four futures bases; they were the only import the hoisted methods used.

On the stopping point I raised: still standing. The `OBSERVER_CLS`/`TRADER_CLS` fold (~600 LOC) is untouched, and this round is a fair illustration of why — a two-method hoist produced two defeatable guards and a half-applied dedup.

Suite 3744 passed, 0 xfailed.

PR #385, opened 2026-08-18
refactor(live): hoist the one identical trade-info builder (#288 step 2)

BY571:
Measured first, as with steps 1 and 5.

The issue describes step 2 as *"4 byte-identical `_create_trade_info`, near-identical `_step`/`_execute_market_order`/`_handle_close_action`/`_execute_fractional_action` (~500 LOC)"*. AST-comparing every method across the four `env.py` files after normalising exchange names:

```
identical across all 4 env.py: 1
  _create_trade_info
```

The issue's own wording is accurate — **one** method is identical; the rest are *near*-identical, and near-identical is where the per-venue facts live (taker fee, lot rounding, position-side handling). Sharing those means parameterising exactly the parts that differ, which is a different change from removing a copy.

## What this does

Hoists the one. Its **defaults** are the point: `_step` returns this dict on paths that did not trade, so a missing key reads downstream as an *absent fact* rather than a default one — four copies were four chances for one venue's default set to drift while every test stayed green.

## Two guards

One asserts the MRO owner is `TorchTradeFuturesLiveEnv` — not a module allowlist, since `SLTPMixin` sits ahead of the shared base for all four SLTP envs and a copy there would shadow the hoisted one on every venue at once (that exact bypass defeated the guard in #384). One pins the five default keys, since that is what a drifted copy would change. Mutation-checked: dropping a default fails.

Full suite **3746 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -173,7 +173,10 @@ def test_no_live_env_overrides_shared_method(env_cls, method):
     )
 
 
-@pytest.mark.parametrize("method", ["_get_observation", "_get_portfolio_value"])
+@pytest.mark.parametrize(
+    "method",
+    ["_get_observation", "_get_portfolio_value", "_create_trade_info"],
+)
 @pytest.mark.parametrize("env_cls", FUTURES_ENVS, ids=lambda c: c.__name__)
 def test_no_futures_env_reforks_the_shared_observation(env_cls, method):
     """The 4 futures exchanges share ONE _get_observation/_get_portfolio_value.
@@ -2870,25 +2873,6 @@ def test_every_live_env_reports_the_same_account_state_layout():
     assert not wrong, f"venues disagree with the universal account_state vector: {wrong}"
 
 
-def test_no_futures_env_re_forks_the_trade_info_builder():
-    """Four identical `_create_trade_info` copies (#288).
-
-    MRO owner, not a module allowlist -- `SLTPMixin` sits ahead of the shared base for
-    all four SLTP envs, so a copy there would shadow the hoisted one on every venue at
-    once while a prefix filter stayed green.
-    """
-    from torchtrade.envs.live.shared.futures_live_base import TorchTradeFuturesLiveEnv
-
-    wrong_owner = {
-        c.__name__: next(b for b in c.__mro__ if "_create_trade_info" in b.__dict__).__name__
-        for c in LIVE_ENVS
-        if issubclass(c, TorchTradeFuturesLiveEnv)
-        and next((b for b in c.__mro__ if "_create_trade_info" in b.__dict__), None)
-        is not TorchTradeFuturesLiveEnv
-    }
-    assert not wrong_owner, f"_create_trade_info re-forked on: {wrong_owner}"
-
-
 def test_the_trade_info_defaults_are_the_whole_contract():
     """A path that did not trade still returns this dict, so a MISSING key reads
     downstream as an absent fact rather than a default one -- which is what four copies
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps review requested — #288 step 2, measured first as with steps 1 and 5.

The issue describes step 2 as "4 byte-identical `_create_trade_info`, near-identical `_step`/`_execute_market_order`/`_handle_close_action`/`_execute_fractional_action` (~500 LOC)". AST-comparing every method across the four `env.py` files after normalising exchange names gives **one** identical method — which is exactly what the issue says. The rest are *near*-identical, and near-identical is where the per-venue facts live (taker fee, lot rounding, position-side handling).

**That distinction is the thing I would like your view on.** Sharing the near-identical methods means parameterising precisely the parts that differ — which is a different change from removing a copy, and arguably makes the shared code harder to read than four honest variants. My inclination is that step 2 is essentially complete after this, and the remaining "~500 LOC" is not duplication so much as four implementations of one shape. If you think the `_step` bodies genuinely should converge behind a template method, say so.

The hoisted method's **defaults** are the real content: `_step` returns this dict on paths that did not trade, so a missing key reads downstream as an *absent fact* rather than a default one. Four copies were four chances for one venue's default set to drift silently.

The guard asserts the MRO owner rather than a module prefix — the prefix version was defeated by `SLTPMixin` in #384, since it precedes the shared base for all four SLTP envs and a copy there shadows the hoisted method on every venue at once. Second guard pins the five default keys; mutation-checked.

Suite 3746 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `08baee4`. My claim that step 2 was essentially complete was **wrong**, and both of my new tests were defective in different ways.

**I reinvented a guard this file already had.** `test_no_futures_env_reforks_the_shared_observation` is parametrized over `FUTURES_ENVS × method` and asserts identity against `TorchTradeFuturesLiveEnv` — which catches an `SLTPMixin` fork perfectly well. My justification for a new MRO-walking test ("a prefix filter stayed green") described the guard from **#384**, not this file's. Adding `_create_trade_info` to that list is a one-line diff with equal coverage and better per-venue failure reporting. My version was also **fail-open** — no floor on how many classes it checked, so an empty `LIVE_ENVS` passed it green.

**The defaults test missed the one semantic this hoist changed.** The only real change is `dict(...).update(kwargs)` → `{**defaults, **kwargs}`, yet the only kwargs I asserted were keys already in the defaults. A merge keeping only *known* keys passed both guards while silently dropping `at_target` — which `core/live.py` reads to re-arm `current_action_level`, so the dropped fact is money-moving. Covered now, mutation-checked.

**And the pairwise matrix answers the question I put to you, against me:**

| method | identical cluster |
|---|---|
| `_execute_market_order` | bitget ≡ bybit ≡ okx byte-identical; binance differs by **one log-string character** |
| `_handle_close_action` | bitget ≡ bybit ≡ okx byte-identical; binance differs by 5 lines |
| `_step` | binance ≡ bitget with **zero** differing code lines |

So two of the four "remaining" methods have an exact 3-of-4 clone cluster — I was reading the all-four intersection and concluding there was nothing left.

**I have not hoisted them, and I want your read on why.** binance's `_handle_close_action` reports `side="CLOSE"` instead of the real side, and omits a `self.position.current_position = 0` the other three carry. That is a question about whether binance *drifted* — the exact defect class #288 exists to catch — not a refactor. Folding it into a dedup would bury it. My inclination is to file it separately and hoist the 3-venue clusters only once binance's difference is understood as intentional or not.

Suite 3757 passed, 0 xfailed.

PR #386, opened 2026-08-18
fix(binance): a close reported "CLOSE" as its side, losing the side it sent

BY571:
The divergence that was blocking #288 step 2's remaining hoists — resolved as **drift**.

## The finding

binance's `_handle_close_action` reported `side="CLOSE"` where bitget, bybit and okx all report the order side actually sent (`"sell"` closing a long, `"buy"` closing a short).

`"CLOSE"` is an *action name*, not a side — and `closed_position=True` sits in the same dict, so the literal added nothing while losing the one fact that field carries. binance also skipped the `self.position.current_position = 0` the other three set on success.

Nothing consumes `side` expecting `"CLOSE"`. A venue reporting an action name where its siblings report a side is exactly the silent per-exchange divergence this issue exists to remove.

## Why this had to come first

Resolving it makes all four `_handle_close_action` **byte-identical**, so the hoist that was blocked is now a clean four-venue one rather than three-plus-an-override.

That ordering was the point: folding the copies together first would have buried the question of which version wins, and one venue's observable output would have changed as a *side effect of a refactor* rather than as a decision.

## Verification

`_handle_close_action` added to the existing parametrized re-fork guard, plus tests pinning the side for both directions and the no-position path. Mutation-checked — the `"CLOSE"` literal fails 2, an inverted side fails 2, an okx re-fork fails 1.

Full suite **3772 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -175,7 +175,12 @@ def test_no_live_env_overrides_shared_method(env_cls, method):
 
 @pytest.mark.parametrize(
     "method",
-    ["_get_observation", "_get_portfolio_value", "_create_trade_info"],
+    [
+        "_get_observation",
+        "_get_portfolio_value",
+        "_create_trade_info",
+        "_handle_close_action",
+    ],
 )
 @pytest.mark.parametrize("env_cls", FUTURES_ENVS, ids=lambda c: c.__name__)
 def test_no_futures_env_reforks_the_shared_observation(env_cls, method):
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps review requested — this is the question I filed on #288 rather than deciding silently, now answered from the code.

binance's `_handle_close_action` reported `side="CLOSE"`; bitget, bybit and okx all report the order side actually sent. My reading is that this is **drift, not design**:

- `"CLOSE"` is an action name, not a side.
- `closed_position=True` is in the same dict, so the literal is redundant with a field that already says it.
- It loses the one fact `side` uniquely carries — which direction went to the venue.
- Nothing consumes `side` expecting `"CLOSE"` (checked every reader).

**Tell me if you read it the other way.** The counter-argument I can construct is that a close is genuinely not a directional order from the policy's perspective, and `"CLOSE"` communicates that intent — but the field is populated from the order side elsewhere in the same codebase, so I think consistency wins.

**The ordering matters more than the fix.** Resolving this makes all four `_handle_close_action` byte-identical, so the previously-blocked hoist is now a clean four-venue one rather than three-plus-an-override. Had I folded the copies together first, one venue's observable output would have changed as a *side effect of a refactor* rather than as a decision — which is the failure mode #288 is about, committed while fixing #288.

binance also skipped the `self.position.current_position = 0` the other three set on success. That one is harmless (`core/live.py` sets it from `sign(desired_action)` anyway), so the three carried a redundant line; it is now shared either way.

Mutation-checked: the `"CLOSE"` literal fails 2, an inverted side fails 2, an okx re-fork fails 1.

Suite 3772 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `31a7054`. The fix was right; my tests pinned everything about it **except the one path where it changes anything**.

binance's missing `self.position.current_position = 0` is invisible on a normal close — `_record_position_after_trade` overwrites the cache from `sign(desired_action)` a moment later. The exception is a **direction switch whose close succeeds and whose open then fails**: that early-returns on `executed=False`, leaving the pre-close value standing. So binance reported a **long against an exchange holding nothing** — invariant 2, exchange truth beaten by the cache — and it is the only thing the zeroing prevents.

Measured: without the fix, `current_position = 1` against `exchange qty = 0.0`. With it, 0. Deleting the three lines now fails this test and one other; before, that mutation only tripped my `SimpleNamespace` unit tests, so nothing exercised the behaviour through a real binance env.

So the PR was fixing two things and I had only argued for one. The `side="CLOSE"` change is cosmetic in practice (nothing reads the field); the `current_position` change is the one with a live consequence, and I had described it as "harmless duplication the other three carry."

Also renamed `TestBitgetClosePositionState` and its "Bitget-specific" banner — the method is shared by all four venues as of this PR, so the class exercises base behaviour through a bitget instance and the label had become false. That is the stale-comment cleanup this repo's own dead-code check mandates on the PR that causes it.

**One divergence I deliberately did not fix:** binance passes uppercase order sides where its siblings pass lowercase, so `trade_info["side"]` now spans both cases. Nothing reads that field — a pre-existing guard asserts no live module does — and the executors normalise, so correcting it means altering the case on a live order call for a cosmetic gain. Say if you'd rather it were uniform.

CI green, suite 3773 passed, 0 xfailed.

PR #387, opened 2026-08-18
refactor(live): normalise binance's order side, then hoist _execute_market_order

BY571:
The last exact clone cluster in #288 step 2 — same sequencing as #386: **resolve the divergence first, then collapse.**

## The divergence

binance built order sides as `"BUY"`/`"SELL"` where bitget, bybit and okx build `"buy"`/`"sell"`. Its executor upper-cases whatever it receives (`order_executor.py:308`), so the **venue never saw a difference** — it surfaced only as a mixed-case `trade_info["side"]` across exchanges, in the field #386 had just unified.

Normalised in both `env.py` and `env_sltp.py`. The latter had four more uppercase literals that my first pass missed — **a guard caught them**, which is the argument for writing the guard before believing the fix.

## The hoist

With that resolved, all four `_execute_market_order` bodies are byte-identical and hoist to `TorchTradeFuturesLiveEnv`.

The "one log-string character" difference I described in #385 was a *symptom, not the cause*: binance omitted `.capitalize()` precisely because its side was already uppercase. Both differences had one root.

## Verification

Added to the existing parametrized re-fork guard, plus a test asserting no venue builds an uppercase side. Mutation-checked — restoring binance's uppercase fails 2, an okx re-fork fails 3.

Three binance SLTP tests asserted the uppercase value and are updated; they encoded the per-venue contract this removes.

Full suite **3797 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -622,7 +622,7 @@ class TestBinanceFractionalPositionResizing:
     ])
     def test_fractional_resizing_executes(self, env, first_action, second_action, should_execute):
         """Changing action level within same direction must trigger trade."""
-        trade_executed = {"executed": True, "amount": 0.01, "side": "BUY",
+        trade_executed = {"executed": True, "amount": 0.01, "side": "buy",
                          "success": True, "closed_position": False}
 
         with patch.object(env, '_execute_fractional_action', return_value=trade_executed) as mock_exec:
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -1020,7 +1020,7 @@ class TestBinanceSLTPNotionalTradeMode:
     @pytest.mark.parametrize("action_tuple,expected_side", [
         (("long", -0.02, 0.03), "buy"),
         (("short", 0.02, -0.03), "sell"),
-    ], ids=["long-BUY", "short-SELL"])
+    ], ids=["long-buy", "short-sell"])
     def test_notional_converts_usd_to_quantity(self, notional_env, action_tuple, expected_side):
         """Notional mode must convert USD to base-asset quantity using current price."""
         env, mock_trader = notional_env
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -242,7 +242,7 @@ class TestBinanceFuturesSLTPTorchTradingEnv:
             # Trade should have been called with BUY
             assert mock_trader.trade.called
             call_kwargs = mock_trader.trade.call_args.kwargs
-            assert call_kwargs["side"] == "BUY"
+            assert call_kwargs["side"] == "buy"
             assert "stop_loss" in call_kwargs
             assert "take_profit" in call_kwargs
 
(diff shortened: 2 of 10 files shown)


BY571:
@CharlieHelps review requested — the last exact clone cluster in #288 step 2.

Same sequencing as #386: resolve the divergence, then collapse. binance built order sides as `"BUY"`/`"SELL"` where its three siblings build `"buy"`/`"sell"`. Its executor upper-cases whatever it receives, so the **venue never saw a difference** — it surfaced only as a mixed-case `trade_info["side"]` across exchanges, in the field #386 had just unified.

**The part worth your scrutiny is that this touches live order construction.** The argument that it is safe is that `order_executor.py:308` calls `side = side.upper()` before anything reaches the API, so binance receives identical bytes either way. I have verified that read, but it is the kind of claim where being wrong means sending a rejected order, so a second opinion is worth having.

Two things I got wrong that are worth reporting:

1. **In #385 I called this "one log-string character".** It was a symptom, not the cause — binance omitted `.capitalize()` *precisely because* its side was already uppercase. Both differences had one root, and describing the visible half as the whole thing is a habit this sweep has repeatedly punished.
2. **My first pass missed four uppercase literals in `env_sltp.py`.** The guard caught them, which is the argument for writing the guard before believing the fix rather than after.

Three binance SLTP tests asserted the uppercase value and are updated; they encoded the per-venue contract this removes.

Mutation-checked: restoring binance's uppercase fails 2, an okx re-fork fails 3.

Suite 3797 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `57c8ec5`. Two findings worth your attention, one of which invalidates the justification I gave you.

**My "now-identical" claim was wrong.** The case change touches *callers*, not the method body — the four bodies still differed by binance's missing `.capitalize()`, and the hoist silently adopted the bitget form, changing binance's error log from `BUY trade failed` to `Buy trade failed`. Non-functional, but the reason I gave for the hoist being safe was not the reason it was safe. I asked you to scrutinise the live-order claim and gave you a false premise for the structural one.

**The guard could not see the value it existed to check.** Hardcoding `side="buy"` in the hoisted method — so a *short* reports `"buy"` — passed the entire suite. `trade_info["side"]` is the field this PR's whole premise is about, and nothing asserted it on any non-SLTP futures path. There is a behavioural test now: a mock trader captures the kwargs and checks the side sent is lowercase **and** matches the direction.

The source grep survives but was defeated five ways: single quotes (this repo mixes styles), `"buy".upper()`, a module-level helper (`getsource` returns only the class block), **alpaca** — in `LIVE_ENVS` but not `FUTURES_ENVS`, so a test named "every venue" checked 12 of 16 classes — and **`futures_live_base` itself**, the very file this hoist moved side construction into, excluded from `FUTURES_ENVS` by construction. It now runs over every live env plus that module, both quote styles.

Your safety question did check out, verified exhaustively: `trade()` is the only method taking a caller-supplied side, `side.upper()` is its first statement, and market/TP/SL/reduce-only/hedge/limit/stop paths all produce **byte-identical API params** either way — including correct bracket-leg inversion.

Dead code the hoist created: `_execute_market_order` was bitget's only `logger` user, so its `import logging` went with it. And binance's non-SLTP `trade_executed` fixture still carried `"BUY"` — the last uppercase env-layer side in `tests/`, re-creating the inconsistency this PR removes.

CI green, suite 3803 passed, 0 xfailed.

PR #388, opened 2026-08-18
fix(vectorized): a broken portfolio value fabricated a +25 reward instead of raising

BY571:
#289 lists this as "align the edge-case behavior" — but the two sides were not merely *different*, one of them was **silent**.

## The divergence

`log_return_reward` **raises** when the previous portfolio value is `<= 0`, and its message says why: that is a *calculation error*, not a market outcome. The prior step already validated it, and bankruptcy terminates before it can go negative.

The vectorized env **clamped to `1e-10`** instead — so `log(new / 1e-10)` produced a reward around **+25** for an env whose accounting had broken, and kept producing rewards for the rest of the batch, with nothing raised.

That is the "guard in the rule" anti-pattern CLAUDE.md names: *a guard that absorbs a nonsense value is worse than none, because it makes the nonsense silent.*

It raises now, naming **which env indices** broke — a batch of 64 with one bad accumulator is unactionable without the index.

## What stays

The **new** value is different: `new_pv <= 0` is reachable, it is bankruptcy, and both paths report `-10.0`. That clamp is correct and stays.

## Why this is not sequenced behind #288

Most of #289 waits on the dedup so parity features land once rather than five times. This is not a dedup candidate — the vectorized env hardcodes the reward arithmetic inline rather than calling the shared function, so **nothing but a test keeps the two in step**. Added one, parametrized over gain / loss / flat / bankrupt / past-zero.

Mutation-checked: restoring the clamp fails 2 tests.

Full suite **3810 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/offline/test_vectorized_reward_parity_289.py
+++ b/tests/envs/offline/test_vectorized_reward_parity_289.py
@@ -34,25 +34,80 @@ class _History:
         self.portfolio_values = values
 
 
-@pytest.mark.parametrize("old,new", [
-    (10000.0, 10100.0),   # gain
-    (10000.0, 9900.0),    # loss
-    (10000.0, 10000.0),   # flat
-    (10000.0, 0.0),       # bankrupt -> both report -10.0
-    (10000.0, -50.0),     # past zero, same
-])
-def test_the_vectorized_reward_matches_log_return_reward(old, new):
-    """Same formula, same bankruptcy constant. The vectorized env hardcodes the
-    arithmetic inline rather than calling the shared function, so nothing but a test
-    keeps the two in step."""
-    scalar = log_return_reward(_History([old, new]))
+@pytest.mark.parametrize("leverage,threshold", [(20, 0.0), (50, 0.0), (20, 0.1)])
+def test_a_wiped_lane_terminates_instead_of_stepping_into_a_broken_reward(leverage, threshold):
+    """The reachable defect, and the reason the raise below is an assertion not a crash.
 
-    safe_new = torch.tensor([new]).clamp(min=1e-10)
-    vectorized = torch.log(safe_new / torch.tensor([old]))
-    vectorized = torch.where(
-        torch.tensor([new]) <= 0, torch.full_like(vectorized, -10.0), vectorized
+    `_apply_liquidation` clamps a wiped balance to exactly 0.0, and termination tested
+    `new_pv < initial * threshold` -- strict. With `bankrupt_threshold=0.0` (explicitly
+    permitted, and documented as "leveraged positions can still terminate on
+    portfolio_value <= 0") that is `0.0 < 0.0`, False. So a dead lane never terminated,
+    never reset, and kept stepping. Measured before the fix: raised at step 1 with
+    `envs [11, 23, 48]` at PV 0.0.
+    """
+    n = 600
+    prices = np.maximum(100 + np.cumsum(np.random.RandomState(0).randn(n) * 2.0), 1.0)
+    df = pd.DataFrame({
+        "timestamp": pd.date_range("2024-01-01", periods=n, freq="1min"),
+        "open": prices, "high": prices + 1, "low": prices - 1,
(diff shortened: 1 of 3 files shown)


BY571:
@CharlieHelps review requested — an independent slice of #289.

The issue calls this "align the edge-case behavior", which understates it. The two sides were not merely different; **one was silent**. `log_return_reward` raises when the previous portfolio value is `<= 0` — a calculation error, since the prior step validated it and bankruptcy terminates before it goes negative. The vectorized env clamped to `1e-10`, so `log(new / 1e-10)` produced a reward around **+25** for an env whose accounting had broken, and kept producing rewards for the rest of the batch.

**The judgement I want checked is raising inside a vectorized env.** One bad accumulator now aborts the whole batch. The argument for it: a fabricated +25 reward trains the policy on a fiction, and it is unrecoverable — you cannot tell afterwards which transitions were real. The argument against: a batch of 64 dies for one env's bug, where clamping degrades gracefully. I went with raising because this repo's stated rule is to validate at the boundary and never guard in the rule, and because "graceful" here means "silently wrong". If you think a vectorized env warrants different treatment from a scalar one, say so — that would be a coherent position and it would change the fix.

The error names the offending env indices, since a batch of 64 with one bad accumulator is unactionable without them.

The **new** value is deliberately left alone: `new_pv <= 0` is reachable, it is bankruptcy, and both paths already report `-10.0`.

Not sequenced behind #288 because it is not a dedup candidate — the vectorized env hardcodes the reward arithmetic inline rather than calling the shared function, so nothing but a test keeps the two in step. Added one, parametrized over gain/loss/flat/bankrupt/past-zero.

Suite 3810 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `a896aa6`, and it **inverted this PR**. My premise was wrong and the fix as originally written was a regression. Worth reading in full before the diff.

**The +25 fabricated reward I cited does not occur.** Measured over 1.5M env-steps across base and SLTP, leverage {5,20,100}, fee {0,4bp}: `old_pv <= 0` happened 1,064,063 times and `old_pv` *strictly* negative **zero** times. Once PV reaches 0 the lane holds nothing, so `new_pv` is 0 too and the `-10.0` branch overrode the clamp. main emitted -10.0 forever — never a positive. I had reasoned about what the arithmetic *could* produce without checking what it *did*.

**And `old_pvs <= 0` is reachable**, so my raise replaced a survivable dead lane with a batch-wide crash — reproduced at step 1, `envs [11, 23, 48]` at PV 0.0. The question you might have asked me ("is this genuinely unreachable?") was the one that mattered, and I asserted it rather than testing it.

**The actual defect is one character in the termination predicate.** `_apply_liquidation` clamps a wiped balance to exactly 0.0, and termination tested `new_pv < initial * bankrupt_threshold` — strict. With `bankrupt_threshold=0.0`, which `__post_init__` explicitly permits and the docstring describes as *still terminating on `portfolio_value <= 0`*, that is `0.0 < 0.0`: False. The lane never terminated, never reset, and kept stepping forever on a dead account. Now OR-ed with `new_pvs <= 0` — the config that crashed at step 1 terminates 6–92 lanes over 60 steps and never raises.

With the reachable path closed, the raise becomes a genuine assertion rather than a new crash mode. It also goes into the **SLTP twin**, which overrides `_step` entirely and still clamped — it emitted **+32.21** on `old_pv = -5.0`. Of four offline reward paths, my PR had fixed one.

**My parity test was a tautology.** It re-implemented the formula in the test body, in float32 where the env's money dtype is float64, and never touched env source. Two mutations survived it — including changing the `-10.0` bankruptcy constant, which had **zero coverage repo-wide**. It drives the real env now, and the bankruptcy constant has its own test that first asserts a lane actually reached bankruptcy.

Also dropped the "matching log_return_reward exactly" claim — `safe_new` still diverges for `0 < new_pv < 1e-10`, which is 1.5% of steps at leverage 50 and the **default** threshold.

Suite 3810 passed, 0 xfailed.

PR #389, opened 2026-08-18
fix(offline): include_base_features silently did nothing — reject it instead

BY571:
#289 lists this as a parity gap: the vectorized envs lack `include_base_features`. They do — but measuring first showed the scalar's is a **dead field**. Declared at `sequential.py:72` and read **zero times** anywhere in `offline/` or `core/`.

So the parity fix is not to propagate it.

## Nor is it to delete it

Hydra configs in `examples/online_rl/ppo/env/sequential.yaml` and several test files pass it, so removing the field turns working configs into a `TypeError` for a flag that never did anything. I tried that first and broke 13 tests.

## The field stays; the lie goes

`__post_init__` rejects `True`. It defaulted to `False` and nothing read it — so anyone who set it `True` trained a policy expecting a `base_features` observation that was **never emitted**, with nothing saying so. That is the silent no-op boundary validation exists to prevent. `False` remains valid, which is what every shipped config sets.

## A guard, and what it found

A test now checks every offline config field is read somewhere. It found a **second** unread field, `symbol` — kept, because it labels the user's data rather than switching behaviour, and the offline envs correctly do not branch on it. Both exceptions are listed with reasons, so the distinction is a decision rather than an oversight.

The **live** envs have a real `include_base_features` that `core/live.py` reads; this is the offline configs only.

Mutation-checked: silently ignoring `True` again fails the guard.

Full suite **3816 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/offline/test_offline_config_has_no_dead_fields_289.py
+++ b/tests/envs/offline/test_offline_config_has_no_dead_fields_289.py
@@ -1,10 +1,12 @@
 """An offline config field nothing reads is worse than a missing one (#289)."""
 
+import ast
 import dataclasses
 import pathlib
-import re
 
 import pytest
+
+from torchtrade.envs.core.common_types import MarginType
 
 from torchtrade.envs.offline import (
     OneStepTradingEnvConfig,
@@ -15,8 +17,15 @@ from torchtrade.envs.offline import (
 )
 
 REPO = pathlib.Path(__file__).resolve().parents[3]
-OFFLINE = REPO / "torchtrade" / "envs" / "offline"
-CORE = REPO / "torchtrade" / "envs" / "core"
+# core/live.py is a LIVE module that happens to sit under core/. Sweeping it in was how
+# the first version of this guard cleared `include_base_features`: live.py genuinely
+# reads it, so the offline field looked consumed.
+OFFLINE_SOURCES = [
+    *(REPO / "torchtrade" / "envs" / "offline").rglob("*.py"),
+    REPO / "torchtrade" / "envs" / "core" / "base.py",
+    REPO / "torchtrade" / "envs" / "core" / "offline_base.py",
+    REPO / "torchtrade" / "envs" / "core" / "state.py",
+]
 
 CONFIGS = [
     SequentialTradingEnvConfig,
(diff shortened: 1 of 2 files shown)


BY571:
@CharlieHelps review requested — an independent slice of #289, and the fix is the opposite of what the issue asks for.

#289 lists `include_base_features` as missing from the vectorized envs. It is — but measuring first showed the **scalar's is a dead field**: declared at `sequential.py:72`, read **zero times** anywhere in `offline/` or `core/`. So propagating it would spread a no-op to two more configs.

Deleting it is also wrong: hydra configs in `examples/` and several tests pass it, so removal turns working configs into a `TypeError` for a flag that never did anything. I tried that first and broke 13 tests.

**So the field stays and `__post_init__` rejects `True`.** The reasoning: it defaulted to `False` and nothing read it, so anyone setting `True` trained a policy expecting a `base_features` observation that was never emitted — with nothing saying so.

**Where I'd like your view:** raising is a breaking change for anyone who *has* set `True` and been running happily on the (silently wrong) behaviour. The alternative is a warning. I chose raising because the failure is not degraded output but a missing observation the policy was configured to receive — silently training on the wrong input space is not something a log line makes safe. But someone with `include_base_features: true` in a yaml will now get a hard failure on upgrade, and you may weigh that differently.

The guard also found a **second** unread field, `symbol`. I kept it: it labels the user's data rather than switching behaviour, and the offline envs correctly do not branch on it. Both exceptions are listed with reasons so the distinction is a decision, not an oversight — but if you think `symbol` should also be enforced or removed, it is a one-line change.

Mutation-checked: silently ignoring `True` again fails the guard.

Suite 3816 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `307cf37`. The guard was disarmed for the field it allowlisted, and it missed a worse defect sitting beside it.

**`margin_type=CROSSED` silently produced isolated liquidation math.** `_liquidation_price` calls `isolated_liquidation_price` unconditionally; the field is copied once and read nowhere. So a config asking for cross margin got isolated liquidation prices — which changes *where a position liquidates* — with nothing saying so. Strictly worse than the flag I opened this PR about. Rejected now, same as `include_base_features`.

**The guard could not have caught either.** Three defects in one regex:

- It swept `core/live.py` — a *live* module that happens to sit under `core/` — which reads `include_base_features` for real, so the offline field looked consumed.
- `self.<field>` inside `__post_init__` counted as a read, so **adding a validation raise permanently disarms the guard for that field**. Exactly backwards: the raise exists *because* the field is dead.
- It concatenated every file into one string, so `margin_type` matching once in `sequential.py` exempted it on all five configs.

It parses now — and fixing that exposed a fourth: my first attempt skipped `__post_init__` by `continue`-ing on the `FunctionDef` inside `ast.walk`, which skips nothing, since walk yields a function's children independently of the function node. The disarm mutation still passed. Removed with a `NodeTransformer`.

**And my justification for keeping the field was false.** I told you hydra configs in `examples/` pass it. No example reads `cfg.env.include_base_features` — all seven env makers hardcode the literal `False`. The honest reason is narrower: it is a public field an outside caller may pass, so it is kept and rejected rather than deleted. That also weakens my answer to your raise-vs-warn question: the blast radius is smaller than I implied, since nothing in-repo would have hit it.

The exemption list is now "rejected at the boundary" rather than "descriptive", and the sibling test proving both actually raise is what earns it — parametrized across Sequential, SLTP and OneStep, which all inherit the validation and only one of which I had pinned.

CI green, suite 3821 passed, 0 xfailed.

PR #390, opened 2026-08-18
feat(vectorized): make reward_function pluggable, with a batched signature

BY571:
The last independent item in #289.

The scalar envs take `reward_function`; the vectorized ones did not, and hardcoded the arithmetic inline — which is exactly why the SLTP twin had drifted from the base and from `log_return_reward` (#388 found it emitting **+32.21** where the scalar raises).

## The signature is batched, deliberately

`fn(old_pvs, new_pvs) -> Tensor`, **not** the scalar `fn(history) -> float`.

Matching the scalar signature would mean calling it once per env with a per-lane `HistoryTracker` — a Python loop over the batch, which is the one thing this class exists to avoid. So parity here is *"the hook exists and is honoured"*, not *"the same callable works in both"*.

`batched_log_return_reward` is the default and lives beside the scalar rewards, so the two are read together and the existing parity test pins them to the same numbers.

## Both twins

The SLTP variant overrides `_step` entirely, so it needed the constructor argument threaded through `super().__init__` — the same override that let its reward drift in the first place.

## Tests

The hook is called; it receives batched tensors of shape `(num_envs,)` rather than scalars; and the default is unchanged when nothing is passed.

Full suite **3823 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/offline/test_vectorized_reward_parity_289.py
+++ b/tests/envs/offline/test_vectorized_reward_parity_289.py
@@ -178,3 +178,65 @@ def test_the_bankruptcy_reward_constant_matches_the_scalar():
         f"bankruptcy rewards {set(bankrupt_rewards)} -- log_return_reward returns -10.0"
     )
     assert log_return_reward(_History([100.0, 0.0])) == -10.0
+
+
+@pytest.mark.parametrize("env_cls,cfg_cls", [
+    (VectorizedSequentialTradingEnv, VectorizedSequentialTradingEnvConfig),
+])
+def test_the_reward_function_is_pluggable(env_cls, cfg_cls):
+    """#289: the scalar envs take `reward_function`; the vectorized ones did not, and
+    hardcoded the arithmetic inline -- which is why the SLTP twin drifted from it.
+
+    The signature is BATCHED (`fn(old_pvs, new_pvs) -> Tensor`), not the scalar
+    `fn(history) -> float`: calling that per env would be a Python loop over the batch,
+    which is the one thing this class exists to avoid.
+    """
+    n = 200
+    prices = 100 + np.cumsum(np.random.RandomState(11).randn(n) * 0.1)
+    df = pd.DataFrame({
+        "timestamp": pd.date_range("2024-01-01", periods=n, freq="1min"),
+        "open": prices, "high": prices + 0.2, "low": prices - 0.2,
+        "close": prices, "volume": np.ones(n) * 1000,
+    })
+    calls = []
+
+    def constant_reward(old_pvs, new_pvs):
+        calls.append((old_pvs.shape, new_pvs.shape))
+        return torch.full_like(old_pvs, 0.5)
+
+    env = env_cls(
+        df,
+        cfg_cls(time_frames=["1Min"], window_sizes=[10], execute_on="1Min", num_envs=6),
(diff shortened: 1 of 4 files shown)


BY571:
@CharlieHelps review requested — the last independent item in #289.

The scalar envs take `reward_function`; the vectorized ones did not, and hardcoded the arithmetic inline. That is exactly why the SLTP twin had drifted from the base and from `log_return_reward` — #388 found it emitting **+32.21** where the scalar raises. So this is the structural fix for a defect we already patched by hand.

**The design call I want challenged is the signature.** I used `fn(old_pvs, new_pvs) -> Tensor`, not the scalar `fn(history) -> float`. Matching the scalar signature would mean calling it once per env with a per-lane `HistoryTracker` — a Python loop over the batch, which is the one thing this class exists to avoid.

The cost is real: **a reward function written for the scalar envs will not work on the vectorized ones**, so "parity" here means the hook exists and is honoured, not that the same callable is portable. Someone who has written `sharpe_ratio_reward`-style logic against `history` gets no reuse. The alternatives I considered were (a) accept the loop and document the cost, (b) adapt scalar functions behind a wrapper that materialises a fake history per lane — which would be slow *and* wrong, since the vectorized envs keep no per-lane history. If you think portability matters more than the batch, that changes the shape of this.

`batched_log_return_reward` is the default and sits beside the scalar rewards so the two are read together; the existing parity test pins them to the same numbers.

Tests assert the hook is called, receives tensors of shape `(num_envs,)` rather than scalars, and that the default is unchanged when nothing is passed.

Suite 3823 passed, 0 xfailed.

BY571:
@CharlieHelps round 1 at `8f01f55`. The PR's stated goal — one reward implementation shared by both twins — **was not achieved**, and the way it failed is the interesting part.

The SLTP twin dropped its `old_pvs <= 0` guard when the check moved into `batched_log_return_reward`. **The base env kept its copy.** So a custom reward that deliberately tolerates a non-positive old PV was preempted in one env and honoured in the other:

```
base : custom reward PREEMPTED by hardcoded guard -> ValueError
sltp : custom reward HONOURED
```

That is the same divergence this change exists to remove, simply relocated from the arithmetic to the guard sitting in front of it. Deleted — and the offline suite already confirmed the copy was dead, since the default reward performs the byte-identical check.

**And the one genuinely new line had no test.** Threading `reward_function` through the SLTP env's `super().__init__` is the only novel plumbing in the diff, and dropping it passed the *entire offline suite* — a user passing `reward_function=` would silently get the default. That is the second time this sweep a "parametrize over env classes" list contained one entry and omitted the twin; both pluggability tests cover both now, and the mutation fails 2.

On your earlier question about the batched signature: the review confirmed the incompatibility is real and **fails loudly** — all three scalar rewards raise `TypeError: takes 1 positional argument but 2 were given` at the first step, rather than silently misbehaving. That was the part I was least sure of.

Smaller items: the function-local `import torch` was unnecessary (no circular import, torch is a hard dependency) and cost 111ns per call in a per-step hot path.

The rest verified bitwise: 8 envs × 150 steps across four configs plus a forced-bankruptcy scenario, all identical to main on reward/terminated/truncated/account_state — including the SLTP twin against what it had before this change.

CI green, suite 3826 passed, 0 xfailed.

PR #391, opened 2026-08-18
refactor(live): extract the 19 SLTP config fields every venue declared identically

BY571:
Step 3's config half of #288 — and the one figure in this issue that measurement did **not** shrink.

AST comparison of the four SLTP config dataclasses:

```
binance 23 fields, bitget 25, bybit 24, okx 24
identical by name, type AND default across all four: 19
```

~80 declarations for 19 facts. Compare with the claims measurement deflated: `base.py` "216/224 (96%)" was **2** identical methods; step 5's "392/506 lines" was ~90% already shared.

Each venue keeps what genuinely differs: `symbol`, `time_frames` and `execute_on` differ only in **default** (venue timeframe spellings), and the margin surface differs in **name**.

## The margin naming split is deliberately untouched

binance has `margin_type: MarginType`, the other three `margin_mode: MarginMode`, and bitget adds `product_type`. #289 lists that separately, and whichever name wins **changes one venue's public API** — a decision that should be made on its own rather than buried in a config merge.

This issue has already produced two behaviour changes wearing a refactor's clothes (#386's `side="CLOSE"`, #387's uppercase sides), both caught only because the divergence was resolved *before* the collapse. A test pins the split so the next dedup pass does not "tidy" it.

## The failure-policy coercion moved too

It was a fifth identical copy — and `'halt' == ObservationFailurePolicy.HALT` is `False`, so a hydra-loaded string silently takes the wrong branch. Each subclass's own `__post_init__` now calls `super()`; forgetting that is what the coercion test pins, across all four configs and both policy values.

## Verification

Mutation-checked: dropping a `super()` call fails 2, re-declaring a shared field fails 1 — a redeclaration **silently shadows the base default**, which is the failure mode inheritance introduces here.

Full suite **3844 passed**, 0 xfailed.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -11,6 +11,7 @@ an exchange that did not answer must not read as either.
 """
 
 import ast
+import dataclasses
 import pathlib
 import re
 import inspect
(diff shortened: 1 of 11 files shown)


BY571:
@CharlieHelps please review.

1. **Field *order* changed.** Dataclass inheritance puts base fields first, so every venue config's positional signature is now different. Does anything in this repo, the examples, or the hydra configs construct one of these positionally?
2. **Every subclass `__post_init__` must call `super()`.** Four sites. A missed one silently skips the failure-policy coercion, and `'halt' == ObservationFailurePolicy.HALT` is `False` — so a hydra string would take the wrong branch with no error.
3. **Did any default change?** I checked all 19 fields matched name+type+default before hoisting, but a silent default shift here changes leverage or bracket levels on a live venue.

Also: the margin naming split (`margin_type` vs `margin_mode`) is deliberately NOT resolved here — flag it if you think it belongs in this PR rather than #289.

BY571:
**Update after two review rounds** — the PR grew a real bug fix and shed the concern I asked you to check.

**Correction to my own review request above.** I claimed `'halt' == ObservationFailurePolicy.HALT` is `False`. It is **True** — the enum subclasses `str`. The coercion is still correct, but because it rejects an invalid string at construction, not because it repairs a comparison. Same for point 1: hoisting `symbol` restores it as parameter #1, so the first **5 positional parameters now match main exactly** (4 for bitget, whose 5th was always `product_type`). The reordering is undone rather than documented.

**A validation gap the duplication was hiding.** The trade_mode/position-sizing block had **seven** copies. Six agreed; alpaca's `elif` read `== "notional"` where the others read `in ("notional", "quantity")` — so in `"quantity"` mode it never checked `quantity_per_trade`, which `alpaca/env_sltp.py:336` sizes every trade from. `quantity_per_trade=0` was a silent no-op order and a negative one a reversed order; both passed construction. Now one `validate_position_sizing()` with seven callers.

**All four `__post_init__` are deleted.** The bodies were byte-identical apart from the normalizer name, and each normalizer is a `partial` of the *same* function differing only in `parse_fn`. The forgot-`super()` footgun is now impossible rather than tested.

**A test this PR had silently disarmed.** `test_every_futures_config_coerces_its_failure_policy` picks its subject with `next(v for k, v in ns.items() if k.endswith("Config") ...)`. Importing the shared base into each `env_sltp.py` put it in that namespace *before* the subclass, so all four SLTP parametrizations resolved to the base. Proven by mutation: deleting binance's `super()` left it green. Fixed with the `__module__` filter the `env_cls` discovery three lines above already had.

**Equivalence evidence:** all 5 SLTP configs, every field value after `__post_init__`, plus every boundary error across 5 (trade_mode, kwarg) combinations, diffed against main. **One** difference in the whole matrix — the alpaca fix.

Diff is now **+224 / −250**. Full suite 3844 passed.

Still worth your eyes: whether folding `symbol` into the base with per-venue default overrides is the right call, and whether `validate_position_sizing` belongs in `core/common.py` rather than nearer the SLTP configs.

BY571:
**Round 3 done — including a correction to my own round-2 claim.**

**I was wrong about alpaca.** I said it "sizes every trade from `quantity_per_trade`" in `"quantity"` mode, making a zero a silent no-op order and a negative one a reversed order. `alpaca/env_sltp.py:336` raises `NotImplementedError` for that mode, and the only read of `quantity_per_trade` is the `"notional"` branch — which the old code *did* validate. **Neither order was ever reachable.** Two reviewers caught it independently; my previous comment and commit message both overstated it.

The divergence did expose the opposite defect, still worth fixing: alpaca accepted an unsupported `trade_mode` at construction and only exploded at the first buy, mid-episode, live. Now a `ValueError` in `__post_init__`.

**The finding that justifies the PR's test work:** deleting `validate_position_sizing`'s body outright failed **zero** tests across all seven callers. Worst offline — `SequentialTradingEnvSLTP` with `quantity_per_trade=0` runs 20 buy actions as silent no-ops, `account_state` flat throughout, reward sum exactly `0.0`. A training run burning compute on a degenerate signal with no diagnostic. Now 28 cases across all seven configs.

**Two of my own tests were not testing what they claimed:**
- `margin_type XOR margin_mode` is symmetric, so it **passed** when binance's `margin_type` was renamed to `margin_mode` — the exact unification its docstring named as the risk. Now asserted per venue by name.
- `"__post_init__" not in vars(cls)` failed identically for a *correct* override that calls `super()` and adds venue logic. It pinned syntax. Deleted; the sizing test catches the real failure behaviourally.
- The `__module__` assertion I added last round could not fail — the generator above it filters on the same predicate. Dead assertion, removed.

**Independent verification of the offline path** (highest-priority code here): specs, `check_env_specs`, and 25-step fixed-seed rollouts through both offline SLTP envs, default and non-default config, diffed against main — byte-identical, 100/100 steps. `_normalize_timeframes` survives `pickle`, `deepcopy`, `dataclasses.replace`, and a real spawn-context subprocess round-trip.

**Equivalence matrix vs main:** 5 SLTP configs × every field × every boundary case. One difference — the alpaca mode rejection.

Diff **+262 / −261**; full suite **3869 passed**. Three rounds complete, findings addressed. Merging unless you object.

PR #392, opened 2026-08-18
fix(live): give binance and bitget the reset/close safety the other two already had

BY571:
Next slice of #288 — and the measurement changes what the work is again.

## `base.py` is not a four-way fold

```
method                     all4   identical pairs
__init__                   False  0/6
_build_observation_specs   False  2/6  bin=bit byb=okx
_execute_trade_if_needed   False  2/6  bin=bit byb=okx
_init_trading_clients      False  0/6
_reset                     False  2/6  bin=bit byb=okx
close                      False  1/6  byb=okx
```

**Nothing is identical across all four.** Two clusters — and the split is a safety gap, not formatting:

- **`_reset`** — bybit/okx check `cancel_open_orders()` and `close_position()` and warn on failure. binance/bitget **discard both return values**. A failed cancel leaves live brackets attached to a position the new episode believes is clean; a failed close leaves real exposure `account_state` will not show.
- **`close()`** — *three* versions. bybit/okx warn on an open position and wrap the cancel; bitget warns but lets a cancel failure propagate out of cleanup; binance does neither.

So: resolve the divergence toward the safer version, **then** hoist. That order is what #386 and #387 taught — collapsing first lets the merge silently pick a winner. Verified all four venues return `True` from both calls when flat, so a flat reset does not warn.

## What is deliberately NOT folded

`_build_observation_specs`: binance and bitget call `observer.get_observations()` — **N live exchange fetches** — purely to count feature columns, where bybit/okx use `get_features()`, which builds a dummy frame. A real improvement, but it changes `observation_spec` construction and needs its own equivalence proof. Separate PR.

## Also closes a gap in #391, reported after it merged

`_normalize_timeframes` is four adjacent one-line assignments differing by one word, and **pointing one at another venue’s parser survived the entire suite**. I measured which swaps are actually detectable: **only bybit’s** — its parser uniquely accepts bare integer minutes (`"60"`, `"240"`) and `"D"`. binance, bitget and okx accept the same set as each other on every spelling tried, so a mix-up among those three is invisible *because it is also harmless*. Pinned what is real rather than asserting parser identity.

Plus the second unguarded `next(... endswith("Config") ...)` discovery in the test file — twin of the one #391 fixed, latent only because its caller passes `.env`.

## Verification

Mutation-checked, each restored from a backup copy rather than `git checkout`:

| mutation | failures |
|---|---|
| drop the cancel-failure warning | 1 |
| drop the close_position-failure warning | 1 |
| reset counts its own bar (`advance_hold` default) | 1 |
| hand-rolled exact-zero instead of the dust rule | 13 |
| a venue re-forks `_reset` | 4 |
| bybit gets okx’s timeframe parser | 3 |

Full suite **3889 passed**. Net **−70 lines**.


BY571:
@CharlieHelps please review.

1. **This is a behaviour change for binance and bitget, wearing a refactor's clothes** — which is exactly the pattern #386/#387 caught. They now log warnings on failed reset cleanup where they previously said nothing, and binance's `close()` gained an open-position warning plus exception handling. I verified all four venues return `True` from `cancel_open_orders()`/`close_position()` when flat, so a flat reset should not warn — please check I did not miss a path where one returns falsy in normal operation, which would make every episode start noisy.
2. **Was bybit/okx's version actually the right one to standardize on?** I picked it because warning strictly dominates silence, but binance's `close()` previously let a cancel failure propagate; now it is logged and swallowed. Is swallowing right in cleanup, or should `close()` raise?
3. **`_build_observation_specs` left alone on purpose.** binance/bitget make N live exchange fetches during `__init__` just to count feature columns. Do you agree that belongs in its own PR, or would you rather see it here?

Also: I could only find bybit's timeframe parser to be behaviourally distinguishable from the others. If you know a binance- or okx-native spelling the other parsers reject, that would let me pin those swaps too.

BY571:
**Round 1 done.** It found two real defects in the hoist itself, and a set of tests named for behaviours they never checked.

**`close()` was missing torchrl's keyword.** `EnvBase.close` is `(self, *, raise_if_closed: bool = True)`, and both `TransformedEnv.close` and the collector's shutdown forward it. All four venues carried a bare `def close(self)` — so closing a wrapped env died with `TypeError` on the way out of a rollout. Pre-existing on main, identical in all four, and `polymarket/env.py` already carried the fix. **Folding first is what let one line reach all four venues** — which is the argument for the fold.

**My comment about `_sync_action_level_after_reset` was false for half the venues.** I hoisted bybit/okx's "no-op … never reads `current_action_level`". `binance/env.py:424` and `bitget/env.py:314` both do exactly that comparison. Driven both ways: without the sync the guard sees `level=1.0` and **refuses the trade**; with it, `nan`, and it executes. Load-bearing for 2 of 4.

**bitget could not report the failure the new warning exists for.** Its `cancel_open_orders` logged each per-order failure then `return True` regardless — so "proceeding with potentially stale orders" could fire for three venues and never for bitget.

**Six tests named `test_reset_logs_warning_on_cleanup_failure` asserted only that reset did not raise.** Deleting the warning left all six green. They assert the log now — that mutation fails 4 where it failed 0. And binance's `test_close_method` asserted `cancel_open_orders.assert_called()`, satisfied by `__init__` which also cancels, so a `close()` that did nothing passed; it now fails 5.

Answering my own question 1 above: the open-position warning had **no test on any venue**, and "warn unconditionally" survived the whole live suite — so nothing pinned the no-false-positive property either. Both covered now.

**Deletions:** 31 lines of dead imports the hoist stranded. The four `base.py` are now 170/183/150/152 lines, from 235/261/211/211.

Also enforced a rule both test files only asked for in a comment — "do not define a live-env subclass in a test module". My own `close()` harness did exactly that and got discovered as an eleventh exchange. Discovery is now package-scoped.

Diff **+429 / −305**; full suite **3880 passed**. Round 2 running.

BY571:
**Round 2 done.**

**The `close()` bug was not theoretical, and alpaca still had it.** `TransformedEnv.close()` forwards `raise_if_closed` unconditionally, and `examples/llm/frontier/live.py:150` closes a wrapped live env — so that path raised `TypeError` on main for every futures venue. Demonstrated both sides.

Round 1 fixed the four futures venues through the fold and **left the fifth**, because my signature test named `TorchTradeFuturesLiveEnv` directly. *A test that hardcodes the class you already fixed cannot tell you about the one you did not.* It now discovers every class defining `close()` — exactly two — and reverting alpaca's signature fails it. Alpaca also discarded `cancel_open_orders()` in both `_reset` and `close()`.

(Its `close_position()` divergence stays: it raises when already flat, so warning on `False` would fire every episode boundary. Documented in place.)

**Trimmed by measurement:** okx's reset-cleanup triplet and close-resilience test duplicate bybit's — every mutation of the shared `_reset` moved binance, bybit and okx in exact lockstep. Three copies, one possible failure. Dropping the cancel warning still fails 5 after the trim.

Added the case none of the copies had: **both cleanup calls failing at once**, which catches a short-circuit reporting only the first.

Also: the caplog assertions now filter by logger name rather than asserting the record list is empty — `assert not caplog.records` fails on any unrelated warning landing in the same test, which is how two of these showed up as order-dependent. Full suite run twice, 3878 both times.

Diff **+449 / −343**. Round 3 next.

BY571:
**Round 3 done — three rounds complete.**

The sharpest finding was against my own round-2 fix. **`caplog.at_level(level, logger=X)` sets X's *level*; it does not restrict `caplog.records`.** I passed `logger=` and reported that as scoping the assertions. Mutation-proven wrong: moving the reset warning to `logging.getLogger("somewhere.else")` still passed **3 of the 5** relevant tests, because they searched every record for a substring. Only the two filtering `r.name ==` caught it. Fixed — that mutant now fails 2 where it failed 0.

**Alpaca's `close()` could crash during teardown** — it called `cancel_open_orders()` unguarded, the exact bug class this PR removed from the other four, whose docstring says teardown is where an exception replaces the error you needed. It had the fix from round 2 but not the cover; both now.

**Removed binance's copies of the cleanup tests.** Deleting okx's was justified by "three copies, one possible failure" — adding binance's in the same diff contradicted it. Seven mutations of the shared `_reset`/`close` are all still caught.

**Removed the timeframe-spelling test** — it guards `_normalize_timeframes`, untouched by this PR (last changed at the merge-base), and its three cases are already parametrized in `tests/envs/bybit/test_obs_class.py` one layer down. #391 fallout that belonged in #391.

Independent verification this round: `check_env_specs` on all 10 live envs; `TransformedEnv(alpaca_env, Compose()).close()` TypeErrors on main and is clean here; `super().close()` confirmed observably inert apart from that fix (double-close, `__del__`, `is_closed` all unchanged); alpaca's mocks return truthy so the new warnings add no test noise.

Diff **+443 / −343**. Full suite run twice from a cleared `__pycache__`: **3874** both times. Merging.

PR #393, opened 2026-08-18
refactor(binance): fold the last parallel observation class onto the shared base

BY571:
Closes the biggest remaining item of #289 — and the fold found a latent bug.

## Measured first

Of the 9 methods `BinanceObservationClass` shared with `BaseFuturesObservationClass`, **4 were already byte-identical**: `_default_preprocessing`, `_get_numpy_obs`, `get_keys`, `reset`. What genuinely differed was the API call and the kline layout — which is exactly what the base's hooks exist for. **239 lines → 92.**

## The latent bug

The base sorts fetched klines by timestamp, because some exchanges return newest-first. **Binance's copy did not** — it was the only venue unprotected. Binance returns ascending today so nothing bit, but a reversed response produced **time-reversed observations with no error**: a policy reading the future as the past.

```
reversed response   main: ascending=False      branch: ascending=True
empty response      main: IndexError: single positional indexer is out-of-bounds
                    branch: RuntimeError: Failed to fetch candles for BTCUSDT on 1Minute
fetch raises        main: bare ConnectionError branch: wrapped with symbol + timeframe
```

## One divergence resolved, not dropped

`get_features()` counts feature columns against a **dummy** frame, and binance klines carry `quote_volume`, `trades` and the taker-buy fields that OHLCV does not. A user's `feature_preprocessing_fn` is free to read them — and against the base's OHLCV-only dummy those functions raise `KeyError`. So the base gets a `_dummy_frame` hook (default OHLCV) that binance extends; the venue-shaped frame stays a venue fact. Removing that override fails 2 tests.

## Equivalence vs main

Keys, feature lists, and every array's shape/dtype/sum — for both the default preprocessing **and** a function reading binance-only columns. Identical.

Mutation-checked: dropping the dummy override fails 2, removing the sort fails 1, re-forking `get_observations` fails 1.

## Deliberately not done

`trade()`/`close_position()` still carry a `position_side` parameter with **no caller anywhere** and three unreachable `!= "BOTH"` branches. bybit and okx have no such parameter; bitget has one, also uncalled. Deleting it and adding the hedge-mode surface #289 also asks for are **opposite resolutions of the same question**, and it is the order-placement path — so it wants its own decision, not a silent pick inside an observation refactor.

Full suite **3878 passed**.

### DIFF
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -208,23 +208,29 @@ class BaseObservationClassTests(ABC):
     # get_features tests
 
     def test_the_declared_feature_width_matches_the_observation(self):
-        """get_features() counts columns on a SYNTHETIC frame; get_observations() runs the
-        same fn on real data. Nothing cross-checked that they agree.
+        """get_features() counts columns on a SYNTHETIC frame; get_observations() runs
+        the same fn on real data. Nothing cross-checked that they agree.
 
-        They can diverge whenever the synthetic frame differs from the parsed klines in a
-        way the preprocessing fn is sensitive to -- a missing column, or the wrong dtype.
-        That count IS the observation spec for the venue transforms, so a disagreement is
-        a spec/data mismatch that surfaces at check_env_specs, or silently as a wrong
-        width. Measured: it was reachable on binance (#289 review).
+        The fn must READ a dummy-supplied column, or this cannot fail: with default
+        preprocessing the width is a constant 4 whatever the dummy contains, and the
+        first version of this test passed even with `volume` deleted from the dummy
+        frame outright. That width is what the venue transforms size their spec from.
         """
+        def fn(df):
+            df = df.copy()
+            df["feature_range"] = (df["high"] - df["low"]) / df["close"]
+            df["feature_vol"] = df["volume"] / df["volume"].mean()
+            return df.dropna()
+
         observer = self.create_observer(
             symbol="BTC/USD",
             timeframes=TimeFrame(1, TimeFrameUnit.Minute),
             window_sizes=10,
+            feature_preprocessing_fn=fn,
         )
         key = observer.get_keys()[0]
         declared = len(observer.get_features()["observation_features"])
-        assert declared == observer.get_observations()[key].shape[1]
+        assert declared == observer.get_observations()[key].shape[1] == 2
 
     def test_get_features_default_preprocessing(self):
         """Test get_features with default preprocessing."""
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -207,6 +207,25 @@ class BaseObservationClassTests(ABC):
 
     # get_features tests
 
+    def test_the_declared_feature_width_matches_the_observation(self):
+        """get_features() counts columns on a SYNTHETIC frame; get_observations() runs the
+        same fn on real data. Nothing cross-checked that they agree.
+
+        They can diverge whenever the synthetic frame differs from the parsed klines in a
+        way the preprocessing fn is sensitive to -- a missing column, or the wrong dtype.
+        That count IS the observation spec for the venue transforms, so a disagreement is
+        a spec/data mismatch that surfaces at check_env_specs, or silently as a wrong
+        width. Measured: it was reachable on binance (#289 review).
+        """
+        observer = self.create_observer(
+            symbol="BTC/USD",
+            timeframes=TimeFrame(1, TimeFrameUnit.Minute),
+            window_sizes=10,
+        )
+        key = observer.get_keys()[0]
+        declared = len(observer.get_features()["observation_features"])
+        assert declared == observer.get_observations()[key].shape[1]
+
     def test_get_features_default_preprocessing(self):
         """Test get_features with default preprocessing."""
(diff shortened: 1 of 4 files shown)


BY571:
@CharlieHelps please review.

Worth a hard look:

1. **The `_dummy_frame` hook is the risky part.** `get_features()` determines the observation spec's feature count. If the dummy frame's columns ever diverge from what `_parse_klines` actually returns, the spec and the data disagree — and that only shows up at `check_env_specs` or, worse, as a silent shape mismatch. Is a dummy frame the right mechanism at all, or should feature counting derive from the real parsed schema?
2. **Behaviour changes for binance:** klines are now sorted, an empty response raises `RuntimeError` instead of a pandas `IndexError`, and fetch errors are wrapped. All strictly better, but the exception *type* changed — does anything you know of catch `ConnectionError` from this path?
3. **The `position_side` fork.** I left it. #289 asks both to remove the dead parameter and to add a hedge-mode surface; those are opposite answers. Which do you want for binance — delete it to match bybit/okx, or make it live?

BY571:
**Round 2 found a regression I introduced in round 1.** Worth spelling out because it inverts a plausible simplification.

Round 1 replaced the `_dummy_frame` method with a `_DUMMY_COLUMNS` name tuple, on the argument that `get_features()` only counts columns so dtypes and values never matter. **They do.** Real binance klines carry `trades` as int64 in the hundreds; the tuple built every column with `np.random.rand` — a float in [0, 1). A preprocessing fn branching on either (`is_integer_dtype(df["trades"])`, or a magnitude threshold) takes a *different branch* on the dummy than on live data:

```
declared (from dummy): 1 feature        actual runtime shape: (10, 2)
```

`BinanceOHLCVTransform` sizes its spec from `get_features()`, so that surfaces as `check_env_specs` failing on a shape mismatch. main was never exposed — its dummy used `randint(1, 100)`, matching real dtype and magnitude. Reverted, with the measurement written into the docstring so the next simplification pass has evidence rather than intuition, and pinned by a test using exactly a dtype-conditional fn.

Other round-2 findings:

- **The guard could shrink silently.** `assert shared` catches only *total* vacuity — making `get_features` a property drops it from a callable filter and the check still passes. Now asserts a named subset is present, and runs over **all four venues**; the three folded by earlier PRs had no such guard at all.
- **The declared-vs-actual cross-check moved to `base_exchange_tests.py`** — it is not binance-specific, and three other venues had nothing pinning it.
- **`# binance's own per-request limit` was false** — 500 is the default page size, max is 1000. main's comment was right and I broke it.
- Deleted binance's `test_an_empty_response_says_so` (bybit/okx/bitget have identical tests on the same shared line — 4-for-1) and `test_custom_preprocessing_with_kline_extra_fields` (dominated by the new dtype case).
- "binance was the last venue with a parallel observation class" — alpaca still hand-rolls one. Now says "last **futures** venue".

One thing round 2 corrected round 1 on: the sort test is **not** a duplicate. bybit and okx sort inside their own `_parse_klines`, so their sort tests stay green if the base's sort is deleted; bitget has none. Binance's is the only test in the four suites reaching the base's sort — kept, and its docstring now says so.

**Pre-existing, not this PR, worth its own issue:** `_halting` catches only `(PositionUnknownError, ValueError)`, so a `RuntimeError` from `get_mark_price()` on a *flat* account bypasses `ObservationFailurePolicy` entirely — verified for binance under both HALT and FLATTEN. bybit and okx have no try/except there at all, so they leak whatever the SDK raises.

Diff **+202 / −253**. Full suite **3884 passed**. Round 3 next.

BY571:
**Round 3 done — three rounds complete.** Both findings were against my own round-2 work.

**The cross-check I added in round 2 could not fail.** It ran with *default* preprocessing, whose width is a constant 4 whatever the dummy frame contains — so it passed under the exact regression it was written for, and passed with `volume` deleted from the base dummy outright: **2876 tests, zero failures**. It now supplies a fn that reads dummy columns; deleting `volume` fails 4.

That's the third dead assertion this sweep, and the shape is always the same: *an assertion whose subject cannot vary*. Here the width was pinned by the preprocessing fn, not by the thing under test.

**"at their real dtypes and magnitudes" was half false.** The dtype half is implemented and pinned. The magnitude half is not — dummy values are `rand()` in [0,1) and `trades` averages ~55 against a live ~1500, so a fn branching on a threshold still diverges, including the "or a threshold" example the base's own docstring gave. Unbounded to chase, so it's now stated as a limitation rather than claimed as a property.

Verified this round in isolated worktrees: the dtype regression fails exactly 1 test; removing the base's sort fails exactly 1 (still the only test in the four venue suites reaching it); a `get_features` re-fork fails the guard on the venue that re-forked it — confirmed on **bitget** as well as binance, so the 4-venue parametrize is doing real work for the three folded by earlier PRs.

**Two pre-existing defects found while reviewing, filed rather than folded in:**
- #394 — `ObservationFailurePolicy` is bypassed when a flat account's mark-price fetch fails, on all four venues (bybit/okx don't even wrap it, so the exception type is unpredictable).
- #395 — `base_features` is built from the unprocessed frame, so a coerced NaN reaches the policy.

Diff **+200 / −258**. Full suite **3884 passed**. No response from @CharlieHelps on this or the two merged PRs, so merging on the three review rounds.

PR #396, opened 2026-08-18
perf(live): stop fetching live klines during __init__ just to count columns

BY571:
The item deferred in #392, now that #393 supplied the proof it needed.

## The problem

binance and bitget built their observation spec from `observer.get_observations()` — which issues **one kline request per timeframe** — purely to read `.shape[1]`. Constructing an env hit the exchange N times, and an outage or a short response failed *construction* rather than the first step. bybit and okx already used `get_features()`, whose docstring says "no network calls".

```
main:   get_observations() calls during __init__: 1
branch: get_observations() calls during __init__: 0
observation_spec identical on both
```

With that resolved the four bodies are byte-identical past their first three lines (AST-verified), so they collapse to one method on the shared base. The four `base.py` drop to **137/149/120/122** lines from 170/183/150/152.

## Why this was deferred, and what changed

#392 explicitly left it: *"it changes observation_spec construction and needs its own equivalence proof."* #393 supplied it — binance got a working inherited `get_features()`, and `BaseObservationClassTests.test_the_declared_feature_width_matches_the_observation` now cross-checks the declared width against a real `get_observations()` **on all five venues**.

On top of that, every preprocessing function in this repo and its examples — 15 of them — was run through both paths on both venues before the switch. **0 failures.**

## The mocks were hiding a real asymmetry

The four binance/bitget `check_env_specs` tests caught the switch immediately: their observer mocks stubbed `get_observations` and **not** `get_features`, so the env read a bare `MagicMock` and declared width 0 against emitted 4. bybit and okx already stubbed it. Those mocks now **derive** the feature list from the observations they emit, so a fixture cannot declare a width its own data contradicts.

## Notes

- The shared base's subclass contract still listed `_reset` (hoisted in #392) and `_build_observation_specs`. Corrected.
- The new guard reads the **AST**, not source text — the method's own docstring names `get_observations`, and a substring check matched its own explanation.

Full suite **3893 passed**.

### DIFF
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -16,6 +16,22 @@ from abc import ABC, abstractmethod
 from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
 
 
+def mirror_features_on(observer):
+    """Derive a mock observer's get_features from the observations it actually emits.
+
+    The envs build observation_spec from get_features() (#288), so a fixture that stubs
+    only get_observations declares width 0 against emitted 4 -- which is exactly what the
+    binance and bitget check_env_specs tests caught when the switch landed. Deriving
+    means a fixture cannot declare a width its own data contradicts.
+    """
+    width = next(iter(observer.get_observations().values())).shape[1]
+    observer.get_features = MagicMock(return_value={
+        "observation_features": [f"feature_{i}" for i in range(width)],
+        "original_features": [],
+    })
+    return observer
+
+
 class BaseObservationClassTests(ABC):
     """
     Base test class for exchange observation classes.
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -29,7 +29,6 @@ def mirror_features_on(observer):
         "observation_features": [f"feature_{i}" for i in range(width)],
         "original_features": [],
     })
-    return observer
 
 
 class BaseObservationClassTests(ABC):
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -1,5 +1,6 @@
 """Tests for BinanceFuturesTorchTradingEnv."""
 
+from tests.envs.base_exchange_tests import mirror_features_on
 import logging
 import pytest
 import torch
@@ -33,13 +34,7 @@ class TestBinanceFuturesTorchTradingEnv:
             return obs
 
         observer.get_observations = MagicMock(side_effect=mock_observations)
-        # Derived from the emitted data, so a fixture cannot declare a width its own
-        # observations contradict -- the env builds its spec from get_features (#288).
-        _w = next(iter(observer.get_observations().values())).shape[1]
-        observer.get_features = MagicMock(return_value={
-            "observation_features": [f"feature_{i}" for i in range(_w)],
-            "original_features": [],
-        })
+        mirror_features_on(observer)
         observer.intervals = ["1m", "5m"]
         observer.window_sizes = [10, 10]
 
@@ -587,13 +582,7 @@ class TestBinanceFractionalPositionResizing:
                 obs["base_timestamps"] = np.arange(10)
             return obs
         observer.get_observations = MagicMock(side_effect=mock_observations)
-        # Derived from the emitted data, so a fixture cannot declare a width its own
-        # observations contradict -- the env builds its spec from get_features (#288).
-        _w = next(iter(observer.get_observations().values())).shape[1]
-        observer.get_features = MagicMock(return_value={
-            "observation_features": [f"feature_{i}" for i in range(_w)],
-            "original_features": [],
-        })
+        mirror_features_on(observer)
         observer.intervals = ["1m", "5m"]
         observer.window_sizes = [10, 10]
         return observer
@@ -678,13 +667,7 @@ class TestMultipleSteps:
         mock_observer.get_observations = MagicMock(return_value={
             "1m_10": np.random.randn(10, 4).astype(np.float32),
         })
-        # Derived from the emitted data, so a fixture cannot declare a width its own
-        # observations contradict -- the env builds its spec from get_features (#288).
-        _w = next(iter(mock_observer.get_observations().values())).shape[1]
-        mock_observer.get_features = MagicMock(return_value={
-            "observation_features": [f"feature_{i}" for i in range(_w)],
-            "original_features": [],
-        })
+        mirror_features_on(mock_observer)
         mock_observer.intervals = ["1m"]
         mock_observer.window_sizes = [10]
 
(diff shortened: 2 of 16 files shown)


BY571:
@CharlieHelps please review.

1. **This changes when a construction failure surfaces.** Before, a binance/bitget env that could not reach the exchange failed in `__init__`; now it constructs fine and fails at the first step. That is the intent — but it is a behaviour change for anyone whose startup check was "does the env construct". Do you see a caller relying on that?
2. **`get_features()` runs the user's preprocessing fn on a SYNTHETIC frame.** If a fn reads something the dummy frame lacks (the timestamp columns are omitted on every venue), it now raises during `__init__` where before it did not. I ran all 15 preprocessing fns in this repo and its examples through both paths on both venues — 0 failures — but that is only what is in-tree. Worth a release note?
3. **The mock change is the part I would most want a second opinion on.** binance/bitget observer mocks stubbed only `get_observations`, so they silently declared width 0 once the env started asking for `get_features`. I made them derive the feature list from the observations they emit. Is deriving right, or does it weaken the fixture by making the two agree by construction rather than by assertion?

Note the equivalence proof lives in `BaseObservationClassTests.test_the_declared_feature_width_matches_the_observation` (added in #393), which asserts declared == actual for all five venues with a fn that actually reads dummy columns — deleting `volume` from the base dummy fails it 4 times.

BY571:
**Round 1 done.** It found that this file had already warned me against what I did.

`core/live.py`'s `get_account_state` docstring — on this same issue:

> *"On TorchTradeLiveEnv rather than the futures base so all FIVE venues share it: hoisting to the futures base left alpaca holding the only copies, which turns 'four identical copies' into two rather than one (#288)."*

I hoisted to the futures base. **Alpaca kept its own copy — still fetching live klines in `__init__`, the exact thing this PR removes.** Five copies became two. Now on `TorchTradeLiveEnv`: one implementation, five venues, alpaca stops fetching too.

**Three claims measurement reversed:**

1. *"a short response failed construction"* — it did not. On main, binance, window 10: an empty response raises, a **3-bar response returns shape (3, 4) and constructs fine**. The spec width is a *column* count; row count never touched it. I'd said this in three places.
2. *"byte-identical past their first three lines (AST-verified)"* — false at every offset. Two comment lines only binance/bitget carry break byte identity well past line 3, and "AST-verified" contradicts "byte" since the AST drops exactly those comments. Measured: they share their last **10 statements**, differing prefix 3 for binance/bitget and **2** for bybit/okx.
3. *"`get_features()`, whose docstring says 'no network calls'"* — it says no such thing. That phrase lived in bybit's and okx's own `_build_observation_specs` docstrings, which this PR deletes.

**Two dead guards deleted.** The `window_sizes[i] if i < len(window_sizes)` fallback and the `isinstance(..., list)` normalization are unreachable — every config's `__post_init__` runs `normalize_timeframe_config`. Instrumented both and ran the suite: **0 hits in 3893 tests**. The fallback was also fail-open in the one case it could fire, so `zip(..., strict=True)` raises there instead.

**Tests.** 22 duplicated mock blocks → one `mirror_features_on()` helper. My AST guard parametrized 12 envs that all resolve to one function — 12 copies of one assertion, and it asserted which method is called *by name*, which a rename of the bad path escapes. Replaced with two behavioural tests, one of which closes a real gap: **collapsing every window to `window_sizes[0]` passed 1366 tests**, because every multi-timeframe env test uses uniform sizes.

Answering my own question 3 above: deriving the mock's `get_features` is right, and it's now a shared helper rather than 22 copies.

Diff **+183 / −226**. Full suite **3881 passed**. Round 2 running.

BY571:
**Round 2 done.** It found a regression I introduced by folding.

`self.account_state = self.ACCOUNT_STATE` existed **only in alpaca's copy** of `_build_observation_specs`. Folding the five lost it — and `examples/llm/frontier/live.py:91` and `examples/llm/local/live.py:97` both read `env.account_state` off an `AlpacaTorchTradingEnv` to label the observation for the LLM actor. Both would have raised `AttributeError` with the whole suite green. Restored on the shared method, so all five venues have it uniformly rather than one by accident.

This inverts the usual dedup risk. Every earlier finding in this series was *divergence hidden by duplication* — four copies where one had drifted. This was the opposite: **one copy had something the other four lacked, and folding silently discarded it.**

Two reviewers disagreed on whether it mattered; I settled it by execution — the grep that said "safe" had looked only in `torchtrade/`, not `examples/`.

Other round-2 findings:

- **`test_the_base_features_guard_has_something_to_check` was a tautology.** `pyproject.toml` sets `empty_parameter_set_mark = "fail_at_collect"`, so an empty parametrize is already a collection error. The **fourth dead assertion this sweep** — and, again, one I added to fix a previous round's finding. The whole discovery apparatus is gone: one unparametrized AST test, plus `_build_observation_specs` added to `test_no_live_env_overrides_shared_method`, which is what actually catches a re-fork.
- **`zip(..., strict=True)` had no test** — dropping it failed nothing. Now pinned by an injected observer whose key count disagrees with the config.
- **`_StubObserver.get_observations` is dead** and the justification I wrote for keeping it was false. Same file's "10 cases resolve to 5 builders" was stale — it is 1 now, so a defect fails all 10.
- **"Collapsing every window passed 1366 tests"** — a fourth false figure; 1366 matches no collectable subset. Restated without the invented number.

Mutation-checked: dropping `account_state` now fails 5 (was 0), `strict=True` fails 1 (was 0), reverting to a live fetch fails 13.

Diff **+255 / −246**. Full suite **3902 passed**. Round 3 running.

BY571:
**Round 3 done — three rounds complete.** The false claim it found guarded the bug this PR itself introduced.

My test's docstring said the *data* was pinned separately by `test_every_live_env_reports_the_same_account_state_layout`. **It is not** — that test reads the class constant `tuple(c.ACCOUNT_STATE)` and never constructs an env, so it never touches the instance attribute. Measured: `self.account_state = ["MUTANT"]` — the LLM actors' label vector replaced wholesale — **passed all 2890 tests in `tests/envs` and all 1008 outside it**.

My own test couldn't catch it either: it was an AST check that *some* `self.account_state = ...` assignment exists. It killed the absent-attribute mutant and not the wrong-data one — and an attribute that is present and wrong is worse than one missing loudly. Both AST tests are gone; the label check is one line inside the test that already builds a real env, and it kills both mutants.

**`_declare_base_features_spec` is inlined.** It was extracted at #61 *because five per-venue builders each called it* — this PR folded those five into one, leaving a single caller. Its AST guard went with it; all five venues have behavioural `test_base_features_declared_in_observation_spec` tests, and disabling the inlined declaration fails 5.

**The `strict=True` comment claimed more than the code does.** It catches a *length* mismatch, not an observer returning keys in a different *order* — which passes strict and still mislabels shapes. Unreachable with real observers, but the comment shouldn't imply the fail-open case is fully closed.

Mutation-checked, all five: wrong `account_state` data 1, absent 1, `base_features` never declared 5, window collapse 1, `strict` dropped 1.

Diff **+225 / −266** (net −41). Full suite **3896 passed**. CI green. No response from @CharlieHelps here or on the three merged PRs, so merging on the three review rounds.

PR #397, opened 2026-08-18
fix(live): a malformed candle put NaN straight into base_features

### LINKED ISSUE #395: base_features is built from the unprocessed frame, so a coerced NaN reaches the policy
Found while reviewing #393; pre-existing, not caused by it. Recorded here because I removed the note from a source comment during that PR and it would otherwise be lost.

## What happens

`BaseFuturesObservationClass.get_observations(return_base_ohlc=True)` builds `base_features` **before** preprocessing runs:

```python
if return_base_ohlc and timeframe == self.time_frames[0]:
    base_df = df.copy()                     # <- unprocessed
    observations["base_features"] = self._get_numpy_obs(
        base_df, columns=["open", "high", "low", "close"])[-window_size:]
...
processed_df = self.feature_preprocessing_fn(df)   # <- this is what drops NaN rows
```

Binance parses klines with `pd.to_numeric(..., errors="coerce")` — deliberately, so one malformed row becomes NaN instead of killing the whole fetch, and `_default_preprocessing`'s `dropna()` then removes it. That protection applies only to the feature path. Measured:

```
malformed row at index 57 (inside the last window):
    base_features  NaN -> True
    feature array  NaN -> False
malformed row at index 30 (outside the window):
    both clean
```

So a single bad candle inside the observation window puts NaN into `base_features`, which the SLTP envs read for bracket pricing and which `include_base_features` exposes directly.

## Why it matters

NaN compares False to every operator, which this repo has already been bitten by — see #349, where `nan <= 0` skipped both price fallbacks and returned the NaN. A NaN in `base_features` will not trip a `<= 0` guard anywhere downstream.

## Scope

- Decide whether `base_features` should be built from the frame *after* NaN rows are dropped, or whether the drop should happen at parse time so both paths see the same rows.
- Affects all four futures venues via the shared base; alpaca has its own copy.
- No test anywhere feeds a malformed row through and asserts `base_features` is NaN-free — the existing base-OHLC tests only check shape.

BY571:
Fixes #395. Reproduced before fixing.

## The bug

The venues parse klines with `to_numeric(errors="coerce")` **on purpose** — one unparseable candle becomes NaN instead of killing the whole fetch, and the feature path drops it in preprocessing. But `base_features` is built from the **raw** frame:

```
bad row INSIDE the window   base_features NaN=True    feature array NaN=False
bad row outside the window  base_features NaN=False   feature array NaN=False
```

A single bad candle inside the observation window reached the policy as NaN. **NaN compares False to every operator**, so it trips no `<= 0` guard anywhere downstream — #349 was this same trap one path over, where `nan <= 0` skipped both price fallbacks and returned the NaN as a price.

## The fix is alpaca's

Alpaca's observer already did `base_df.dropna()` before slicing. The futures base did not. One venue had solved it and the shared code had not — the shape #392 hit from the other direction.

## Alpaca's own protection was equally untested

Deleting alpaca's `dropna` failed **zero** tests, because every mock client returns clean candles. So did my first attempt at a guard — a harness-level "base_features has no NaN" assertion running against those same clean mocks, which could not fail. (The sixth vacuous assertion this sweep; the tell is always the same — an assertion whose subject cannot vary.)

Both venues now have a test that **injects** the malformed candle: binance through a raw kline payload, alpaca at the frame level, since which payload coerces to NaN is venue-specific. Removing either `dropna` now fails one.

Full suite **3898 passed**.

### DIFF
--- a/tests/envs/alpaca/test_obs_class.py
+++ b/tests/envs/alpaca/test_obs_class.py
@@ -8,6 +8,7 @@ and stricter assertions live here.
 
 from unittest.mock import MagicMock
 import numpy as np
+import pandas as pd
 from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
 
 from torchtrade.envs.live.alpaca.observation import AlpacaObservationClass
@@ -142,11 +143,6 @@ def test_a_malformed_candle_never_reaches_alpaca_base_features(monkeypatch):
     unguarded on the one venue that had it. Injected at the frame level, since which
     malformed payload produces a NaN is venue-specific.
     """
-    import numpy as np
-    import pandas as pd
-    from torchtrade.envs.live.alpaca.observation import AlpacaObservationClass
-    from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
-
     observer = AlpacaObservationClass(
         symbol="BTC/USD",
         timeframes=TimeFrame(1, TimeFrameUnit.Minute),
@@ -163,7 +159,8 @@ def test_a_malformed_candle_never_reaches_alpaca_base_features(monkeypatch):
             "volume": np.full(n, 10.0),
             "timestamp": pd.date_range("2024-01-01", periods=n, freq="1min"),
         })
-        df.loc[57, "close"] = np.nan  # inside the last 10, as a coerced parse would leave it
+        df.loc[57, "close"] = np.nan  # inside the last 10, (alpaca's SDK returns floats, so a
+        # gap arrives already NaN rather than via a coerced parse)
         return df
 
     monkeypatch.setattr(observer, "_fetch_single_timeframe", frame_with_a_hole)
--- a/tests/envs/alpaca/test_obs_class.py
+++ b/tests/envs/alpaca/test_obs_class.py
@@ -150,7 +150,7 @@ def test_a_malformed_candle_never_reaches_alpaca_base_features(monkeypatch):
         client=MagicMock(),
     )
 
-    def frame_with_a_hole(timeframe, limit=None):
+    def frame_with_a_hole(timeframe):
         n = 60
         df = pd.DataFrame({
             "symbol": ["BTC/USD"] * n,  # alpaca's default preprocessing drops this column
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps please review.

Two things worth a hard look:

1. **Is dropping the row the right resolution?** A NaN candle could instead be forward-filled, or the fetch could be rejected outright. I took alpaca's existing answer (drop it) on the grounds that a candle whose price didn't parse is a hole rather than data, and that interpolating a price is a decision the framework shouldn't make silently. But it does mean `base_features` can now be *shorter* than the requested window when rows are dropped near the end — same as the feature path already behaved. Is that acceptable, or should a short `base_features` raise?

2. **The test only injects at one point per venue.** binance gets a malformed raw kline; alpaca gets a NaN at the frame level (its coercion happens inside the SDK, so there's no equivalent raw payload). bitget/bybit/okx are covered by the shared fix but have no injection test of their own — they'd need one malformed-payload builder each. Worth it, or is the shared-path coverage enough?

Context: the fix itself is three characters of logic — alpaca already had it and the shared base didn't. The more interesting finding is that alpaca's protection had **zero** coverage: deleting its `dropna` failed nothing, and my first guard was a harness assertion that ran against clean mocks and could never fail.

BY571:
**Round 1 done. Both of my earlier attempts at this fix were wrong, in the same way.**

`base_features` and `market_data` go to the policy *together* when `include_base_features` is set, so row *i* of one must be the **same bar** as row *i* of the other. `_default_preprocessing` selects rows with `dropna()` **and** `drop_duplicates()`:

- **Attempt 1**, `dropna(subset=["open","high","low","close"])` — a NaN in `volume` removed the bar from the feature window and kept it here:
  ```
  base : 23:08 23:09 23:10 23:11 23:12
  feat : 23:07 23:08 23:09 23:11 23:12
  ```
- **Attempt 2**, bare `dropna()`, which I described as *"matching the feature path exactly"* — it doesn't. A **repeated bar** desynced them the same way.

Both are worse than the NaN they were fixing: a NaN is at least visible; a misaligned bar is not. And neither was catchable by the NaN-only assertion I'd written. It's `dropna().drop_duplicates()` now, with all four corruption modes parametrized — the duplicate case is what caught attempt 2.

**Three claims corrected:**

1. *"The venues parse klines with `to_numeric(errors='coerce')`"* — **only binance does.** bitget/bybit/okx use `astype(float)`, which *raises* on a malformed string and yields NaN only from a null field. The comment sits in shared code and generalized a binance-only parse to all four.
2. *"trips no `<= 0` guard downstream"* — **stale.** #347 added `if not math.isfinite(price) or price <= 0: raise` to the three SLTP sites reading `base_features[-1, 3]`. The genuinely unguarded consumer is `include_base_features`, which puts the array straight into the policy's TensorDict. **Worth stating plainly: this fix trades a loud ValueError at those three sites for bracket pricing off the previous bar.** Defensible — the previous bar is real data and the exception wasn't recoverable — but that's the actual delta, and it answers my question 1 above.
3. *"as alpaca's observer already did"* — alpaca does a **bare** dropna, not a subset one. The real in-repo precedent is `offline/infrastructure/sampler.py:141`.

Also swept up two things #396 left on main: a duplicated assertion and a reference to the helper it inlined.

Mutation-checked: no drop fails 5, subset-only fails 3, bare dropna fails 1, alpaca losing its own dropna fails 1. Full suite **3902 passed**. Round 2 next.

BY571:
**Round 2 found attempt three was also wrong — same way as one and two.**

The row-selection rule in `_default_preprocessing` is **four** operations: `dropna → drop_duplicates → build features → dropna`. Each of my attempts matched a prefix:

```
attempt 1  dropna(subset=[ohlc])        desyncs on a NaN volume
attempt 2  dropna()                     desyncs on a repeated bar
attempt 3  dropna().drop_duplicates()   desyncs on a ZERO-PRICED bar
```

The last because `open=0, close=0` survives the first dropna, becomes `feature_open = 0/0 = NaN`, and is removed by the dropna that runs *after* features are built. Reproduced against the shipped fix:

```
base : 23:08 23:09 23:10 23:11 23:12
feat : 23:07 23:08 23:09 23:11 23:12
```

**So the problem was never a stale copy of the rule — it's that a copy exists at all.** `base_features` is now sliced from the frame `feature_preprocessing_fn` returned, the same frame the market data comes from, so row *i* of each is the same bar **by construction**. Alpaca had its own incomplete copy and gets the same treatment.

That also extends the guarantee to a **custom** preprocessing fn, which no copy of the default's rule could ever cover — such a fn previously got a silently misaligned `base_features`.

One consequence worth naming: with a custom fn, `base_features` is that fn's open/high/low/close, so a fn that drops those columns now raises `KeyError` instead of misaligning silently. `binance/env_sltp.py:233` reads `base_features[-1, 3]` as the current price — unchanged under default preprocessing, and under a custom fn it was already reading a misaligned bar.

Mutation-checked by re-deriving the rule four ways: `df.copy()` fails 6, `dropna(subset=[ohlc])` 4, `dropna()` 2, `dropna().drop_duplicates()` 1. Each parametrized case is one attempt's counterexample.

Diff **+114 / −23**. Full suite **3903 passed**. Round 3 next.

BY571:
**Round 3 done — three rounds complete.** It found the alpaca half of my own fix was a live regression, with zero coverage.

Alpaca's SDK returns a `(symbol, timestamp)` MultiIndex; only `_default_preprocessing` turns it into a column via `reset_index()`. Reading `processed_df['timestamp']` unconditionally raised `KeyError` for any custom preprocessing fn that leaves the index alone. Main did its own `df.reset_index()` and was immune:

```
main    OK  base_features (10, 4)
branch  KeyError: 'timestamp'
```

**Reachable on the money path** — `alpaca/env_sltp.py:275` calls `get_observations(return_base_ohlc=True)` on every bracket entry regardless of `include_base_features`. The in-repo examples escape only because each happens to call `reset_index(drop=False)`.

Timestamps now resolve from the column *or* the index level — but **not** with a silent `df.index.values` fallback, which would hand back positions as timestamps for a frame whose OHLC came from real bars. If it's in neither, it raises with the reason.

**The coverage finding is the sharper one:** substituting main's entire `get_observations` back into alpaca left `tests/envs/alpaca` at **159 passed**. The alpaca change was completely unguarded. Two tests now — a custom fn keeping the MultiIndex, and one losing the timestamp entirely.

Also: the docstring still said `base_features` is "the raw OHLC data" — it isn't, it's whatever the preprocessing fn returned, so a fn that rescales `close` silently reprices brackets. Now stated where a user reads the contract. Plus an observation-layer guard that no venue re-forks `get_observations` (the env layer had one, this layer didn't) — a bitget re-fork fails 3.

While adding a finiteness assertion I found **`inf` reaching market_data** — `pct_change` off a zero close, which the trailing `dropna` doesn't remove. Identical on main, so out of scope here: filed as #398.

Diff **+257 / −25**. Full suite **3910 passed**. Merging on the three rounds — no response from @CharlieHelps here or on the four merged PRs.

Issue #394, opened 2026-08-18
ObservationFailurePolicy is bypassed when a flat account's mark-price fetch fails


BY571:
Found while reviewing #393 (not caused by it), and quantified per venue.

## What happens

`_halting` in `torchtrade/envs/live/shared/futures_live_base.py` catches only `(PositionUnknownError, ValueError)`. `get_status()` on all four futures venues wraps network failures into the `POSITION_UNKNOWN` sentinel, whose `__getattr__` raises `PositionUnknownError` — so that path is caught and the policy engages correctly.

The gap is `get_mark_price()`, reached from `_current_mark_price()` only when the account is **flat**:

| venue | `get_mark_price()` on network failure | caught by `_halting`? |
|---|---|---|
| binance | wrapped → `RuntimeError` | **no** |
| bitget | wrapped → `RuntimeError` | **no** |
| bybit | **no try/except at all** — raw SDK/network exception propagates | **no** |
| okx | **no try/except at all** — same | **no** |
| alpaca | n/a — no `_halting`/policy machinery; `_get_current_price` degrades to `0.0` | n/a |

Reproduced end to end on binance with a flat account (`get_status()` succeeds with `position_status=None`, `get_mark_price()` raises):

```
policy=ObservationFailurePolicy.HALT    -> BYPASSED, raw RuntimeError escaped _step
policy=ObservationFailurePolicy.FLATTEN -> BYPASSED, raw RuntimeError escaped _step
```

`LiveObservationHalt` is never constructed and the policy is never consulted.

## Why it matters

Bounded to the flat branch — an open position reads mark price off the already-successful `position_status` and never re-fetches — so `FLATTEN` has nothing to flatten in this exact case. But **`HALT` loses its structured signal**: orchestration built around catching `LiveObservationHalt` gets an uncontrolled crash instead. bybit and okx additionally leak whatever raw exception the SDK raises, so the type is not even predictable.

`futures_live_base.py` already carries a comment noting "bybit and okx raise RuntimeError, which `_halting` does not even catch" for a *different* path, so the shape of this is known.

## Suggested scope

- Decide the contract: should a failed market-data read be a `ValueError` (caught) or should `_halting` widen? Widening to bare `Exception` would swallow bugs, so probably the former plus explicit wrapping.
- Give bybit and okx the try/except their siblings have, so the exception type is at least uniform.
- One test per venue asserting the configured policy actually engages on a flat-account mark-price failure — nothing covers this today.

PR #399, opened 2026-08-19
fix(live): dropna removes NaN but not inf, and a zero close produces inf

### LINKED ISSUE #398: inf reaches the observation: pct_change off a zero close survives dropna
Found while reviewing #397; pre-existing, verified identical on `main`.

## What happens

`_default_preprocessing` computes `feature_close = df["close"].pct_change()`. If a candle has `close == 0`, the **next** bar's pct_change is `inf`. The function's trailing `dropna()` removes NaN but **not** `inf`, so it reaches the observation:

```
market_data_1Minute_5 (binance, bar after a zero-priced candle):
  [[0.00651, 0.647, 0.653, 0.640],
   [0.00647, 0.643, 0.649, 0.636],
   [0.00643, 0.638, 0.645, 0.632],
   [inf,     0.630, 0.637, 0.624],     <- here
   [0.00630, 0.626, 0.633, 0.620]]
```

Measured on `main` and on the #397 branch — same output, so #397 neither causes nor fixes it.

## Why it matters

This is the same family as #349: `inf` and NaN compare False to every ordering operator, so a `<= 0` or range guard downstream does not catch them. `inf` additionally propagates through arithmetic rather than poisoning it visibly, so a policy receives a feature column with one enormous value and no error anywhere.

A zero close is not hypothetical — it is what a venue returns for a malformed or halted candle, and binance's `to_numeric(errors="coerce")` exists precisely because malformed candles happen.

## Scope

- `torchtrade/envs/live/shared/futures_base_obs.py::_default_preprocessing` (all four futures venues) and the alpaca copy in `torchtrade/envs/live/alpaca/observation.py`.
- Options: drop non-finite rows rather than only NaN (`df[np.isfinite(df[feature_cols]).all(axis=1)]`), or refuse a zero close at parse time. The first matches what the trailing `dropna()` was clearly meant to do.
- The offline sampler forward-fills and clips instead of dropping, so it is not exposed the same way — worth checking whether the two should agree.
- No test anywhere asserts observations are finite; a `np.isfinite(...).all()` assertion in `BaseObservationClassTests` would cover all five venues.

BY571:
Fixes #398. Found while adding a finiteness assertion to #397, and verified identical on `main` — so it was not that PR to fix.

## The bug

`_default_preprocessing` computes `feature_close = close.pct_change()`. A candle with `close == 0` — what a venue returns for a halted or malformed bar — makes the **next** bar's pct_change `inf`. The trailing `dropna()` removes NaN and **not** inf, so it reached the policy's observation tensor:

```
market_data_1Minute_5, bar after a zero-priced candle:
  [..., [inf, 0.630, 0.637, 0.624], ...]
```

**inf is the worse of the two.** NaN at least poisons visibly; inf propagates through arithmetic looking like a number — and like NaN it compares `False` to every ordering operator, so no `<= 0` or range guard downstream catches it. #349 was this same family: `nan <= 0` skipped both price fallbacks and returned the NaN *as a price*.

The price consumers were hardened by #349 (`math.isfinite` at three SLTP call sites). The **feature columns** had nothing.

## The fix

Additive in both copies: after the existing `dropna`, drop rows non-finite in any numeric column. Can only drop more rows, never fewer, and covers NaN and inf in one filter.

## Coverage

Nothing asserted observations were finite on **any** venue. The shared harness now does, for all five.

That assertion alone is not enough, and this is the part worth noting: every mock client returns clean candles, so the invariant holds whether or not the filter exists — **removing alpaca's filter failed zero tests** until a test injected the zero-priced bar. Binance's #397 case now asserts `market_data` too, where it previously had to exempt it. Removing either filter fails 1.

Full suite **3916 passed**.

### DIFF
--- a/tests/envs/alpaca/test_obs_class.py
+++ b/tests/envs/alpaca/test_obs_class.py
@@ -6,6 +6,7 @@ Only Alpaca-specific tests (client injection, data integrity, feature semantics)
 and stricter assertions live here.
 """
 
+import pytest
 from unittest.mock import MagicMock
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


BY571:
@CharlieHelps please review.

1. **Is dropping the row the right call, or should a zero close be refused at parse time?** I drop it, matching what the trailing `dropna()` was evidently trying to do. But a zero close means the venue sent something wrong, and silently continuing on the remaining bars is a judgement call — the alternative is to raise and let `ObservationFailurePolicy` handle it. (Related: #394, where that policy turns out to be bypassed on one path anyway.)
2. **The filter is on all numeric columns, not just `feature_*`.** That means a non-finite `volume` or `quote_volume` now also drops the bar, where before only NaN did. Strictly more conservative, but it changes which bars survive for a preprocessing fn that doesn't care about those columns. Too broad?
3. **A custom `feature_preprocessing_fn` can still emit inf** — this only fixes the default. #397 established that copying a rule always drifts; here there's no way to enforce it structurally short of validating at the emission point, which would mean raising mid-episode. I left it. Reasonable?

The finding I'd most want a second opinion on isn't the fix — it's that **removing alpaca's filter failed zero tests** until I injected a zero-priced bar. Every mock returns clean candles, so a finiteness assertion over mock data is vacuous. That's the seventh time this sweep a test couldn't fail; I'd like to know if there's a pattern here you'd guard against differently.

Issue #400, opened 2026-08-19
A burst of malformed candles silently yields an observation shorter than the declared spec


BY571:
Found while reviewing #399. Pre-existing — `dropna` alone could already trigger it on `main` — so it is filed separately rather than folded into that fix.

## What happens

`get_observations()` slices `processed_df.iloc[-window_size:]` with no length check. Preprocessing drops rows (NaN, duplicates, and since #399 non-finite ones), so a burst of malformed candles exceeding the `window_size + 50` fetch buffer silently yields a **shorter array than the declared spec**:

```
56 of 60 candles malformed, window_size=10:
  market_data_1Minute_10  (4, 4)   declared spec (10, 4)
  base_features           (4, 4)   declared spec (10, 4)
```

`env.reset()` and `env.rollout()` both **succeed silently** and emit the short array. `check_env_specs` does catch it — but only against a real observer with corrupted data, and every existing `check_env_specs` test uses `MockObserver`, which bypasses `_default_preprocessing` entirely. So nothing in the suite exercises it.

## Why it matters

A policy receives a window that is not the one it was built for, and a collector or replay buffer pre-allocating from the spec gets a shape it did not expect. This is the silent shape-corruption class the repo guards hard against elsewhere — and the two arrays stay aligned with each other (#397's guarantee holds), so the corruption is uniform and therefore easy to miss.

#399 lowers the threshold: it drops non-finite rows in addition to NaN, so fewer bad candles are needed to reach it. It does not create the hazard.

## Scope

- `BaseFuturesObservationClass.get_observations` and the alpaca copy.
- Options: raise when fewer than `window_size` rows survive (loud, matches `_fetch_single_timeframe`'s existing empty-response `ValueError`); or fetch more and retry. Raising is the smaller change and matches invariant 4.
- Worth adding one `check_env_specs` test that uses a *real* observer with corrupted data — the existing ones cannot see this class of bug at all.

BY571:
Fixed in #401 (merged as part of the observation-failure-path work).

`get_observations` now refuses a frame shorter than `window_size` instead of silently slicing it, in both copies — the shared futures base and alpaca's hand-rolled one. The existing empty-frame check turned out to be the degenerate case of this one, so it was replaced rather than joined.

Two things the review rounds established that are worth recording here:

[...]

PR #401, opened 2026-08-19
fix(live): engage the failure policy on a mark-price read, and refuse a short observation

### LINKED ISSUE #394: ObservationFailurePolicy is bypassed when a flat account's mark-price fetch fails

BY571:
Fixes #394 and #400 — both are the live observation-failure path, and the second is one line from the first.

## #394 — a mark-price fetch failure bypassed `ObservationFailurePolicy`

Reproduced before fixing, on a flat account:

```
HALT     -> RuntimeError: policy BYPASSED
FLATTEN  -> RuntimeError: policy BYPASSED
```

`_halting` catches `(PositionUnknownError, ValueError)` and deliberately **not** `RuntimeError`, because adapters wrap timeouts in it. No adapter raised either type from this call, so whatever the SDK produced escaped and the configured policy was never consulted. Orchestration built around catching `LiveObservationHalt` got an uncontrolled crash instead.

Failing to *read* the mark is the same event as reading an unusable one, and the latter already raises `ValueError` two lines down. So the fetch raises the same way, in the shared base, covering all four venues at once.

**On the safety of converting the type:** `_current_mark_price` itself is *not* flat-gated — five call sites in `binance/env.py`, `bitget/env.py`, `bybit/env.py`, `bybit/env_sltp.py` and `okx/env_sltp.py` fetch the mark with a position open. They are outside `_halting`, so the type change is inert there: an exception propagates uncaught exactly as before, with identical `close_position` counts measured on both sides. The only `_halting`-reachable caller is `_acquire_pre_trade_state`, which passes `position_status` and therefore takes the fetch branch only when flat. `close_position()` *is* invoked under FLATTEN, but every adapter re-reads the venue and returns early when flat, so no market order results.

(An earlier version of this description claimed the branch "only runs when flat", that "bybit and okx don't wrap it at all", and that "FLATTEN has nothing to close". Review measured all three false; the paragraph above is what the measurements actually support.)

## #400 — an observation shorter than the declared spec

`get_observations` slices `iloc[-window_size:]` with no length check. Preprocessing drops rows, so a burst of malformed candles exceeding the `+50` fetch buffer emitted an array shorter than the spec, and `reset()` and `rollout()` both succeeded on it silently.

The existing empty-frame check turns out to be the degenerate case of this one, so `if processed_df.empty` is **replaced** rather than joined. Applied to both copies: the shared futures base and alpaca's hand-rolled one.

## Tests

The #394 test sits at the `_current_mark_price` level next to its sibling, not at env level — the first version asserted `close_position.assert_called()` under FLATTEN, which `close_position_on_init=True` already satisfies at construction. Disabling `_halting`'s flatten dispatch entirely left all six cases green.

All three error shapes are kept because each is a distinct exception family: narrowing the catch to `RuntimeError` fails 2 of 3, to `(RuntimeError, OSError)` fails 1 of 3.

The #400 test goes in `base_exchange_tests.py`, which all five venues inherit. Alpaca's copy had no coverage at all — reverting it alone fails `[malformed-burst]` while `[total-outage]` still passes, which is what proves the new case is load-bearing rather than decorative.

`test_no_live_env_re_forks_the_shared_accessors` now also covers `_current_mark_price` and `_halting`. Verified it catches a re-fork into `base.py`, which the sibling source grep cannot see.

Full suite: **3929 passed, 16 skipped**.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -805,22 +805,29 @@ size, exposure and liquidation distance the env has just admitted it cannot veri
 
 So the env **halts**: it raises `LiveObservationHalt` and produces no transition.
 
-**Which reads this covers.** Only the post-bar state read — the portfolio value and
-observation taken after the bar closes. Every *other* venue read raises its original
-exception with no latch and no emergency flatten:
+**Which reads this covers.** The post-bar state read — the portfolio value and
+observation taken after the bar closes — and the pre-trade state read. Both include the
+observation itself, so one the env refuses to emit (a stale last bar, or one shorter than
+its declared spec) engages the policy like any other unreadable state.
+
+The pre-trade read also covers the mark price: a fetch that fails now raises `ValueError`
+rather than the venue's own type, so `ObservationFailurePolicy` is consulted (#394).
+Before that fix the raw exception escaped and the policy was silently bypassed.
+
+Every *other* venue read still raises its original exception with no latch and no
+emergency flatten:
 
 | read | raises |
 |---|---|
 | position read at the top of `_step()` (position sync, duplicate-action guard) | `PositionUnknownError` |
 | the initial read inside `_reset()` | `ValueError` |
-| `_current_mark_price()`, called at the top of every futures `_step()` | `ValueError` |
+| `_current_mark_price()` at the **sizing** call sites in `_execute_fractional_action` / `_execute_trade_if_needed` | `ValueError` |
 | the sizing balance / portfolio checks | `ValueError` |
 | the candle close that prices SL/TP brackets | `ValueError` |
 
-So `except LiveObservationHalt` alone does **not** give you a policy-driven emergency
-close on an unreadable price or balance — it catches the post-bar read and nothing else.
-Catch `ValueError` and `PositionUnknownError` too if you need to cover those. Alpaca and
-Polymarket do not route through this path at all. Closing the gap is #355.
+So `except LiveObservationHalt` covers the two state reads and nothing else. Catch
+`ValueError` and `PositionUnknownError` too if you need the sizing paths. Alpaca and
+Polymarket do not route through this path at all. Closing the rest of the gap is #355.
 
 ```python
 from torchtrade.envs.core.live import LiveObservationHalt
@@ -847,10 +854,16 @@ as requiring an operator to check the account.
 ### What does not halt
 
 Only failures that mean *the venue told us something impossible about our own money* are
-terminal. A transient error is not: every adapter wraps arbitrary failures in
-`RuntimeError("Failed to get account balance: ...")`, so treating that as terminal would
-end an episode — and under `flatten`, close a live position — on a read timeout. Those
-propagate unchanged, as do configuration errors such as a missing feature key. Retry and
-staleness handling are tracked in #295.
+terminal. A transient error is generally not: adapters wrap arbitrary failures in
+`RuntimeError("Failed to get account balance: ...")`, and treating every one of those as
+terminal would end an episode — and under `flatten`, close a live position — on a read
+timeout. Those propagate unchanged, as do configuration errors such as a missing feature
+key.
+
+The mark-price read is the deliberate exception (#394): there a failed read is
+indistinguishable from reading an unusable price, which was already terminal, so it
+raises the same way. Escalating it is safe because the only halt-wrapped caller reaches
+the fetch solely when the account is flat. Retry and staleness handling are tracked
+in #295.
 
 Alpaca and Polymarket do not route through this path at all (#355).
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps please review.

Two things worth a hard look:

1. **Is converting the exception type the right fix, or should  widen?** The code comment says `NOT RuntimeError: adapters wrap timeouts in it` — a deliberate decision I did not want to overturn. So I converted at the source instead. The alternative is a narrower sentinel exception the venues raise and `_halting` catches by name. Which do you prefer?
2. **Is "flat" really the only branch that reaches this?** My whole safety argument rests on it — if an open position can ever re-fetch the mark, then `FLATTEN` would emergency-close on a transient network blip, which is exactly the hazard that bit me in #399. I traced `_current_mark_price(position_status)`: with a position it reads `position_status.mark_price` and never calls the trader. Please sanity-check that.

Context: this is one of two issues I filed during review rather than fixing inline. The other is #400 (a burst of malformed candles yields an observation shorter than the declared spec) — I'm doing that next rather than leaving it open, since it lives in the same function.

BY571:
@CharlieHelps could you review this one? It closes #394 and #400 — both on the live observation-failure path.

Two things I'd most like a second opinion on, because both are judgement calls rather than mechanics:

**1. Is escalating a short observation to FLATTEN the right call?** Measured, with an open 0.5 BTC long and a malformed burst mid-episode:

```
main:  step SUCCEEDED                 close_position calls: 0
HEAD:  LiveObservationHalt raised     close_position calls: 1
```

I kept that behaviour. My argument: a window shorter than its declared spec is not a usable market view, so it cannot be emitted, and getting flat is what the operator asked for by choosing FLATTEN. What makes it tolerable rather than trigger-happy is that it takes 50 of 60 candles malformed to fire (the `+50` fetch buffer absorbs the rest), duplicate bars never fire it at all, and a config that can *never* fill the window raises at `reset()` while flat with `close_position` at 0 calls. If you think a transient feed problem could realistically close a real position here, that's the thing to push back on.

**2. `except Exception` around the mark-price fetch.** Broad on purpose — the failure shapes span four SDKs and any explicit tuple fails open the first time one adds a type. The cost is that an `AttributeError` or `TypeError` in our own code now reads as a venue outage instead of crashing loudly. I judged the diagnostic cost lower than the fail-open risk; worth a second opinion.

Three review rounds have already run and found real defects in my own work, including a FLATTEN assertion that couldn't fail (satisfied by `close_position_on_init` at construction), an off-by-one that passed the entire suite, and a docs passage that argued against #394 in advance. Those are all fixed — I'm more interested in what the rounds missed than in re-confirming what they caught.

Full suite: 3934 passed, 16 skipped.

Issue #298, opened 2026-08-08
SLTP envs: the documented liquidation-over-bracket priority is unenforced in both offline envs


BY571:
Found during the review of #297 (fix for #292). **Pre-existing on `main`, not a regression** — flagging so it isn't mistaken for coverage that PR added.

## The gap

Both offline SLTP envs implement, and their docstrings state, that liquidation takes priority over an SL/TP bracket when a bar could trigger either:

- `torchtrade/envs/offline/sequential_sltp.py` — `_step` checks `_check_liquidation` before `_check_sltp_trigger`
- `torchtrade/envs/offline/vectorized_sequential_sltp.py::_apply_exit_checks` — the liquidation block runs first and feeds `has_position`

Nothing enforces it. Swapping the two checks in the scalar env leaves the entire offline suite green, because **no test constructs a bar where both would fire**.

## Why it matters

The two produce materially different outcomes on the same bar: a bracket books a bounded loss at the SL price, a liquidation wipes the margin. A refactor that reorders them — or a future exit type inserted between them — changes what a levered position is worth on a gap bar, with no test objecting.

## Suggested shape

A futures case at high leverage where the bar's low breaches *both* the stop-loss and the liquidation price, asserting `action_types[-1] == "liquidation"` and pinning the balance. Worth doing in both envs plus a parity case, since they implement the precedence in structurally different ways (an if/else chain vs mask ordering).

Related: #293 covers a different scalar-vs-vectorized SLTP divergence at high leverage.

PR #402, opened 2026-08-19
fix(sltp): mirror short bracket magnitudes instead of swapping the lists

### LINKED ISSUE #279: SLTP short actions carry the inverse risk/reward of the mirrored long action — intentional?

BY571:
Fixes #279.

Shorts were built as `("short", tp, sl)` — sides right, **magnitudes wrong**. Both halves iterate the same `product(stoploss_levels, takeprofit_levels)`, so the nth short carried the nth long's risk and reward exchanged:

```
long  1: risk 2.5%  reward  5.0%      short 10: risk  5.0%  reward 2.5%
long  3: risk 2.5%  reward 20.0%      short 12: risk 20.0%  reward 2.5%
```

Long stops were {2.5, 5, 10}% and short stops {5, 10, 20}% — **no short action had a 2.5% stop anywhere**. Now `("short", -sl, -tp)`. Measured end-to-end through a real fill, entry 100.40:

| | long SL / TP | short SL / TP |
|---|---|---|
| main | 97.892464 / 105.422653 | **105.422653 / 97.892464** |
| this PR | 97.892464 / 105.422653 | **102.912590 / 95.382401** |

Main's short prices are literally the long row's two numbers swapped. That is #279 in one line.

## Breaking, deliberately

**Retrain — do not reload.** The action-space size is identical in all 128 config combinations swept, so a pre-fix checkpoint loads with no error and trades a different strategy. Equal magnitude *sets* are **not** an exemption: with `stoploss_levels=[-0.05,-0.1]`, `takeprofit_levels=[0.05,0.1]` the old and new maps hold the same four short tuples at **permuted indices** (6 and 7 swap meaning), and the index is what the policy emits. Only a long-only config, or a single stop and target of equal size, leaves the map untouched.

Per @CharlieHelps, a version stamp or action-map checksum is the right long-term answer but spans save/load formats and belongs in its own design change.

## Also in this PR

**A live boundary was open.** `BinanceFuturesSLTPTradingEnvConfig(stoploss_levels=(0.02,), …)` was *accepted* — only the two offline configs validated signs. A positive stop level silently inverts the **long** bracket (stop above entry). `validate_sltp_levels` now holds that boundary for all seven config classes from one place, replacing two hand-rolled copies. `abs()` was rejected: it would immunize the short leg while leaving the long leg inverted.

**Three bracket calculators became one.** `calculate_long_bracket_prices` / `calculate_short_bracket_prices` had statement-identical bodies and existed only to narrate the swap; with no swap they had nothing left to say. Verified a pure refactor across 2250 input combinations — bit-identical, error strings included.

## What the review rounds corrected in my own work

Recorded because the process is the interesting part:

- **A vacuous test.** The first FLATTEN-style assertion pattern recurred: 8 of 14 config-guard cases ran identical bytecode, because all four futures venues inherit `BaseFuturesSLTPConfig.__post_init__`. Replaced with one representative plus an identity guard.
- **Two of three re-derived rows still don't discriminate the bug** — liquidation fires wherever the wrongly-moved stop lands. They earn their place on exit-priority parity (#298/#300), and the comment now says so instead of implying otherwise.
- **A doctest that returned the wrong answer**, and `torchtrade/envs/utils/README.md` asserting in bold the *opposite* of the shipped code while citing #279 as "deliberately open".
- **The checkpoint-compatibility claim took three attempts** — unconditional, then over-generous, then precise. Set equality was the wrong relation; index-wise equality is.

Reverting the fix fails **10**. Full suite: **3975 passed, 16 skipped**. Net **+293/−314** — a semantic fix that removes more than it adds.

### DIFF
--- a/docs/environments/offline.md
+++ b/docs/environments/offline.md
@@ -191,11 +191,19 @@ Extends `SequentialTradingEnv` with **bracket order risk management**. Supports 
 
 ### Configuration
 
+Short actions **mirror** their long counterpart: action *k* short carries the same risk
+and reward as action *k* long, with the stop above entry and the target below. Before
+#279 the two level lists were swapped rather than negated, so short *k* carried long
+*k*'s risk and reward **exchanged** — with the levels below, long stops were
+{2%, 5%} while short stops were {5%, 10%}, and no short action had a 2% stop at all.
+The action-space size is unchanged, so an SLTP checkpoint trained before that fix loads
+without complaint and trades a different strategy. **Retrain rather than reload.**
+
 ```python
 from torchtrade.envs.offline import SequentialTradingEnvSLTP, SequentialTradingEnvSLTPConfig
 
 config = SequentialTradingEnvSLTPConfig(
-    stoploss_levels=[-0.02, -0.05],
+    stoploss_levels=[-0.02, -0.05],   # magnitudes apply to BOTH directions
     takeprofit_levels=[0.05, 0.10],
     include_hold_action=True,
 
--- a/docs/environments/offline.md
+++ b/docs/environments/offline.md
@@ -193,11 +193,9 @@ Extends `SequentialTradingEnv` with **bracket order risk management**. Supports 
 
 Short actions **mirror** their long counterpart: action *k* short carries the same risk
 and reward as action *k* long, with the stop above entry and the target below. Before
-#279 the two level lists were swapped rather than negated, so short *k* carried long
-*k*'s risk and reward **exchanged** — with the levels below, long stops were
-{2%, 5%} while short stops were {5%, 10%}, and no short action had a 2% stop at all.
-The action-space size is unchanged, so an SLTP checkpoint trained before that fix loads
-without complaint and trades a different strategy. **Retrain rather than reload.**
+#279 the levels were swapped rather than negated, so short *k* carried long *k*'s risk
+and reward exchanged. The action-space size is unchanged either way, so a checkpoint
+trained before that fix loads without complaint. **Retrain rather than reload.**
 
 ```python
 from torchtrade.envs.offline import SequentialTradingEnvSLTP, SequentialTradingEnvSLTPConfig
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -473,23 +473,6 @@ class TestCombinatoryActionMap:
         assert sl == -0.02
         assert tp == 0.03
 
-    def test_action_map_short_sign_flip(self):
-        """Test that short actions have flipped signs."""
-        from torchtrade.envs.utils.action_maps import create_sltp_action_map as combinatory_action_map
-
-        action_map = combinatory_action_map(
-            stoploss_levels=[-0.02],
-            takeprofit_levels=[0.03],
-            include_short_positions=True
-        )
-
-        # Action 2 should be the short action
-        side, sl, tp = action_map[2]
-        assert side == "short"
-        # For shorts: SL is above entry (positive), TP is below entry (negative)
-        assert sl == 0.03  # From takeprofit_levels (positive values become SL for shorts)
-        assert tp == -0.02  # From stoploss_levels (negative values become TP for shorts)
-
 
 class TestMultipleSteps:
     """Test multiple environment steps."""
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -286,9 +286,9 @@ class TestBinanceFuturesSLTPTorchTradingEnv:
         with patch.object(env, "_wait_for_next_timestamp"):
             env.reset()
 
-            # Action 5: SHORT with flipped SL/TP
-            # For short: SL above entry, TP below entry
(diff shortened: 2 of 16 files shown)


BY571:
@CharlieHelps could you review this one? It fixes #279 and it is a **deliberate breaking change** to SLTP action semantics, so the judgement call matters more than the mechanics.

**What changed.** Shorts were built as `("short", tp, sl)` — a swap, which gets the sides right (stop above entry, target below) but takes the magnitudes from the opposite list. Both halves iterate the same `product(stoploss_levels, takeprofit_levels)`, so the nth short carried the nth long's risk and reward exchanged, unconditionally:

```
long  1: risk 2.5%  reward  5.0%      short 10: risk  5.0%  reward 2.5%
long  3: risk 2.5%  reward 20.0%      short 12: risk 20.0%  reward 2.5%
```

Long stops were {2.5, 5, 10}% and short stops {5, 10, 20}% — no short action had a 2.5% stop anywhere. Now `("short", -sl, -tp)`.

**The two things I would most like challenged:**

1. **Is negation right, or should it be `abs()`?** Negation assumes the documented sign convention (stoploss_levels negative, takeprofit_levels positive) and there is no validation enforcing it. If a user passes positive stop levels, negation puts the short's stop *below* entry — an inverted bracket, which is a real-money bug. `abs()` would be robust to that but silently accepts a malformed config, which CLAUDE.md's invariant 4 argues against ("validate at the boundary, never guard in the rule"). I chose negation plus no new guard; I am not certain that is the right call.

2. **Breaking semantics with an unchanged action-space size.** 19 actions either way, so a pre-fix SLTP checkpoint loads without complaint and trades a different strategy — no shape error, no warning. Documented in `docs/environments/offline.md` as "retrain, do not reload", but documentation is a weak guard against a silent behaviour change. If you think this warrants something stronger (a config version stamp, a checksum on the action map), say so.

**The finding worth reading in the diff:** three short rows in the scalar/vectorized equivalence table needed re-deriving, but **only one of them actually failed**. The other two kept passing with their stop silently relocated from 115 to 150 — still liquidating, but for the wrong reason. I re-derived all three from intent rather than patching until green.

Reverting the fix fails 10. Full suite: 3960 passed, 16 skipped. Net −16 lines.

BY571:
Round 2 update — @CharlieHelps, two of your framings turned out to matter, and one claim of mine was wrong.

**Your `abs()` question is settled, against `abs()`.** Measured: with `stoploss_levels=[0.02]` the *long* bracket is inverted (stop at 102 above entry 100) under **both** formulations. `abs()` on the short branch alone would make the short silently "work" while leaving that long inversion in place — a guard that absorbs a nonsense config on one leg and not the other, which is exactly CLAUDE.md invariant 4's failure mode. So: negation kept, and the boundary guarded instead.

**The boundary was genuinely open.** `BinanceFuturesSLTPTradingEnvConfig(stoploss_levels=(0.02,), takeprofit_levels=(-0.05,))` was **accepted** before this round, as was alpaca's — only the two offline configs raised. `validate_sltp_levels` now holds it for all seven config classes from one place, replacing two hand-rolled offline copies.

**A claim of mine was false and is now corrected.** I said a pre-fix checkpoint "trades a different strategy." Not unconditional — where every pair has `tp == -sl` the two maps are **byte-identical**:

```
create_sltp_action_map([-0.05], [0.05], include_short_positions=True)
  ->  {2: ('short', 0.05, -0.05)}    on BOTH main and this branch
```

Same for any long-only config, which is all of alpaca. Those checkpoints trade the *same* strategy. Both places that stated it now carry the condition.

**On your second question — whether docs are a strong enough guard for a silent semantic break.** I did not add a version stamp. The action-space size is identical in all 128 config combinations I swept, so nothing would catch a stale checkpoint at load time. I still think a stamp is the right call *if* you want it, but it is a config-format change with its own blast radius, so I did not make it unilaterally.

**Also this round:** the three bracket calculators collapsed into one — `calculate_long_bracket_prices` and `calculate_short_bracket_prices` had byte-identical bodies and existed only to document the swap, so with the swap gone they had nothing left to say (84 lines → 31). That left `side` doing nothing but validation, and deleting the validation kept 861 tests green — so it now has a test, since it is public API and a miscapitalised `"Long"` would price a bracket off the other direction's numbers and return silently.

Full suite: **3978 passed, 16 skipped**. Diff vs main: **+250/−307**.

PR #403, opened 2026-08-19
refactor(live): delete the position_side parameter nothing ever set

BY571:
Refs #289 — closes its `position_side` bullet; see the round-1 comment for the two bullets that stay open.

## The dead parameter

`trade()` and `close_position()` took a `position_side` argument on binance, bitget and the replay executor. **No caller anywhere passes it** — `grep -rn "position_side="` across `torchtrade/`, `tests/` and `examples/` returns nothing — so on binance it was always `"BOTH"` and the four `if position_side != "BOTH"` branches were unreachable. Hedge mode is unimplemented on every venue; the parameter advertised a capability that did not exist.

Checked before deleting, because dropping a positional parameter silently shifts everything after it: no caller passes more than two positional arguments, and `position_side` was fourth.

**okx is deliberately untouched.** It has a real `PositionMode` config and derives `posSide` internally from margin mode (`order_executor.py:218-231`) — an implemented feature, not a dead knob. bitget keeps `position_side_str`, an unrelated local.

## Four of #289's bullets were already closed by earlier PRs

Verified rather than assumed:

| Bullet | State |
|---|---|
| binance `observation.py` doesn't inherit `BaseFuturesObservationClass` (236 LOC parallel impl) | **Done** — inherits now, 80 lines |
| alpaca lacks `close_position_on_init`/`_on_reset` | **Done** — both present |
| vectorized lacks `include_base_features` / pluggable `reward_function` | **Done** — `reward_function` wired at `vectorized_sequential.py:156`; `include_base_features` now *raises* with an explicit #289 message instead of silently doing nothing |
| `log_return_reward` raises on `old_value <= 0` while vectorized clamps to `1e-10` | **Already aligned** — the batched version raises identically (`default_rewards.py:144-148`); the `clamp(min=1e-10)` only stops `log(0)` producing `-inf` before `torch.where` masks it |
| vectorized SLTP has 10 tests vs scalar's 31 | **Stale** — 27 in `test_vectorized_sequential_sltp.py`, plus 58 equivalence and 12 reward-parity tests |

## What this PR does not do, and why

The `margin_type: MarginType` (binance) vs `margin_mode: MarginMode` (bitget/bybit/okx) naming split is left to **#288**. `MarginType` lives in `core/common_types.py` but is used by binance alone, while `MarginMode` is redefined three times locally with venue-specific values (`'ISOLATED'` vs `'isolated'` — API strings that must stay different). Unifying it is a three-way dedup plus a breaking config rename across 79 references, which lands naturally when those files are consolidated. #289's own text says it is "best sequenced AFTER the dedup refactor".

## Post-merge corrections to #402

Its round-3 review finished after the merge and found four things still in live source:

- The `side`-guard docstring said deleting the check left **3974** tests green. It is **3971** — the same commit that measured it also shrank the file, and I did not recompute.
- That file's re-fork guard picked its config with a `next(...)` scan over the module namespace — the exact shape recorded as having silently re-targeted a guard in this repo before. It is `getattr` by name now; verified it still fails `[okx]` when okx declares its own `__post_init__`.
- `calculate_bracket_prices`'s docstring narrated the two deleted functions and called them "byte-identical" — they were statement-identical but differed by two comment lines.
- `validate_sltp_levels` had no parameter annotations while its neighbour did.

`docs/environments/online.md` now states the SLTP sign contract for the live futures envs: they newly raise at construction on levels they previously accepted, and the mirrored short geometry applies to them identically.

Full suite: **3975 passed, 16 skipped**. Net **−9 lines**.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -203,6 +203,13 @@ env = BitgetFuturesTorchTradingEnv(
 
 ### BitgetFuturesSLTPTorchTradingEnv
 
+Stop levels must be **negative** and target levels **positive**; the config raises
+otherwise. The sign is what places each leg relative to entry, and a positive stop would
+put a long's stop *above* entry — an inverted bracket that exits on a favourable move.
+Short actions mirror their long counterpart (action *k* carries the same risk and reward
+either way, #279), so a checkpoint trained before that fix should be retrained rather
+than reloaded: the action count is unchanged but short indices mean something different.
+
 ```python
 from torchtrade.envs.live.bitget import BitgetFuturesSLTPTorchTradingEnv, BitgetFuturesSLTPTradingEnvConfig
 from torchtrade.envs.live.bitget.order_executor import MarginMode, PositionMode
@@ -216,8 +223,8 @@ config = BitgetFuturesSLTPTradingEnvConfig(
     time_frames=["5min", "15min"],
     window_sizes=[6, 32],
     execute_on="1min",
-    stoploss_levels=[-0.025, -0.05, -0.1],
-    takeprofit_levels=[0.05, 0.1, 0.2],
+    stoploss_levels=[-0.025, -0.05, -0.1],   # negative, both directions
+    takeprofit_levels=[0.05, 0.1, 0.2],      # positive, both directions
     include_hold_action=True,
     product_type="USDT-FUTURES",
     leverage=5,
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -203,13 +203,6 @@ env = BitgetFuturesTorchTradingEnv(
 
 ### BitgetFuturesSLTPTorchTradingEnv
 
-Stop levels must be **negative** and target levels **positive**; the config raises
-otherwise. The sign is what places each leg relative to entry, and a positive stop would
-put a long's stop *above* entry — an inverted bracket that exits on a favourable move.
-Short actions mirror their long counterpart (action *k* carries the same risk and reward
-either way, #279), so a checkpoint trained before that fix should be retrained rather
-than reloaded: the action count is unchanged but short indices mean something different.
-
 ```python
 from torchtrade.envs.live.bitget import BitgetFuturesSLTPTorchTradingEnv, BitgetFuturesSLTPTradingEnvConfig
 from torchtrade.envs.live.bitget.order_executor import MarginMode, PositionMode
--- a/tests/envs/test_executor_helpers_288.py
+++ b/tests/envs/test_executor_helpers_288.py
@@ -8,6 +8,7 @@ from torchtrade.envs.live.binance.order_executor import BinanceFuturesOrderClass
 from torchtrade.envs.live.bitget.order_executor import BitgetFuturesOrderClass
 from torchtrade.envs.live.bybit.order_executor import BybitFuturesOrderClass
 from torchtrade.envs.live.okx.order_executor import OKXFuturesOrderClass
+from torchtrade.envs.replay.order_executor import ReplayOrderExecutor
 from torchtrade.envs.live.shared.executor_helpers import (
     ExecutorHelpersMixin,
     TickSizeMixin,
@@ -15,6 +16,10 @@ from torchtrade.envs.live.shared.executor_helpers import (
 
 EXECUTORS = [BinanceFuturesOrderClass, BitgetFuturesOrderClass,
              BybitFuturesOrderClass, OKXFuturesOrderClass]
+
+# The three that carried the dead `position_side` (#289). bybit and okx never had it.
+POSITION_SIDE_FREE = [BinanceFuturesOrderClass, BitgetFuturesOrderClass,
+                      ReplayOrderExecutor]
 
 
 @pytest.mark.parametrize("cls", EXECUTORS, ids=lambda c: c.__name__)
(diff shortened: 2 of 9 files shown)


BY571:
@CharlieHelps could you review this one? It closes #289 and it is mostly a deletion.

**The core change:** `trade()` and `close_position()` took a `position_side` argument on binance, bitget and the replay executor. No caller anywhere passes it, so on binance it was always `"BOTH"` and four `if position_side != "BOTH"` branches were unreachable. Hedge mode is unimplemented on every venue — the parameter advertised a capability that does not exist.

**The one thing I would most like double-checked:** dropping a positional parameter silently shifts every argument after it. I verified no caller passes more than two positional args to `trade()` (and `position_side` was fourth), but that check is only as good as my grep — if you find a call site I missed, especially in a mock or test double, that is a money bug rather than a cosmetic one.

**A judgement call worth challenging:** okx is deliberately untouched. It has a real `PositionMode` config and derives `posSide` internally from margin mode (`order_executor.py:218-231`), so its parameter is an implemented feature rather than a dead knob. If you think the interface should be uniform across venues regardless, say so.

**What I deferred:** the `margin_type: MarginType` (binance) vs `margin_mode: MarginMode` (bitget/bybit/okx) split goes to #288. `MarginType` sits in `core/common_types.py` but serves binance alone, while `MarginMode` is redefined three times locally with venue-specific values (`'ISOLATED'` vs `'isolated'` — API strings that must stay different). That is a three-way dedup plus a breaking rename across 79 references, and #289's own text says it is best sequenced after the dedup.

Four of #289's other bullets turned out already closed by earlier PRs, and one was stale — details in the PR body, all verified rather than assumed.

Also folded in: four post-merge corrections to #402 that its round-3 review surfaced after the merge, including a re-fork guard that picked its subject with a `next(...)` namespace scan — the exact pattern that has silently re-targeted a guard in this repo before.

Full suite: 3975 passed, 16 skipped. Net −9 lines.

BY571:
**Round 1 — two claims in the PR description were wrong, and I'm correcting them here rather than quietly editing them out.**

**1. #289's test-depth bullet is NOT closed.** I said it was stale. Measured: `test_vectorized_sequential_sltp.py` = **27** tests, `test_sequential_sltp.py` = **73**. That is up from the issue's 10/31, but it is still 37%, not parity. The gap is real and it overlaps #274.

**2. The binance bullet had two halves and this PR closes one.** It reads "No `PositionMode`/hedge-mode config surface; `trade()` accepts a `position_side` param nothing ever sets". The dead parameter is gone; the missing config surface is not added — and bitget, bybit and okx all *do* have a real `position_mode` field with reachable behaviour (`bitget/order_executor.py:305,324`; `bybit/:175,274,637`; `okx/:228,329,683`). **Binance is now the only futures venue without one, deliberately**: the decision on this issue was to delete the dead knob rather than build hedge mode across four integrations for a capability nothing uses. Recording it so the gap does not close silently — #288 already lists "hedge-mode surface" among the fixes that landed on some exchanges and not others.

Given both, **#289 should stay open on those two bullets** rather than being closed by this PR. I have changed the PR to reference the issue instead of closing it.

**Also from round 1** (all applied in `55c0fde`):

- The deletion is now verified by **AST parse** rather than grep — zero `position_side=` keywords and zero `trade()` calls with >2 positional args across `torchtrade/`, `tests/`, `examples/`, `docs/`. Independently, a seeded 20-step binance rollout hashes **identically** on both branches with identical call-arg sequences, and 1501 live+replay tests pass identically either side.
- Removed the residue the parameter left: an orphaned `# Add position side for hedge mode` comment and dangling empty `Args:` headers in two `close_position` docstrings.
- Reshaped the #402 re-fork guard again. My replacement for the namespace scan used `venue.capitalize()`, which yields `"Okx"` and so needed a fallback plus a comment explaining itself. It now parametrizes on the imported class — immune by construction rather than by comment.
- Dropped the docs paragraph: it documented an invariant enforced for **seven** config classes but sat under Bitget SLTP alone.

Diff is now **+26/−41**. Full suite: **3975 passed, 16 skipped**.

BY571:
**Round 2 — I made a false claim in this PR and it needs correcting in the open.**

I wrote: *"Hedge mode is not implemented on any venue; the parameter advertised a capability that did not exist."* **That is wrong.**

- **bybit implements it** — `order_executor.py:274` derives `positionIdx` from `PositionMode.HEDGE`, `:637` closes both legs. `pytest tests/envs/bybit/test_futures_order_executor.py -k hedge` → **8 passed**.
- **bitget implements it** — `hedged=(self.position_mode == PositionMode.HEDGE)` at `:305`/`:324`, configurable via `bitget/env_sltp.py:44`.
- **okx** uses `posSide`.

The accurate claim is narrower, and actually stronger: hedge mode is driven by a **venue-level `position_mode` config**, so the per-call `position_side` was a *second, dead surface* for a capability already implemented elsewhere. And binance and replay have **no position-mode concept at all** — `grep PositionMode|position_mode|dualSidePosition` in those two directories returns nothing. That is exactly why those three executors are the ones the new guard covers.

**New guard.** A dead parameter cannot fail a test — a default makes it invisible to every call site that omits it, which is why finding this one took an AST scan rather than a red suite. `test_the_executor_trade_surface_advertises_nothing_it_cannot_do` now pins the parameter names on binance/bitget/replay; re-adding `position_side` to any one fails exactly that venue's case.

Its first version resolved classes via `importlib.import_module(path)` + `getattr(name)` — the same string-resolution shape I had just deleted from the other test file one commit earlier. Parametrized on the imported classes now.

**I dropped its parameter-order assertion after measuring what it was worth**, rather than keeping it because it sounded prudent:

| mutation | caught by existing tests? |
|---|---|
| reorder `side`/`quantity` on **replay** | **yes** — 14 of its own tests fail (callers pass positionally) |
| same reorder on **bitget** | **no** — all 151 pass (callers are all keyword) |

Where the order matters it is already covered; where nothing covers it, every caller is keyword and the order is harmless. So the assertion guarded nothing real.

Also this round: `_reforkable_futures_configs()` flattened to a module-scope list (`parametrize` evaluates at decoration time either way, so the imports were never deferred), three lines of rejected-alternative prose removed from a docstring, and two stray blank lines left inside binance function bodies where the dead branches were cut.

Diff **+47/−43**. Full suite **3978 passed, 16 skipped**. Round 3 running.

PR #404, opened 2026-08-19
refactor(live): one get_observations for all five venues (#288, slice 1)

BY571:
Refs #288 — first slice, biggest first.

## What was duplicated

Alpaca kept a parallel copy of the entire observation window pipeline. Normalising the three naming differences, the two `get_observations` implementations differed by **one real line**:

```
-  df = self._fetch_single_timeframe(timeframe)
+  df = self._fetch_single_timeframe(timeframe, limit=window_size + 50)
```

Everything else in that 64-line method was docstring prose. `_default_preprocessing` differed by two alpaca SDK-shape lines.

**That copy has cost three times running.** Row alignment (#395), the stale last bar (#398/#399), and the short window (#400) each had to be pasted into both halves. Alpaca's half of #397 had **zero** test coverage until #401 added a shared one — deleting its `dropna` failed no tests at all.

## The split, measured rather than assumed

Of the seven abstract hooks on the futures base, the window logic calls only **three** (`_create_client`, `_validate_timeframe`, `_get_timestamp_column`). The boundary follows what the code needs:

| module | holds | lines |
|---|---|---|
| `shared/base_obs.py` *(new)* | the window pipeline; abstract `_fetch_single_timeframe` + the three above | **299** |
| `shared/futures_base_obs.py` | the four kline hooks and the kline-based fetch | 361 → **97** |
| `alpaca/observation.py` | a date-RANGE fetch and the MultiIndex undo | 321 → **109** |

The four futures venues are untouched — same class name, same module, same abstract contract. `timeframes=` stays alpaca's public kwarg and maps to `time_frames` inside `__init__`.

## Three deviations from behaviour-identical, all deliberate, all now tested

Everything else was verified identical by **hashing `get_observations` output** across all five venues, re-driving all four recent fixes with crafted malformed frames, and running a real env rollout — branch vs main.

1. **Length mismatch now raises at construction.** On main, `timeframes=<one TimeFrame>, window_sizes=[10, 20]` is *accepted* and `get_keys()` silently truncates via `zip()` — the policy trains on a window it wasn't configured for.
2. **A daily timeframe is refused at construction**, not on the first fetch — which previously happened during `reset()`, on a live account.
3. **`symbol` is dropped before the shared `dropna`.** A bar whose metadata came back null but whose OHLCV is sound now survives; losing usable bars to a metadata gap is what pushes a window under its spec. @CharlieHelps found the other side of this boundary: two bars alike in every OHLCV field but differing in *symbol* would now collapse to one. That's outside the `symbol: str` single-symbol contract, and it's recorded at the seam as a constraint on any future multi-symbol work.

## What the review rounds found in my own work

- **The test I added to pin deviation 3 could not fail.** It asserted a row count, and with 60 bars fetched for a 5-bar window, dropping one row still fills the window. It asserts the timestamp now, and fails under main's ordering.
- **`df.copy()` in `_normalise_frame` is load-bearing and nothing guarded it.** Preprocessing drops rows `inplace=True`, and the #399 stale-bar check compares the processed frame against the fetched one — share one object and the comparison can never differ, so a stale bar reaches the policy. Returning `df` instead passed all 3999 tests. Now fails 4 of 5 venues.
- **A coverage skew the fold created.** Three tests asserting shared preprocessing lived only in alpaca's file; the other four venues ran that code with no equivalent. Promoted — swapping high/low in the slice now fails 5 of 5 instead of 1.
- **A guard my own change silenced.** `test_no_venue_reimplements_the_shared_base` read one class's `__dict__`, so hoisting `get_observations` dropped it from the guarded set. It walks the MRO now.
- **A 74-line `__main__` block making live Alpaca network calls**, and dead `numpy`/`Dict` imports left by the move.

Net **+515/−558**. Full suite: **4004 passed, 16 skipped**.

### DIFF
--- a/tests/envs/alpaca/test_obs_class.py
+++ b/tests/envs/alpaca/test_obs_class.py
@@ -151,7 +151,7 @@ def test_a_malformed_candle_never_reaches_alpaca_base_features(monkeypatch):
         client=MagicMock(),
     )
 
-    def frame_with_a_hole(timeframe):
+    def frame_with_a_hole(timeframe, limit=None):
         n = 60
         df = pd.DataFrame({
             "symbol": ["BTC/USD"] * n,  # alpaca's default preprocessing drops this column
@@ -202,7 +202,7 @@ def test_alpaca_base_features_survive_a_fn_that_leaves_the_timestamp_in_the_inde
             [["BTC/USD"] * n, stamps], names=["symbol", "timestamp"]
         ),
     )
-    observer._fetch_single_timeframe = lambda timeframe: frame
+    observer._fetch_single_timeframe = lambda timeframe, limit=None: frame
 
     observations = observer.get_observations(return_base_ohlc=True)
     assert observations["base_features"].shape == (10, 4)
@@ -231,7 +231,7 @@ def test_alpaca_refuses_a_fn_that_loses_the_timestamp_entirely():
         client=MagicMock(),
     )
     n = 60
-    observer._fetch_single_timeframe = lambda timeframe: pd.DataFrame(
+    observer._fetch_single_timeframe = lambda timeframe, limit=None: pd.DataFrame(
         {"open": np.full(n, 100.0), "high": np.full(n, 101.0),
          "low": np.full(n, 99.0), "close": np.full(n, 100.5),
          "volume": np.full(n, 10.0)},
@@ -268,7 +268,7 @@ def test_alpaca_drops_a_zero_priced_bar_rather_than_emitting_inf():
             names=["symbol", "timestamp"],
         ),
     )
-    observer._fetch_single_timeframe = lambda timeframe: frame
+    observer._fetch_single_timeframe = lambda timeframe, limit=None: frame
 
     observations = observer.get_observations(return_base_ohlc=True)
     for key, array in observations.items():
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps this is the first slice of #288 and it touches the live observation path — the code #397/#399/#400 just fixed — so **behaviour neutrality is the whole review**.

**What was duplicated.** Alpaca kept a parallel copy of the entire window pipeline. After normalising three naming differences, the two `get_observations` implementations differed by exactly one real line:

```
-  df = self._fetch_single_timeframe(timeframe)
+  df = self._fetch_single_timeframe(timeframe, limit=window_size + 50)
```

Everything else in that 64-line method was docstring prose. That copy has cost three times running — #395, #398/#399, #400 each had to be pasted into both — and alpaca's half of #397 had **zero** test coverage until #401 added a shared one.

**The two things I'd most like challenged:**

1. **Is `_normalise_frame` a real seam or a hook I invented to force the fold?** It exists because alpaca's SDK returns a `(symbol, timestamp)` MultiIndex plus a constant `symbol` column, while `_parse_klines` returns a plain frame. If you think that's papering over a difference that should have stayed as two methods, say so.

2. **One ordering change I believe is safe but cannot prove by types.** Alpaca's preprocessing used to run `reset_index()` → `dropna` → `drop_duplicates` → `drop(columns=["symbol"])`. `_normalise_frame` now drops `symbol` *before* `drop_duplicates`. I claim that's identical because `symbol` is constant for a single-symbol request — but if it ever isn't, this silently changes which rows survive, which is exactly the observation-corruption class this path keeps producing. Worth a hard look.

**Also worth noting:** the refactor found a real bug in a guard. `test_no_venue_reimplements_the_shared_base` read `vars(BaseFuturesObservationClass)` — one class's `__dict__` — so hoisting `get_observations` to the parent silently dropped it from the guarded set. It went red during the refactor, which is the guard doing its job, and it walks the MRO now.

Two things the fold surfaced and I kept as fixes rather than preserving as quirks: alpaca's length check only fired when *both* arguments were lists (so `timeframes=[a,b], window_sizes=5` was accepted and failed later at the spec), and its `_dummy_frame` now carries `symbol` so `get_features` measures a frame shaped like real data.

Net **−83 lines**. Full suite **3985 passed, 16 skipped**.

PR #405, opened 2026-08-20
refactor(live): delete an unreachable hook, and give the margin field one name

### LINKED ISSUE #289: Cross-env parity gaps: binance observation base class + hedge surface, alpaca close_position_on_*, vectorized reward_function/include_base_features

BY571:
Refs #288 (slice 2) and closes #289's naming bullet.

Three deletions and a rename, each measured before touching anything.

## `_get_default_lookback` was unreachable

An abstract hook implemented by **four** venues, called from exactly one place — the `if limit is None` fallback in the kline fetch. Planting `raise AssertionError` at that call site and running the full suite:

```
4004 passed, 16 skipped
```

Nothing takes the branch. `get_observations` always passes `limit=window_size + 50`, and the only argument-free call is alpaca's own override, which ignores it. Hook, branch and four implementations deleted; `limit` is now required on the futures fetch, which is what every caller already did.

## Two smaller ones

- **bitget's 56-line `__main__` block** — the twin of alpaca's, deleted in #404. Duplicated the test suite, needed live credentials to run.
- **`shared/__init__` now exports `BaseObservationClass`** — it became the venue-agnostic entry point in #404 and was reachable only by full path.

## One concept, two spellings (#289)

binance used `margin_type: MarginType`; bitget, bybit and okx used `margin_mode: MarginMode`. Same concept, so exchange-agnostic config code could not set it. Renamed binance's to match across 24 files.

`futures_change_margin_type` — the Binance **SDK's own** method — is deliberately untouched. The `\b` boundary handles that automatically because `_` is a word character, but it is the kind of thing worth checking rather than assuming.

**The rename surfaced something worse.** There are **six** `PositionStatus` dataclasses — one per venue plus replay — and replay's carried *both* `margin_type` and `margin_mode` as separate fields to satisfy both sets of readers. The collision is what turned it into a syntax error rather than a silent duplicate. One field now, and a guard fails if any venue reintroduces the other spelling on a config or a status.

**Values stay venue-specific, and must.** They are API strings: okx's is `cross` where bitget's and bybit's are `crossed`, and binance's is `CROSSED`. A single shared enum cannot carry all four — which is why this is a naming fix, not a consolidation. That distinction is why #289 deferred it here rather than attempting it inline.

## Left for slice 3

**Six `PositionStatus` copies** is the headline — larger than #288's original text describes, and now the main remaining duplication in the live stacks.

Full suite: **4005 passed, 16 skipped**. Net **+111/−156**.

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -183,7 +183,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_mode: isolated
 maintenance_margin_rate: 0.004
 seed: 0
 train_envs: 5
@@ -213,7 +212,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_mode: isolated
 maintenance_margin_rate: 0.004
 seed: 0
 train_envs: 10
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -183,7 +183,7 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
+margin_mode: isolated
 maintenance_margin_rate: 0.004
 seed: 0
 train_envs: 5
@@ -213,7 +213,7 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
+margin_mode: isolated
 maintenance_margin_rate: 0.004
 seed: 0
 train_envs: 10
--- a/examples/online_rl/dqn/env/sequential.yaml
+++ b/examples/online_rl/dqn/env/sequential.yaml
@@ -13,7 +13,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
 maintenance_margin_rate: 0.004
 exp_name: ${logger.exp_name}
 seed: 0
--- a/examples/online_rl/dqn/env/sequential_sltp.yaml
+++ b/examples/online_rl/dqn/env/sequential_sltp.yaml
@@ -17,7 +17,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
 maintenance_margin_rate: 0.004
 exp_name: ${logger.exp_name}
 seed: 0
--- a/examples/online_rl/dsac/env/sequential.yaml
+++ b/examples/online_rl/dsac/env/sequential.yaml
@@ -13,7 +13,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
 maintenance_margin_rate: 0.004
 exp_name: ${logger.exp_name}
 seed: 0
--- a/examples/online_rl/dsac/env/sequential_sltp.yaml
+++ b/examples/online_rl/dsac/env/sequential_sltp.yaml
@@ -17,7 +17,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
 maintenance_margin_rate: 0.004
 exp_name: ${logger.exp_name}
 seed: 0
--- a/examples/online_rl/iql/env/sequential.yaml
+++ b/examples/online_rl/iql/env/sequential.yaml
@@ -13,7 +13,6 @@ bankrupt_threshold: 0.1
 include_base_features: false
 max_traj_length: null
 random_start: true
-margin_type: isolated
 maintenance_margin_rate: 0.004
 exp_name: ${logger.exp_name}
 seed: 0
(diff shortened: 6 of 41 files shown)


BY571:
@CharlieHelps this is slice 2 of #288 plus #289's naming bullet. Three deletions and a rename, and **the rename is the risky part** — it touches a public config field across 24 files.

**What I'd most like checked:**

1. **The rename was done with `sed` across 24 files.** `futures_change_margin_type` — the Binance SDK's own method — survived because `_` is a word character and `\b` doesn't match inside it. That's correct but it's luck-adjacent reasoning, so please audit for anything else that should *not* have been renamed: string literals, doc examples, yaml, comments.

2. **The wire values must be unchanged.** binance's enum values are uppercase (`'ISOLATED'`/`'CROSSED'`), the others lowercase, and okx is `CROSS='cross'` where bitget/bybit are `CROSSED='crossed'`. A rename that altered the string sent to an exchange would be a live-trading bug, not a cosmetic one. That asymmetry is also why this is a *naming* fix and not a shared enum — no single enum can carry all four.

3. **I deleted a fallback.** `_get_default_lookback` was an abstract hook with four implementations and one call site, the `if limit is None` branch in the kline fetch. I verified unreachability by planting `raise AssertionError` there and getting 4004 passed — but "no test reaches it" and "no production path reaches it" are different claims, and the second is the one that matters.

**What the rename surfaced, which I think is the real finding:** there are **six** `PositionStatus` dataclasses — one per venue plus replay — and replay's carried *both* `margin_type` and `margin_mode` as separate fields to satisfy both sets of readers. The collision is what made it a syntax error instead of a silent duplicate; had the names not collided, I'd never have looked. That's slice 3's headline and it's bigger than #288's original text describes.

Full suite: 4005 passed, 16 skipped. Net +111/−156.

BY571:
**Round 2 found a real regression I introduced, on a live-money wire value. Flagging it here since it changes what this PR is.**

Four venues have four *different* margin enums sharing one name, and the values are API wire strings: core/binance `ISOLATED`/`CROSSED`, bitget and bybit `isolated`/`crossed`, okx `cross`. Renaming binance's `MarginType` → `MarginMode` made the package re-exports collide. Measured on the branch:

```
torchtrade.envs.MarginMode       ->  torchtrade.envs.live.bybit.order_executor
torchtrade.envs.live.MarginMode  ->  torchtrade.envs.live.bybit.order_executor
live.__all__.count("MarginMode")  ==  2
```

Two consequences:
- A binance config built from the package export would send **`marginType='crossed'`** to an API expecting `'CROSSED'`.
- `SequentialTradingEnvConfig(margin_mode=MarginMode.ISOLATED)` **raised**, because the imported enum was bybit's while the offline validator compares against core's.

**All 4013 tests passed on that**, because every test imports from venue modules directly and nothing covered the export surface. There's now a guard that fails if either re-export returns.

**What makes this worth reading:** the codebase had already hit this exact hazard and written it down — okx's enum is aliased `OKXMarginMode` with the comment *"to avoid shadowing Bybit's MarginMode/PositionMode."* I renamed a fourth enum into that namespace without reading the comment explaining why the third was aliased.

`torchtrade.envs.MarginMode` is core's again and `torchtrade.envs.live.MarginMode` is bybit's — what both meant before the rename.

**@CharlieHelps — the thing I'd most value your eyes on:** removing those two re-exports is itself a **breaking change to a public namespace**. I checked `examples/`, `docs/` and the READMEs and found no dead import path, and `tests/test_documented_imports.py` executes the README imports and passes. But that is my own grep, and it is the same kind of check that missed the collision in the first place.

Also this round: the vectorized offline config accepted `margin_mode=CROSSED` silently where the scalar raised — a policy would train against isolated liquidation math under a crossed label. Guard added in round 1, and round 2 found **the guard itself was untested** for the config it was written for (deleting it passed all 4013). Both now covered.

Full suite: 4016 passed, 16 skipped.

PR #406, opened 2026-08-20
refactor(live): five PositionStatus dataclasses become one (#288, slice 3)

BY571:
Refs #288, slice 3.

Four venues plus replay each declared their own `PositionStatus`. Measured before touching anything:

| | |
|---|---|
| **binance, bitget, okx** | **byte-identical** |
| replay | `leverage: int` where the venues have `float` |
| bybit | `margin_mode: Optional[str]` where the others have `str` |

Dataclasses don't enforce annotations, so the shared class takes the widest of each and behaviour is unchanged.

**Alpaca is deliberately not folded.** It is spot, and its status is a genuinely different *shape* — `market_value`, `avg_entry_price`, `unrealized_pl` — not this one under other names. A second test asserts that, so nobody folds it by symmetry later.

Each venue module re-exports the shared class, so all 41 existing `from <venue>.order_executor import PositionStatus` sites keep working.

## The guard checks identity, not shape

A field-list assertion would have passed on **all five copies** — three were byte-identical — so it could never catch the duplication it exists to prevent. Verified: re-declaring a byte-identical `PositionStatus` on bybit fails `[bybit]` by name.

## This duplication had already cost

Replay carried **both** `margin_type` and `margin_mode` as separate fields for one concept, to satisfy binance's readers and everyone else's. It surfaced only when #405 collapsed the two spellings and the names collided into a `SyntaxError`. Five copies is how that survives unnoticed — and it's the second time in this slice sequence that unifying names exposed a duplication neither name revealed alone.

Full suite: **4021 passed, 16 skipped**. Net **−13** lines across 7 files.

### DIFF
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -401,27 +401,6 @@ class TestBinanceFuturesOrderClass:
         assert success is False
 
 
-class TestPositionStatusDataclass:
-    """Tests for PositionStatus dataclass."""
-
-    def test_position_status_creation(self):
-        """Test creating PositionStatus."""
-        from torchtrade.envs.live.binance.order_executor import PositionStatus
-
-        pos = PositionStatus(
-            qty=0.001,
-            notional_value=50.0,
-            entry_price=50000.0,
-            unrealized_pnl=0.1,
-            unrealized_pnl_pct=0.002,
-            mark_price=50100.0,
-            leverage=10,
-            margin_mode="isolated",
-            liquidation_price=45000.0,
-        )
-
-        assert pos.qty == 0.001
-        assert pos.leverage == 10
 
 
 
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -401,9 +401,6 @@ class TestBinanceFuturesOrderClass:
         assert success is False
 
 
-
-
-
 class TestBinanceLotSizeRounding:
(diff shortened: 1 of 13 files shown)


BY571:
@CharlieHelps could you review this one? Slice 3 of #288 — five `PositionStatus` dataclasses collapse to one.

**The measurement that justifies it:** binance, bitget and okx were *byte-identical*; replay and bybit differed only in an annotation (`leverage: int` vs `float`, `margin_mode: Optional[str]` vs `str`). Dataclasses don't enforce annotations, so the shared class takes the widest of each.

**Two things worth challenging:**

1. **Is widening the annotations really behaviour-neutral?** Nothing enforces them at runtime today, but if anything downstream introspects `dataclasses.fields(...)[i].type` — a serializer, a spec builder — the widening becomes visible. I grepped and found nothing, but that is the same grep that missed the export collision two PRs ago.

2. **Alpaca is deliberately excluded.** Its status is a different shape (`market_value`, `avg_entry_price`, `unrealized_pl`) rather than the same shape under different names, so folding it would mean renaming spot concepts into futures ones. If you think it should be unified anyway, say so — I've asserted the exclusion in a test, which makes it a decision rather than an oversight.

**The guard checks identity, not shape**, deliberately: a field-list assertion would have passed on all five copies, so it could never have caught what it exists to prevent.

Worth noting why this mattered: replay carried *both* `margin_type` and `margin_mode` as separate fields for one concept, and that only surfaced when #405 collapsed the spellings and the names collided into a `SyntaxError`. Five copies is how a duplicate field survives unnoticed.

Full suite: 4021 passed, 16 skipped. Net −13 lines.

PR #407, opened 2026-08-20
refactor: delete twelve __main__ demo blocks (#288, slice 4)

BY571:
Refs #288, slice 4 — a pure mechanical deletion, deliberately kept in its own PR so it doesn't bury reviewable content.

**549 lines** of `if __name__ == "__main__":` scaffolding across 12 modules, plus three imports they were the only user of.

## Checked before deleting, not after

| | |
|---|---|
| **Nothing invokes them** | `grep` for `python -m torchtrade` / `python torchtrade/` across `docs/`, `README.md`, `examples/`, `.github/` — zero hits |
| **All twelve need the network** | the nine live blocks read API keys and construct real clients; the three offline ones download a HuggingFace dataset |
| **All twelve duplicate the test suite** | each builds an env and steps it, which `tests/envs/` does deterministically with mocks |

## Why they were worth removing rather than fixing

`examples/` is the place for runnable demos, and `tests/test_examples.py` actually executes those. These blocks were the version that *couldn't* be run and therefore couldn't be kept honest — alpaca's `observation.py` block (deleted in #404) still constructed a config with arguments that no longer matched the signature, and nothing noticed for months.

## Verification

Mechanical and per-file: every module is re-parsed with `ast.parse` after the cut and re-imported afterwards. An AST scan then found the three orphaned imports — `TimeFrameUnit` in alpaca's `env.py` and `env_sltp.py`, `load_dotenv` in its `order_executor.py` — which existed only for the deleted code.

Precedent: #404 removed alpaca's observation demo (74 lines, live calls), #405 bitget's (56). This finishes the sweep rather than leaving eleven behind.

Full suite: **4020 passed, 16 skipped**. Net **−573** (+2/−575).


BY571:
@CharlieHelps this is slice 4 of #288 and it's a pure deletion — 549 lines of `__main__` demo scaffolding across 12 modules, plus three imports they were the only user of.

**What I verified before deleting** (rather than after):
- Nothing invokes them: no `python -m torchtrade` / `python torchtrade/` in `docs/`, `README.md`, `examples/` or `.github/`.
- All twelve need the network — the nine live ones read API keys and construct real clients; the three offline ones download a HuggingFace dataset.
- All twelve duplicate `tests/envs/`, which does the same thing deterministically with mocks.

**The one thing worth challenging:** three of these are *offline* modules (`sequential.py`, `sequential_sltp.py`, `onestep.py`, 193 lines between them). They don't need exchange credentials, so there's a case that they serve as quick smoke checks in a way the live ones don't. I judged that the HuggingFace download makes them equally un-runnable in CI, and that `examples/` plus `tests/test_examples.py` already own that role. If you think the offline three should stay, say so — they're separable from the other nine.

**The argument I'd make for deleting rather than fixing them:** these were the version that couldn't be run and therefore couldn't be kept honest. Alpaca's `observation.py` block, deleted in #404, still constructed a config with arguments that no longer matched the signature — and nothing noticed for months, because nothing ever executed it.

Verification is mechanical and per-file: each module is re-parsed with `ast.parse` after the cut and re-imported afterwards, and an AST scan found the three orphaned imports.

Full suite: 4020 passed, 16 skipped. Net −573.

PR #409, opened 2026-08-24
feat(live): surface status staleness and truncate a prolonged outage (#295)

### LINKED ISSUE #295: Live envs: surface exchange-status staleness to the policy and truncate on a prolonged outage

BY571:
Closes #295.

## The issue's premise was wrong about the current code

#295 says *"#270 falls back to the last known cached state, which is the best available truth but is silently stale."* It does not. #270 closed via #296, which added exactly two policies — `HALT` (raise) and `FLATTEN` (close, then raise) — and nothing in `torchtrade/` catches `LiveObservationHalt`.

So an outage did not produce a stale observation. It produced **an uncaught exception that killed the process**. There was no staleness to flag until something was allowed to continue, which is why §1 (`status_unknown`) and §2 (truncation) could not be built independently — you cannot have "the observation says I don't know" without first deciding the env may keep going.

That is a risk-posture decision, so I asked rather than picking it silently. The answer was: grace period, then truncate.

## §3 answered, with a trap

Drove a real env through both `env.rollout` and a real `Collector` emitting `truncated=True` with `terminated=False`. It works — `terminated` survives as 0, so the critic bootstraps rather than being taught the episode ended at zero value.

**The trap:** setting *only* `truncated=True` does **not** end the episode.

```
truncated=True only  ->  done=0 term=0 trunc=1   # collector never resets
done written too     ->  done=1 term=0 trunc=1   # correct
```

TorchRL does not derive `done` from `truncated` in `_step` output, and `truncated` has been write-only since #313 — so nothing in the suite would have caught it. `done` is now written explicitly.

## What shipped

`max_unknown_status_steps` (default **0** = today's behaviour, refuse to act on state you cannot confirm). Above 0:

| bar | `status_unknown` | `done` | `terminated` | `truncated` |
|---|---|---|---|---|
| healthy | `0.0` | 0 | 0 | 0 |
| outage 1 | `1.0` | 0 | 0 | 0 |
| outage 2 | `1.0` | 0 | 0 | 0 |
| outage 3 | `1.0` | **1** | 0 | **1** |

- **Truncates, never terminates** — an unreachable API and a liquidated account must not share a value target.
- **Never fabricates** — a failure with nothing cached raises regardless of budget.
- **`FLATTEN` ignores the budget** — it exists to get you out while blind; riding out the outage would defeat the only thing it is for.
- **`status_unknown` is a separate key**, not a 7th `account_state` slot: `account_state` is shared verbatim with the offline envs, which have no notion of an unreadable venue.
- A shared `_finalize_step_flags` replaces **ten byte-identical done blocks**, which is #288 step 2.

## Driving a real env caught two defects the unit tests could not

1. `status_unknown` was stamped at observation-**build** time, so the grace period served a *cached* observation carrying the healthy `0.0` — the policy was told "all confirmed" on exactly the bars it was not. Worse than having no flag.
2. **Both** read sites advanced the counter, so a budget of 3 truncated after **two** bars. Only the pre-trade read counts now, so the budget means bars.

Both are pinned by `test_a_real_env_flags_the_outage_and_truncates_on_the_budgeted_bar`, which exists precisely because `_halting` in isolation cannot see either.

## Evidence

7 mutants, each killed by the test that claims it: terminate-instead-of-truncate, leave-`done`-to-TorchRL, FLATTEN-honours-grace, lifetime-instead-of-consecutive counting, fabricate-without-cache, bake-flag-at-build-time, count-reads-not-bars.

Suite **4090 passed / 16 skipped**.

## Deliberately not in scope

Alpaca has no `observation_failure_policy` at all (it reads `get_status()` raw at four sites), so its counter never advances and `status_unknown` is always `0.0` there. The key is still declared on all ten envs so the observation contract does not fork by venue. Giving Alpaca a failure policy belongs with #288's parity work.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -918,6 +918,15 @@ order. That last one matters — threading the position and price only fixes the
 balance too, and its cache is seeded from reset's confirmed read so a first trade during
 an outage is still possible.
 
+**One boundary the grace period does not cover.** On binance and bitget the SL/TP
+bracket price comes from a candle close, and that read only happens on a bar that
+actually trades — the hold path returns before it. So after a reset followed by nothing
+but HOLD actions, there is no confirmed close to serve, and the first *trading* bar of an
+outage raises instead of being ridden out. Closing it needs either an extra observer read
+per reset or an explicit change of price source; grace deliberately will not substitute
+the mark, because that silently changes where brackets are priced from. Tracked with the
+#288 parity work.
+
 **This trades on unconfirmed state, which is why it is opt-in.** During the grace period
 the env sizes orders against the last known account; it knows the position it *had*, not
 the one it *has*. `flatten` deliberately ignores the budget: it exists to get you out
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -871,7 +871,47 @@ key.
 The mark-price read is the deliberate exception (#394): there a failed read is
 indistinguishable from reading an unusable price, which was already terminal, so it
 raises the same way. Escalating it is safe because the only halt-wrapped caller reaches
-the fetch solely when the account is flat. Retry and staleness handling are tracked
-in #295.
+the fetch solely when the account is flat.
 
 Alpaca and Polymarket do not route through this path at all.
+
+### `max_unknown_status_steps` — riding out a short outage (#295)
+
+Default `0`, which is the behaviour above: the first unreadable read raises. Set it above
+zero and the env will instead keep stepping on the **last confirmed** read for that many
+bars before ending the episode:
+
+```python
+config = BinanceFuturesTradingEnvConfig(
+    symbol="BTCUSDT",
+    observation_failure_policy="halt",   # grace is a HALT-only concession
+    max_unknown_status_steps=3,          # ride out 3 bars, then truncate
+)
+```
+
+| bar | `status_unknown` | `done` | `terminated` | `truncated` |
+|---|---|---|---|---|
+| healthy | `0.0` | 0 | 0 | 0 |
+| outage 1 | `1.0` | 0 | 0 | 0 |
+| outage 2 | `1.0` | 0 | 0 | 0 |
+| outage 3 | `1.0` | **1** | 0 | **1** |
+
+Three things are load-bearing here:
+
+- **It truncates, never terminates.** Value estimators read `terminated` as "true
+  return-to-go is 0" and `truncated` as "bootstrap from the final observation".
+  Terminating on an unreachable API would teach the critic that a broker outage and a
+  liquidated account carry the same value target.
+- **`status_unknown` is a separate observation key**, not a seventh `account_state` slot.
+  `account_state` is shared verbatim with the offline envs, which have no notion of an
+  unreadable venue; widening it would desync live/offline parity for a concept only one
+  side has. It is emitted by all ten live envs, and is always `0.0` on Alpaca, which has
+  no failure policy.
+- **The budget counts bars, and needs a confirmed read to stand on.** A failure with
+  nothing cached — the first bar of an episode — raises regardless of the budget, because
+  the alternative is inventing an account state.
+
+**This trades on unconfirmed state, which is why it is opt-in.** During the grace period
+the env sizes orders against the last known account; it knows the position it *had*, not
+the one it *has*. `flatten` deliberately ignores the budget: it exists to get you out
+while blind, so riding out the outage would defeat it.
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -911,6 +911,13 @@ Three things are load-bearing here:
   nothing cached — the first bar of an episode — raises regardless of the budget, because
   the alternative is inventing an account state.
 
+**Which reads are covered.** All of them on the trade path: the pre-trade position and
+mark, the post-bar account read, the reset read, and the balance read that *sizes* the
+order. That last one matters — threading the position and price only fixes the decision
+*whether* to trade; a policy that actually opens or resizes during an outage needs the
+balance too, and its cache is seeded from reset's confirmed read so a first trade during
+an outage is still possible.
+
 **This trades on unconfirmed state, which is why it is opt-in.** During the grace period
 the env sizes orders against the last known account; it knows the position it *had*, not
 the one it *has*. `flatten` deliberately ignores the budget: it exists to get you out
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -516,13 +516,19 @@ def assert_the_step_emits_the_whole_done_family(env):
 
     for key in ("done", "terminated", "truncated"):
         assert key in nxt.keys(), (
-            f"{key} missing from the emitted step. The live envs no longer write the done "
-            "family by hand (#313), so a key absent here means the shared full_done_spec "
-            "stopped declaring it."
+            f"{key} missing from the emitted step. The live envs write the done family "
+            "through the shared _finalize_step_flags (#295), so a key absent here means "
+            "an env stopped routing through it."
         )
         assert nxt[key].dtype is torch.bool, f"{key} is {nxt[key].dtype}, not bool"
         assert nxt[key].shape == (1,), f"{key} has shape {tuple(nxt[key].shape)}, not (1,)"
-    assert not nxt["truncated"].any(), "a live env never truncates itself"
+    # Scoped to the DEFAULT config, which disables the grace period. Unscoped this reads
+    # as "a live env never truncates", which #295 made false -- and it would keep passing
+    # only because these fixtures leave max_unknown_status_steps at 0.
+    budget = getattr(env.config, "max_unknown_status_steps", 0)   # absent on alpaca
+    assert budget == 0 and not nxt["truncated"].any(), (
+        "a live env does not truncate itself with the outage budget disabled"
+    )
 
 
 # The two inputs that pre-fix produced a SILENT TRADE rather than a crash, which is what
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -608,3 +608,16 @@ def assert_an_invalid_action_cannot_move_an_open_position(env, action):
     assert env.position.current_position == 1, (
         f"action {action!r} moved the open position to {env.position.current_position}"
     )
+
+
+def wire_outage_state(env, budget=0):
+    """Give a stub the #295 staleness attributes `__init__` would have set.
+
+    Ten sites were pasting these four lines. A stub that skips them fails with an
+    AttributeError from whichever read happens to touch them first, which says nothing
+    about what the test was checking.
+    """
+    env.consecutive_unknown_status = 0
+    env._status_unknown_this_step = False
+    env._last_confirmed_read = {}
+    env._max_unknown_status_steps = budget
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -15,6 +15,7 @@ import numpy as np
 import math
 from unittest.mock import ANY, MagicMock, patch
 from tensordict import TensorDict
+from torchtrade.envs.core.state import position_qty_from_status
 
 
 class TestBinanceFuturesTorchTradingEnv:
@@ -659,7 +660,9 @@ class TestBinanceFractionalPositionResizing:
 
         with patch.object(env, '_execute_fractional_action', return_value=trade_executed) as mock_exec:
             env.position.current_action_level = first_action
-            result = env._execute_trade_if_needed(second_action, current_qty=env._get_current_position_quantity(), current_price=env._current_mark_price())
+            result = env._execute_trade_if_needed(second_action, current_qty=position_qty_from_status(
+                env.trader.get_status().get("position_status")),
+            current_price=env._current_mark_price())
 
             if should_execute:
                 mock_exec.assert_called_once_with(
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -662,7 +662,7 @@ class TestBinanceFractionalPositionResizing:
             result = env._execute_trade_if_needed(second_action)
 
             if should_execute:
-                mock_exec.assert_called_once_with(second_action)
+                mock_exec.assert_called_once_with(second_action, current_qty=None, current_price=None)
             else:
                 mock_exec.assert_not_called()
                 assert result["executed"] is False
--- a/tests/envs/binance/test_torch_env_futures.py
+++ b/tests/envs/binance/test_torch_env_futures.py
@@ -13,7 +13,7 @@ import torch
 from torchrl.envs.utils import check_env_specs
 import numpy as np
 import math
-from unittest.mock import MagicMock, patch
+from unittest.mock import ANY, MagicMock, patch
 from tensordict import TensorDict
 
 
@@ -659,10 +659,12 @@ class TestBinanceFractionalPositionResizing:
 
         with patch.object(env, '_execute_fractional_action', return_value=trade_executed) as mock_exec:
             env.position.current_action_level = first_action
-            result = env._execute_trade_if_needed(second_action)
+            result = env._execute_trade_if_needed(second_action, current_qty=env._get_current_position_quantity(), current_price=env._current_mark_price())
 
             if should_execute:
-                mock_exec.assert_called_once_with(second_action, current_qty=None, current_price=None)
+                mock_exec.assert_called_once_with(
+                second_action, current_qty=ANY, current_price=ANY
+            )
             else:
                 mock_exec.assert_not_called()
                 assert result["executed"] is False
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -9,6 +9,7 @@ import numpy as np
 from unittest.mock import MagicMock, patch
 from tensordict import TensorDict
 
+from torchtrade.envs.core.live import LiveObservationHalt
 from tests.envs.base_exchange_tests import (
     INVALID_ACTIONS,
     assert_an_invalid_action_cannot_move_an_open_position,
@@ -723,7 +724,7 @@ class TestCriticalEdgeCases:
         # "Handle gracefully" used to mean a conditional assertion that ran only if the
         # trade executed -- so it passed whether or not brackets were priced off zero.
         # A zero close now refuses outright, which is the graceful handling (#347).
-        with pytest.raises(ValueError, match="unusable close price"):
+        with pytest.raises(LiveObservationHalt, match="unusable close price"):
             env._execute_trade_if_needed(action_tuple)
         mock_trader.trade.assert_not_called()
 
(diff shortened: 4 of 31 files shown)


BY571:
@CharlieHelps review requested.

Two things I'd most like challenged:

**1. The grace period trades on unconfirmed state.** That is the accepted risk of the chosen posture, but it means that during an outage the env sizes orders against the account it *had*, not the one it *has*. It is opt-in and defaults off. If you think the default should instead be "keep stepping but force hold while blind", say so — that was one of the options considered and rejected as its own kind of lie to the critic.

**2. `status_unknown` is stamped in `_finalize_step_flags`, not at observation-build time.** That is deliberate and was a real bug in my first pass — the grace period serves a *cached* observation, so a build-time flag carried the healthy `0.0` into exactly the bars it should have flagged. But it means the key's value now comes from a different place than every other observation key, which is a wart. If there is a cleaner shape I would take it.

Also worth a look: `max_unknown_status_steps` counts only the pre-trade read, because that runs exactly once per `_step`. A post-bar-only failure sets `status_unknown` but does not advance the budget. I think that is right — the venue being down will fail the next bar's pre-trade read anyway — but it is an asymmetry a reviewer should confirm rather than take on faith.

BY571:
**Round 1 done. Three real defects, and a correction to a claim I made in the PR body.**

### I was wrong about why `done` must be written

The PR body says *"TorchRL does not derive `done` from `truncated` in `_step` output."* **That is false.** `EnvBase._complete_done` fills `terminated | truncated` when the key is absent:

```
done ABSENT    -> {'terminated': False, 'truncated': True, 'done': True}   # derived
done PRESENT=0 -> {'done': False, 'terminated': False, 'truncated': True}  # left stale
```

My original experiment was **confounded**: `_step` had already written `done`, so I was observing the second case and attributing it to the first.

The write stays, for the reason that actually holds — these envs emit the whole family from `_step`, pinned by name in `assert_the_step_emits_the_whole_done_family`, and many tests drive `_step` directly where `_complete_done` never runs. Dropping it broke 13 tests. But it stays with an honest docstring.

### The aliasing bug underneath it

`_last_confirmed_read` cached the post-bar tuple **by reference**, and that tuple carries the observation tensordict the collector then stamps with reward and the done family. So a grace bar re-served *last bar's flags* — which is what made derivation look broken.

That is CLAUDE.md's own rule (`TensorDict.set()` stores references, not copies) and I walked straight into it. Fixed at the cache with a clone.

**It failed zero tests**, so it is now pinned directly rather than left to the `done` write to paper over — removing the clone fails 1 test.

### The crash only moved

`_reset`'s own reads bypassed `_halting`. Truncation fires *because* the outage is ongoing, so the very next `reset()` is the likeliest moment for the venue to still be down — and it raised a bare `PositionUnknownError` one bar later, uncaught, with FLATTEN not acting. The whole point of this PR was to remove that crash; it just relocated it.

`_halting`'s docstring already claimed the reset read was covered. It never was. Now routed through, with **no** `cache_key`: an episode must not *start* on unconfirmed state — the grace period is for continuing one.

### The counter outlived the episode

Nothing cleared it on reset, so a truncated episode left the next one already at budget → truncate on step 1, forever. Shared `_reset_outage_state` plus an AST guard over `RESETTING_ENVS` that accepts `super()` delegation (the SLTP envs override `_reset`).

### Dead code and false prose

- The flatten call inside the grace branch is **unreachable** — `grace` already excludes FLATTEN — and its comment described behaviour that cannot happen.
- `validate_unknown_status_budget` had **zero tests**; deleting its body failed nothing. Now 40 cases across all 8 configs, and the validator is 4 lines instead of 19.
- `getattr(self, "_status_unknown_this_step", False)` was defensive-against-impossible. Initialised in `__init__` instead.
- Four `cache_key != "post_bar"` checks → one `counts` predicate.
- The same 4-line config comment pasted 5×; it belongs in the docs, where it now is.
- `core/base.py` still claimed live `_step` no longer writes the done family.
- `base_exchange_tests` asserted *"a live env never truncates itself"* — which this PR made false. It passed only because the fixtures leave the budget at 0.

Suite **4140 passed / 16 skipped**. Rounds 2–3 next.

BY571:
**Round 2 + @CharlieHelps. Three real defects — one of them the feature not working at all on three venues in four.**

### The grace contract held on one venue in four

Charlie caught this and it is the important one. `binance`, `bitget` and `bybit` called `_get_current_position_quantity()` and `_current_mark_price()` **inside** `_execute_fractional_action` — fresh, unguarded — even though `_step` had already acquired both under the halt policy.

So on a grace bar **with an open position**, those two calls hit the dead venue and raised. No order, no `status_unknown`, no truncation — on exactly the bars the grace period exists to cover. `okx` already threaded them through; the other three now do.

**Every test to that point drove the outage flat**, which is why nothing caught it. There is now a test that opens a position first, and an AST guard that every futures `_step` threads its halted read.

### My round-1 clone fixed the harmless half

I copied *into* the cache. The hazard is on the way *out*: both `return cached` sites handed back the same physical tensordict every grace bar, which `_step` then stamps with reward and the done family — so a collector holding it in two rollout slots reads the later bar's values in **both**.

```
grace bar 1 obs is grace bar 2 obs: True
bar-1 reward after bar 2 is served: 1.0
```

Copy moved to the serve side, and pinned directly: two grace bars must not be the same object.

### The budget could never be spent by a post-bar-only outage

Charlie's second finding. Counting only the pre-trade read meant a healthy pre-trade read reset the counter every bar *while every bar was still flagged unknown* — flagged forever, truncating never. Counting both read sites double-counts a bar where both fail.

The counter now advances **once per bar** in `_finalize_step_flags`, the only thing that runs exactly once per step; `_halting` just records that the bar was unconfirmed. That also let `_halting` stop consulting the budget entirely — "has the outage outlasted its budget" is a question about the *bar*, not about a read.

### Coverage closed

- an **open position** through a grace outage: `account_state` must freeze at the last confirmed read and say so, never be re-derived from a stale mark
- a new episode cannot be served the previous episode's account
- alpaca's `status_unknown` is a third hand-written copy of the same line; pinned as always `0.0`

### One flagged deletion I did *not* take

A reviewer recommended deleting `test_a_recovered_venue_does_not_truncate_the_next_episode` as a dead assertion, having mutation-proved it passed with `_reset_outage_state` gutted. That was true **at the commit reviewed** — but I had since moved the counter out of `_halting`, and under the new design the evidence flips: gutting the body now fails that test, while the AST sibling catches a *removed call*. They are complementary. Re-verified rather than deleted on stale evidence.

### Deletions

`_flatten_if_policy_says_so` was down to one caller after round 1 removed the unreachable one — inlined, taking 8 test stub blocks with it. The validator sweep went 40 → 25 cases on mutation evidence: deleting the call from one plain config fails only its own cases (four independent regression points), while deleting it from `BaseFuturesSLTPConfig` fails all four SLTP configs together — one call site wearing four names, now pinned by identity instead.

Also corrected prose this PR made false: `test_spec_contracts` still said live `_step` no longer writes `truncated`, and a docstring still carried the "TorchRL does not derive `done`" claim I retracted in round 1.

Suite **4132 passed / 16 skipped**. Round 3 next.

BY571:
**Round 3. Two blocking defects, and one of them was a test I destroyed without noticing.**

### I deleted a test with a truncating edit

Consolidating the validator sweep I wrote `s = s[:i] + new`, which drops everything after that index — taking `test_every_futures_step_threads_its_halted_read` with it. The suite went **4146 → 4132** and I attributed the whole drop to the consolidation.

Three source comments then cited that test **by name** as the thing preventing a regression. So round 2's load-bearing safety claim was guarded by nothing, and documented as guarded. Both reviewers found it independently; one proved the consequence:

| un-threaded | result |
|---|---|
| binance | 1 failed |
| **bitget + bybit together** | **4132 passed, 0 failed** |

Two of the three venues this PR "brought to parity" could silently revert to reading a dead venue, with a green suite.

**Fixed structurally rather than by re-adding the guard.** The threaded reads are now **required** keyword arguments — as okx already had them — so omitting one is a `TypeError` on the first step, not a silent fallback. Un-threading now fails 14–19 tests per venue instead of 0.

While there: okx's own `current_qty: float = 0.0` default was the *worse* sentinel — an omitted argument sizes as if **flat**, opening a full target instead of resizing an existing position. Gone.

### The read that SIZES the order was never covered

Threading qty and price fixed the decision *whether* to trade. `_calculate_fractional_position` still called `get_account_balance()` raw, in **all four** venues — so the first grace bar on which a policy actually wanted to open or resize died with a bare error rather than trading on cached state and truncating on budget.

Now under `_halting` with its own cache slot, seeded from reset's confirmed read so a first trade during an outage is still possible.

**Why no test caught it:** every grace test held the action *constant*, so `_execute_trade_if_needed`'s early return fired every bar and the sizing path never ran. The new test changes the action, which is its entire purpose.

### The pattern across all of it

Every defect in this PR — five across three rounds — was invisible because the tests drove the easy state. Flat account, so the open-position path never ran. Constant action, so the sizing path never ran. Pre-trade failures only, so a post-bar-only outage never surfaced. Store-side clone, so the serve side never got exercised. The bugs were never in the code I was looking at; they were in the states I wasn't putting it in.

Suite **4133 passed / 16 skipped**. Diff **+1070/−110**.

@CharlieHelps this has moved a long way since your last pass — both of your findings are fixed and a third of the same family turned up underneath them.

BY571:
**Round 4. The same fix again, on the copies it missed — which is #288's thesis stated as a bug.**

Both reviewers found the same shape independently: **round 3's fix landed on the plain envs and not the SLTP ones, and was tested on one venue of four.**

### The SLTP envs size a bracket off a raw `get_account_balance()`

All four. A separate line, in a separate file, doing exactly what round 3 had just fixed on the plain path. In `trade_mode="fractional"` a grace bar that opened or resized died with a bare `PositionUnknownError` instead of truncating — reproduced live by a reviewer.

**No test drove an SLTP env through an outage at all**, on any venue. There is one now, parametrized across all four.

### bybit and okx SLTP also re-read the mark

Threaded from `_step` now, like the plain envs. binance and bitget price from `base_features` and never had that read — so only a per-venue sweep could see it.

The threaded price is **re-validated where it is used**: threading moved the read out of `_current_mark_price` and took its `isfinite`/`<= 0` guard with it, and a `0.0` divides in the sizing path. An existing test caught that, which is the system working.

### Round 3's own balance fix was pinned on binance only

Reverting it on bitget/bybit/okx failed **zero** tests — the identical shape the round-3 commit message congratulated itself for closing, one commit later.

The test now runs on all four, and fixing it taught me something specific: it had to pick its action **by value**, not by index. binance ships `[-1, 0, 1]` and the others `[-1, -0.5, 0, 0.5, 1]`, so a hardcoded index `2` is *flat* on three of the four and took the close path instead of the sizing path. That is exactly how the fix went untested there — the test ran, passed, and never entered the code it was written for.

### Dead code round 3 created

`_get_current_position_quantity` lost its last production caller when the trade path stopped re-reading. Deleted — along with the `hasattr`-discovered `_SIZING_ENVS` list, which would have silently emptied and **skipped** rather than failed (the "tests that discover their subject" hazard, firing).

The dust-rule test moved to the seam sizing actually reads now. The structural guard against a venue re-forking a private copy **stays** — deleting the shared method makes that more tempting, not less.

`test_position_sizing_refuses_an_unknown_status` is gone with its reason recorded: it guarded an outage beginning *between* the two `get_status()` calls in one step, and threading closed that window by construction.

### Also

`_reset` no longer reaches into `_halting`'s cache — two halted reads instead of a closure plus a manual seed, so `_last_confirmed_read` has exactly one writer. And two more claims round 3 falsified are corrected ("okx alone threads the price", "recomputes qty live" ×2). That is the **sixth** false-prose correction in this PR; each one has been mine.

Suite **4133 passed / 16 skipped**. Diff **+1323/−203**.

BY571:
@CharlieHelps **re-review requested.** Your `CHANGES_REQUESTED` predates four review rounds — both your findings are fixed, and three more of the same family turned up underneath them.

**Your two, fixed:**
- The trade path re-read the venue behind `_halting`'s back on binance/bitget/bybit. Now threaded from `_step` as **required** keyword args (okx's shape), so omitting one is a `TypeError` rather than a silent unguarded read.
- A post-bar-only outage could never spend the budget. The counter moved to `_finalize_step_flags`, the only thing running once per bar; `_halting` now just records that the bar was unconfirmed and no longer consults the budget at all.

**Three more of the same family, found after:**
- The read that **sizes** the order (`get_account_balance`) was still outside `_halting` — on all four plain envs, then again on all four **SLTP** envs, which no test drove through an outage at all.
- bybit/okx SLTP also re-read the mark inside the trade path.
- I deleted the guard test protecting the first fix with a truncating edit, while three source comments cited it by name.

**What I'd most like challenged**, since it is the thing I keep getting wrong:

> Is there any remaining venue read on a trade path that does not go through `_halting`? I have now found four of these in four consecutive rounds — plain envs, then SLTP envs, then the mark on two venues — each time believing I had finished. I would rather you enumerate them than have me claim parity a fifth time.

Also worth your eye: `_reset` seeds the `balance` cache from its own confirmed read, so a grace bar can size a trade it never sized before. A reviewer noted the cached balance is **not** refreshed on a close (the sizing path early-returns before the read), so it can predate current equity by an arbitrary number of bars. That is documented now, but tell me if it should instead invalidate on close.

Suite **4133 passed / 16 skipped**. Diff **+1191/−208** after a trim pass — down from +1323/−203, with no test deleted.

BY571:
@CharlieHelps **Your finding 2 was the most valuable review comment on this PR.** All four addressed; two of them not the way you proposed, and I want to be explicit about which.

### 2. Adapter exception type — you were right, and it meant the feature never worked

```
binance/order_executor.py  raise RuntimeError(f"Failed to get account balance: {e}")
bitget, bybit, okx         identical
```

`_halting` caught `PositionUnknownError` and `ValueError`. So the eight balance reads were **lexically** wrapped and functionally not: a real balance outage escaped every plain and SLTP sizing call. Every test injected `PositionUnknownError`, which only the *status* path raises.

That is the second time in one round I shipped a test asserting a failure shape the venues do not produce — the other was `_reset`, where the adapters **return** `POSITION_UNKNOWN` rather than raising, so wrapping the call caught nothing.

`RuntimeError` is now **grace-eligible only**, re-raised untouched otherwise. That distinction is load-bearing, and #394 is why: a read timeout also arrives as `RuntimeError`, and halting on one would market-close a live position under FLATTEN. Grace doesn't flatten — it serves last-known state and flags the bar — so riding a timeout out is safe where halting on it is not. Both halves mutation-pinned (removing the catch fails 8; removing the re-raise fails 1).

### 4. Stale balance after a close — taken exactly as specified

Invalidated on **success only**; a failed close leaves the position, and with it the equity the cache describes.

### 1. SLTP price — fixed, but wrapped rather than threaded

I did not thread the guarded price into binance/bitget. Those two price brackets off the **candle close**, deliberately (bybit/okx use the mark), so substituting the mark would change behaviour rather than guard it. Instead the observer read *and its verdict* now run inside `_halting`, which closes the escape you describe — the bar flags and truncates instead of raising. Mutation-verified on both venues. **If you think the price source should be unified anyway, say so and I'll do it as its own change** — it's a behaviour change, not a safety fix.

### 3. Reset's second acquisition — deliberately rejected, and I'd like your read

`_reset` returns `_get_observation(advance_hold=False)` unwrapped. That is pinned by an existing test whose parametrize id is literally `reset-is-not-halt-wrapped`:

> reset must NOT route through `_halting`: a metadata gap would then market-flatten on every episode start

`_get_observation` also reads the **observer**, so a window that can never fill is a config error, not an outage — and halting it would close a real position every time an episode starts. I wrapped it, saw that test fail, and reverted with the reasoning recorded in source.

Your alternative — *"build reset observations from the already acquired snapshot"* — avoids the dilemma entirely by removing the second read rather than wrapping it. I think that's the right answer and it's a bigger refactor than this PR should carry. **Do you agree it belongs as a follow-up, or do you consider it blocking?**

---

Suite **4151 passed / 16 skipped**. Diff +1420/−287.

Worth saying plainly: this is the **fifth** consecutive round finding "the fix landed on some copies and not others" — plain envs but not SLTP, four venues in code but one in tests, two venues threaded and two not, and now an exception type that only one read path produces. That is #288's thesis, and it is becoming an argument for doing the dedup before finishing this.

BY571:
@CharlieHelps All four fixed, with the four regression tests you named. Thank you for the strategy call as well — I'll open #288 immediately after this merges rather than folding it in.

**1. You were right that my previous fix was cosmetic.** `_halting(read_close)` had no `cache_key`, so `cached` was `None`, grace could never apply, and binance/bitget still raised — I'd changed the exception type and nothing else. Now `cache_key="candle_close"`, its own slot, keeping the candle-close source as you advised.

**One residual boundary I'm flagging rather than hiding:** the hold path early-returns before that read, so a policy that holds for N bars and *then* trades during an outage still has no confirmed close to serve. The test seeds via a trading bar and says so in its docstring. Closing it means either an extra observer read per reset or carrying the close through `_step` — tell me if you want that in this PR.

**2. Reset.** Your narrower framing was the right one: `_get_observation` now takes an optional `snapshot` and reset passes the pair it already confirmed. Not wrapped — `reset-is-not-halt-wrapped` is correct and stays.

**3. SL/TP closure.** Caught. That path has no `close_position()` call for the close-site guard to find, because the env never asked for the close — but the realised P&L moved equity all the same.

**4. The abort-path flag.** Also caught: `_status_unknown_this_step` was set before the re-raise and consumed only by `_finalize_step_flags`, which that path never reaches. A caught-and-retried timeout would have flagged the next healthy bar and counted it toward truncation.

Each fix mutation-pinned: reverting them fails 4 / 1 / 1 / 1 tests respectively.

---

**On the parity misses** — your read matches mine, and the evidence got worse while you were writing. Between your last two comments I fixed the balance-cache invalidation in `_handle_close_action`, reported it done, and had missed **six** other close sites in three code shapes. Your inline had *enumerated* them; I acted on the summary. That is now seven consecutive rounds of the same failure, one of them committed while fixing an instance of it.

So I've stopped adding behavioural tests for that class and started adding **structural** ones — `test_every_successful_close_invalidates_the_cached_balance` walks every `close_position()` call and requires an invalidation in the same function, with two exemptions justified by ordering. That is the only thing that has actually caught a missed copy before you did.

Suite **4160 passed / 16 skipped**. Diff +1629/−293.

PR #410, opened 2026-08-25
refactor(288): fold the four futures base.py files into the shared base

BY571:
Slice 1 of #288. **Net −109 lines** (four `base.py`: 524 → 246), suite 4184 green.

`base.py` was **66–96% identical** after exchange-name substitution, in two tight clusters:

| | binance | bitget | bybit |
|---|---|---|---|
| **bitget** | 90% | | |
| **bybit** | 71% | 67% | |
| **okx** | 70% | 66% | **94%** |

524 lines across four files → 336, plus ~90 shared. Each venue's base is now its two class attributes, a typed `__init__` that forwards, and a `_trader_kwargs` override naming only what is genuinely venue-specific.

## The credential shape was never the blocker I recorded

I had this issue marked stuck on exactly one thing: okx and bitget take a `passphrase` and binance and bybit don't, so folding looked like a choice between `**credentials` (losing the typed signature) and an optional argument two venues ignore — the dead-parameter defect #289 deleted.

**The passphrase never reaches anything except the trader.** A `_trader_kwargs()` hook needs neither horn. And once it existed it absorbed `trade_mode`, `position_mode` and `product_type` as well — which vary across *different, overlapping* venue sets, a problem I hadn't counted when I called it blocked.

```python
def _trader_kwargs(self, api_key, api_secret):
    return {**super()._trader_kwargs(api_key, api_secret),
            "position_mode": self.config.position_mode,
            "passphrase": self._passphrase}
```

## What I deliberately did not normalise

**`TRADER_FIRST`.** bybit and okx build the trader first so the observer can reuse its client; binance and bitget build the observer first. Both constructors talk to the exchange, so that order decides which side effects have already landed when the other one fails. Preserving it costs one class attribute; normalising it would be a real behaviour change smuggled into a refactor.

## Verification

**A correction: the "IDENTICAL constructor kwargs" proof originally claimed here was invalid.** The probe patched one class at a time, so on the `TRADER_FIRST` venues the real trader constructor ran first, hit the network and threw — the observer was never reached. It recorded `"observer": "ERR ValueError"` on both sides, and I diffed two identical *error strings*.

It hid a live-money regression: okx's observer was handed the trader's `Trade.TradeAPI`, which has no `get_candlesticks`. Fixed, and the real verification is now a test rather than a script — `test_init_trading_clients_wires_each_venue_as_before` runs the refactored method with recording fakes, network-free, asserting construction order and complete kwargs per venue.

Measured against three wiring regressions, old approach vs new:

| mutation | new | old |
|---|---|---|
| ignore `TRADER_FIRST` | 2 failed | **0 failed** |
| observer built from trader kwargs | 4 failed | **0 failed** |
| dependency injection ignored | 1 failed | **0 failed** |

## Three structural guards fired — updated, not weakened

- **`execute_on`** — now admits `_finish_futures_init` **by name**. Admitting the whole shared module would restore the MRO hole the guard exists to close (`TorchTradeLiveEnv` lives there too, and initialises `execute_on = None`).
- **bankruptcy baseline** — accepts delegation to the shared tail; its load-bearing half (a re-forked balance *read*) is untouched.
- **close-site invalidation** — exempts `_finish_futures_init`: its startup flatten runs at construction, before any read is cached, so there is nothing to invalidate. Same ordering argument as `_reset`.

## Scope

Deliberately slice 1 only, per @CharlieHelps on #409: *"prioritize #288 immediately after this; expanding this already-large safety change into the full refactor would make the invariants harder, not easier, to review."*

Remaining: `env_sltp.py` incl. the 70-line config dataclass ×4 (~900), `env.py` trade-path helpers (~500), the verbatim executor helpers, and routing more per-venue tests through `base_exchange_tests`.

### DIFF
--- a/tests/envs/bitget/test_obs_class.py
+++ b/tests/envs/bitget/test_obs_class.py
@@ -85,13 +85,6 @@ class TestBitgetObservationClass(BaseObservationClassTests):
                 symbol="BTC/USDT:USDT", time_frames=TimeFrame(2, TimeFrameUnit.Minute),
                 window_sizes=10, client=mock_ccxt_client)
 
-    def test_product_type_demo(self, mock_ccxt_client):
-        """demo=True sets the product type."""
-        observer = BitgetObservationClass(
-            symbol="BTC/USDT:USDT", time_frames=TimeFrame(1, TimeFrameUnit.Minute),
-            window_sizes=10, product_type="USDT-FUTURES", demo=True, client=mock_ccxt_client)
-        assert observer.product_type == "USDT-FUTURES"
-
     def test_get_keys_multi(self, observer_multi):
         assert observer_multi.get_keys() == ["1Minute_10", "5Minute_20", "1Hour_15"]
 
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -3983,3 +3983,100 @@ def test_no_venue_redeclares_the_account_state_contract(env_cls):
         f"shared contract on TorchTradeLiveEnv; a reordered copy is a permuted "
         f"observation that no test compares against anything else"
     )
+
+
+@pytest.mark.parametrize("venue,expect_order,observer_extras", [
+    ("binance", ["trader" if False else "observer", "trader"], set()),
+    ("bitget",  ["observer", "trader"], set()),
+    ("bybit",   ["trader", "observer"], {"client"}),
+    ("okx",     ["trader", "observer"], set()),
+], ids=["binance", "bitget", "bybit", "okx"])
+def test_init_trading_clients_wires_each_venue_as_before(
+    venue, expect_order, observer_extras
+):
+    """Characterisation of the code #288 actually refactored, network-free.
+
+    The previous test called `_observer_kwargs()` directly on a `__new__`'d instance, so
+    it never ran `_init_trading_clients` -- the method that selects the order, constructs
+    both classes and consumes both hooks. A mutation of a hook failed it while a wiring
+    regression at any of the eight call sites passed. Same error as the kwargs probe in
+    the PR body: verifying something adjacent to the thing that changed.
+
+    Construction ORDER is asserted, not just kwargs. bybit's observer reuses the trader's
+    session, so it must be built second; okx must NOT, because its trader client is a
+    Trade API and klines live on the MarketData one -- the round-1 regression.
+    """
+    import importlib
+
+    module = importlib.import_module(f"torchtrade.envs.live.{venue}.env")
+    cls = next(v for k, v in vars(module).items()
+               if k.endswith("TorchTradingEnv") and v.__module__ == module.__name__)
+
+    log = []
+
+    class FakeObserver:
+        def __init__(self, **kw):
+            log.append(("observer", kw))
+
+    class FakeTrader:
+        def __init__(self, **kw):
+            log.append(("trader", kw))
+            self.client = "TRADER_SESSION"
+
+    env = cls.__new__(cls)
+    env.config = SimpleNamespace(
+        symbol="X", time_frames=["1m"], window_sizes=[10], demo=True,
+        leverage=5, margin_mode="isolated", position_mode="one_way",
+        product_type="COIN-FUTURES", trade_mode="fractional",
+    )
+    env._feature_preprocessing_fn = None
+    env._passphrase = env._api_passphrase = "PASS"
+
+    with patch.object(cls, "OBSERVER_CLS", FakeObserver), \
+         patch.object(cls, "TRADER_CLS", FakeTrader):
+        env._init_trading_clients("KEY", "SECRET", None, None)
+
+    assert [name for name, _ in log] == expect_order, (
+        f"{venue} built {[n for n, _ in log]}, expected {expect_order}"
+    )
+
+    kw = dict(log)
+    base_obs = {"symbol", "time_frames", "window_sizes", "feature_preprocessing_fn", "demo"}
+    assert set(kw["observer"]) == base_obs | observer_extras, (
+        f"{venue} observer kwargs {sorted(kw['observer'])}, expected "
+        f"{sorted(base_obs | observer_extras)}"
+    )
+    if "client" in observer_extras:
+        assert kw["observer"]["client"] == "TRADER_SESSION", (
+            f"{venue} shares the trader's session, so it must receive that object"
+        )
+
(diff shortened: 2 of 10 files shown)


BY571:
@CharlieHelps review requested. This is slice 1 of #288, kept deliberately small per your note on #409.

**What I'd most like checked**, given this is a refactor claiming zero behaviour change across four venues — the exact shape that has bitten this codebase repeatedly:

1. **Is the equivalence proof sound?** I patched `__init__` on all 8 observer/trader classes, captured the kwargs each venue passes, and diffed against a worktree at `main` — identical. But that only covers *construction*. If there is a behavioural difference I could not see that way, I'd rather hear it now.

2. **`TRADER_FIRST`.** I preserved per-venue construction order instead of normalising it, on the grounds that both constructors hit the exchange and the order decides which side effects have landed when the other fails. Is that over-cautious? Normalising would delete the attribute and one branch.

3. **The three guards I updated.** Each fired because the code they pinned moved. I believe I widened them exactly as far as the move requires and no further — particularly `execute_on`, where I admitted `_finish_futures_init` *by name* rather than admitting the shared module, because the latter would reopen the MRO hole the guard was written to close. Please check I didn't quietly weaken any of them.

4. `_finish_futures_init` is called by each venue's `__init__` rather than by the base. That is deliberate — the base's `__init__` runs before the venue has set `_feature_preprocessing_fn` — but it means a fifth venue could forget the call. Worth a structural guard, or is that over-engineering?

BY571:
**Round 1: a live-money regression, and my equivalence proof was empty.**

## Correcting the PR body

It says *"IDENTICAL constructor kwargs on all 8 classes"*. **That is false.** My probe patched one class at a time, so on the `TRADER_FIRST` venues the real trader constructor ran first, hit the network, and threw — the observer was never reached:

```
"okx.env": { "observer": "ERR ValueError", "trader": [...] }
```

I diffed two identical **error strings** and reported equivalence. The claim is in the PR body and a commit message; both are wrong.

## The regression it hid

`okx`'s observer was being handed the **trader's** client. I copied bybit's `_observer_kwargs` override — bybit shares one pybit session between both roles, okx does not:

```
TradeAPI  has get_candlesticks: False
MarketAPI has get_candlesticks: True
```

Construction succeeds; the first bar raises `AttributeError`. **The okx live env could not fetch a single observation.** Both reviewers found it independently.

## Why nothing caught it

83 env constructions in the suite inject an observer. 2 don't — and those need live credentials, so they skip. **`_observer_kwargs` was unreachable by the entire test suite** — the exact code path this PR exists to change. There is now a per-venue test asserting the observer extras as a contract (bybit shares a client; the other three must not), mutation-verified against all three failure modes.

## A reviewer corrected my fix

bitget's observer *did* stop receiving `product_type` — and it never **read** it. The observer stores the attribute; only the order executor uses one. So restoring the argument fed a dead parameter, and my comment about "reading candles from the wrong product line" was false. Deleted the parameter instead, plus the attribute-echo test pinning it.

## A guard I claimed I hadn't over-widened, had been

I wrote above that I'd widened the `execute_on` guard *"exactly as far as the move requires and no further."* Mutation-proven wrong: deleting bybit's `_finish_futures_init()` call still **passed**, restoring the precise MRO hole the guard's own comment says version 1 had. Widening the **predicate** instead — same mutation now fails 3 tests.

---

**Diff +226/−288 (net −62).** Suite 4163 green.

@CharlieHelps re-review welcome. Given the above, the question I'd put to you is not "is this correct" but: **is a refactor that changes eight constructor call sites reviewable at all when the suite cannot reach the code being changed?** The DI-everywhere test convention made this fold effectively untested, and I only learned that from a reviewer who instrumented it rather than reading it.

BY571:
@CharlieHelps Blocker fixed, and your diagnosis was exact.

**You were right that my test verified the wrong thing.** It called `_observer_kwargs()` on a `__new__`'d instance and never ran `_init_trading_clients` — the method this PR actually refactored. Measured against three wiring regressions:

| mutation in `_init_trading_clients` | new test | old test |
|---|---|---|
| ignore `TRADER_FIRST` | 2 failed | **0 failed** |
| observer built from trader kwargs | 4 failed | **0 failed** |
| dependency injection ignored | 1 failed | **0 failed** |

The old test caught **none** of them. It was the same error as the kwargs probe in the PR body, one layer up: verifying something *adjacent* to the thing that changed.

`test_init_trading_clients_wires_each_venue_as_before` now runs the real method with recording fakes patched over `OBSERVER_CLS`/`TRADER_CLS`, network-free, across all four venues — asserting construction **order** and complete kwargs. bybit receives `client=self.trader.client`; binance, bitget and okx do not. A second test pins that dependency injection still skips construction entirely, which 83 of the suite's 85 env constructions depend on and nothing asserted directly.

**PR body corrected** — the invalid proof is replaced with the mutation table above, and the figures updated (net **−109**, four `base.py` 524 → 246, suite **4184**).

Also landed since your round-1 review: `ACCOUNT_STATE` was declared **five times**, identical, alpaca included. It is the observation contract itself, so five copies means a venue could reorder an element with every test still green — each env is only ever compared against its *own* list, and what breaks is a trained checkpoint at a different index. Hoisted to `TorchTradeLiveEnv` and guarded.

---

One thing I'd value your read on, since it is the through-line of both this PR and #409: **the DI-everywhere test convention made the refactored path structurally unreachable.** 83 constructions inject an observer; the 2 that don't need live credentials and skip. That is why a regression shipped and why my first two attempts at proving equivalence both measured something else. Is the right follow-up a convention that every shared constructor path gets one non-DI characterization test, or does that belong in the #288 test-consolidation slice?

BY571:
@CharlieHelps Pinned exactly, and your four mutations now all fail:

| mutation | before | now |
|---|---|---|
| binance drops `trade_mode` | passed | **1 failed** |
| bitget drops `product_type` | passed | **1 failed** |
| okx drops `passphrase` | passed | **1 failed** |
| bybit changes `position_mode` | passed | **2 failed** |

You caught the same defect one field over: I made the **observer** expectation exact and left the **trader** one a `<=` subset check, in the same test. A subset assertion cannot pin what *varies* — which is the only thing `_trader_kwargs` exists to do. Exact per-venue dicts now.

**Convention taken as you framed it**, narrower than mine: *every orchestration path that normal tests bypass through dependency injection needs one network-free characterization test at the real production seam*. Recorded as its own slice on #288 rather than folded into test-consolidation, with your reasoning that the failure mode is **path reachability, not deduplication** — and with the concrete evidence from this PR attached, including that two of my own equivalence proofs measured the wrong thing.

Suite **4184** green. Ready for another look.

PR #411, opened 2026-08-25
refactor(288): fold the four SLTP _step and _reset copies into the mixin

BY571:
Slice 2 of #288. **Net −78**, suite 4184 green.

`_step` was **100% identical within each venue pair and 88% across** — and the 12% was a single line, the dispatch call. `_reset` was 100% identical across all four.

| method | binance≡bitget | bybit≡okx |
|---|---|---|
| `_step` | 100% | 100% |
| `_reset` | 100% | 100% |
| `_execute_trade_if_needed` | 96% | 94% |

`env_sltp.py`: 389/400/321/327 → **316/327/268/276**.

## Why this file specifically

This is where #295 kept finding the same fix applied to some copies and not others — the unguarded balance read, the re-read mark, the close that left the cache stale. Four copies of the money path is the reason.

`_dispatch_sltp_trade` owns the only genuine difference: bybit and okx take the threaded mark (re-reading it inside the trade path bypassed the halt policy, #295); binance and bitget price brackets off a candle close and take the default. One hook instead of four `_step` copies differing by a line.

## Three behaviours of the folded code had no test

Found by mutating it. Each failed **zero** tests before:

| mutation | before | after |
|---|---|---|
| reward zeroed (placeholder never overwritten) | 0 failed | **4 failed** |
| `position_closed` not propagated | 0 failed | **4 failed** |
| `_reset_sltp_state()` removed | 0 failed | **4 failed** |

The gaps predate the fold, but one copy means one break now reaches all four venues.

## The fold was silently incomplete, and nothing noticed

I ran `git checkout` on bybit's file during mutation testing and it kept its own `_step`/`_reset`. **The suite stayed green** — of course it did, the copy still worked. An incomplete fold is invisible behaviourally and stays invisible right up until a shared fix fails to land on the venue that never folded. `test_no_sltp_env_reforks_the_shared_step` fails on exactly that.

## Two guards followed the code rather than being widened

- The pre-trade halt guard now resolves `_step` **through the MRO** instead of reading the venue's file. Strictly stronger: a venue that re-forks `_step` without the halted read still fails, because `getsource` then returns *its* copy.
- `test_an_outage_stops_the_step_before_it_can_trade` discovered its subject with `"_step" in vars(c)` and silently stopped finding it once `_step` moved — pytest reported `StopIteration`, not a missing guard. Discovers by name now.


BY571:
@CharlieHelps review requested. Slice 2 of #288, kept to one file's worth of folding per your guidance on #410.

**What I'd most like attacked, in priority order:**

1. **Is `_dispatch_sltp_trade` the right seam, or am I encoding an accident?** bybit/okx take the threaded mark; binance/bitget price brackets off a candle close. That difference is real today, but it exists because #295 fixed the mark re-read on two venues and left the other two alone (they had a different read to fix). If the price source *should* be unified, this hook is cementing a split that ought to close — you told me on #409 not to unify it in that PR, so this is the follow-up question landing.

2. **Three behaviours of the folded `_step` had zero test coverage** — reward, `position_closed`, `_reset_sltp_state`. I found them by mutating the freshly-shared method and pinned all three. Are there others in that method I haven't thought to mutate? It is 40 lines of money path that until now no test constrained beyond "it runs".

3. **The two guards I moved rather than widened.** The pre-trade one now resolves `_step` through the MRO instead of reading the venue's file. I believe that is strictly stronger, but on #410 I made exactly this claim about a guard and was wrong — mutation-proven. Please check it independently.

4. `test_an_outage_stops_the_step_before_it_can_trade` discovered its subject via `"_step" in vars(c)` and silently stopped finding it when `_step` moved — pytest reported `StopIteration`, not a missing guard. I fixed that instance. **Are there other tests keyed on where a method lives rather than what it does?** That pattern is in my notes as a recurring hazard and I would rather sweep it than hit it a third time.

One process note, since it bears on how much to trust the green suite here: I reverted bybit's file with `git checkout` mid-mutation-testing and the fold silently stayed incomplete — bybit kept its own `_step`, and **all 4184 tests still passed**, because a private copy works fine right up until a shared fix fails to land on it. `test_no_sltp_env_reforks_the_shared_step` closes that, but it is worth knowing the suite could not tell.

BY571:
**Round 1. Runtime equivalence proven; everything around it was broken.**

The fold itself holds up — **42,668 lines of state** compared across 4 venues × 5 configs × 3 entry states, **zero diff**, plus an AST comparison showing 18 statements in the same order with only the intended dispatch hook differing. The damage was all collateral.

## Two discovery lists silently dropped all four SLTP venues

`STEPPING_ENVS` and `RESETTING_ENVS` filter on `"_step" in c.__dict__` — and this fold is exactly the event that makes that false.

```
STEPPING_ENVS  10 -> 6
RESETTING_ENVS  7 -> 3
```

Gone with them: the **#295 outage-truncation guard**, the action-validate-before-trade ordering guard, and the done-family guard — on every SLTP venue at once.

**And the collected test count did not move.** 795 == 795, because this PR's 12 new cases happened to offset the 12 that vanished. Proven live: a hand-written `truncated` alongside the shared writer — the exact defect the done-family guard rejects, now pinning truncation to `False` for all four venues — **passed all 4184 tests**. It fails 4 now.

Discovery is by *resolution* now, with the length assertions those lists never had. Without them an emptied list **skips** rather than fails, which is how this would have shipped.

## Alpaca broke, and it is not in this PR's diff

It also inherits `SLTPMixin`, so `super()._reset()` now lands on the mixin's copy first and `_reset_sltp_state` ran **twice per reset**. Idempotent today; a real bug the moment that method gains a counter, a log, or a `cancel_open_orders()`. Its `_reset` was byte-identical to the mixin's — deleted.

## Two invariant guards were left scanning only the venue files

`position_status.qty` and `trade_info["side"] ==` are policed by substring over `live/<venue>/env_sltp.py`. The code they police now lives on the mixin, so reintroducing either pattern there **passed 20/20 cases**. They scan the mixin too now.

---

**Diff +243/−305 (net −62).** Suite **4199** — up from 4184, because 12 guards are running again.

@CharlieHelps the question I put to you earlier now has evidence behind it. Every finding in this round is the same shape: **a refactor moves code, and the things that were watching that code go quiet without failing.** Not one of them showed up as a red test. The count stayed identical, the guards stayed green, and a mutation that kills the truncation channel on four venues sailed through the whole suite.

Is there a general defence here beyond "add a length assertion to every discovery list", or is per-list vigilance genuinely the state of the art?

BY571:
@CharlieHelps Both blockers fixed. Your structural-guard finding was already addressed in the round-1 push you reviewed just before.

## The guard I called "strictly stronger" was still a substring check

That is the sharper half of your review. Resolving through the MRO fixed *where* it looks — it still only asserted that the text appears. Your two mutations:

| mutation | before | now |
|---|---|---|
| commented-out `self._acquire_pre_trade_state()` | passed | **4 failed** |
| real call moved **after** the dispatch | passed | **4 failed** |

It uses `_first_call_position` now — the AST call-position helper that has been two hundred lines up in that same file since #408, comparing acquisition against the dispatch. I built that tool precisely because `ast.walk` order isn't source order, then reached for `in source` anyway one file over.

## The dispatch seam had no production-routing test

Mutating bybit's override to call `self._current_mark_price()` instead of forwarding the threaded value left **237 tests green** — and that mutation *is* the unwrapped mark re-read #295 and #409 spent two rounds removing, reintroduced at the one place this fold created.

The new test uses **deliberately different** threaded and fresh values (`100.0` vs `999.0`), because equal values make forwarding and re-reading indistinguishable — which is exactly how a re-read hides. Fails on both bybit and okx under your mutation.

## `position_closed` — you were right, it's dead

No production reader; `_sync_position_from_exchange` already acts on its own return value. All four copies set it and nothing consumed it. Asserting a dead field is asserting metadata, so the field **and** the test I wrote to pin it are both gone rather than the test being rewritten to assert observable effects — there was nothing observable to assert.

Mixin docstring updated too; it claimed to provide "shared methods" and now owns the step.

---

**Diff +292/−315 (net −23).** Suite **4199**.

The through-line of your last three reviews, which I think is worth naming: every one found a check that *looked* like it constrained behaviour and didn't — a subset assertion where the varying part mattered, a hook test that never ran the wiring, a substring where ordering was the contract. In each case I had already written the stronger tool and used the weaker form. I have started asking "what mutation does this kill?" before writing the assertion rather than after, which is what produced the different-values detail in the seam test above.

PR #412, opened 2026-08-25
refactor(288): fold the four plain _step copies, and fix a fallback that never ran

BY571:
Slice 3 of #288. **Net −42**, suite **4209** green.

The four plain futures `_step` bodies were **25 lines each and differed by one trailing comment**:

```
binance vs bitget : 100%
bybit   vs okx    : 100%
across pairs      :  96%   <- the 4% was `# Placeholder, will be set after reward calculation`
```

`env.py`: 438/328/236/228 → **388/278/194/182**.

MRO checked *before* writing: SLTP envs resolve `SLTPMixin` ahead of the futures base and keep their own step; alpaca doesn't inherit the futures base at all.

## A graceful-degradation path that crashed instead

Writing the test for the `new_price` fallback surfaced this in the post-bar mark handler:

```python
except Exception:
    logger.warning("post-bar mark unavailable for %s; the history row will carry the "
                   "pre-trade price", self.symbol)     # <- no such attribute
```

These envs have no `symbol` attribute — it's `self.config.symbol`. So whenever the post-bar mark fetch failed, the handler raised `AttributeError` rather than falling back, which is precisely what the comment two lines above promises. **Pre-existing, unreachable by any test**, and found only because a reviewer flagged the fallback as unpinned.

## Three mutations that failed zero tests

| mutation | before | after |
|---|---|---|
| plain: `new_price` fallback dropped | 0 failed | **1 failed** |
| sltp: `new_price` fallback dropped | 0 failed | **1 failed** |
| plain: history `action` sign flipped | 0 failed | **8 failed** |

The action sign is the one that matters — that row is what the reward function scores, so flipping it inverts the training signal with nothing crashing. I pinned exactly this on the SLTP path in #411 and left the plain one: **this issue's own defect, inside the guard against it.** Both paths now.

## One more guard broken by the move

`test_the_size_an_env_reports_actually_reaches_the_position` scanned the venue's `env.py` for a call that now lives on the base. It's AST over the **resolved** `_step` now — which also fixes a weakness it always had: a substring can't tell `trade_info` from `trade_info["qty"]`, and half-passing is the bug it exists to reject.

## Mutation coverage of the folded step

| mutation | result |
|---|---|
| sync after the trade | 6 failed |
| recorder gets a field, not `trade_info` | 4 failed |
| trade re-reads instead of threading qty | 1 failed |
| `new_price` fallback dropped | 1 failed |
| history action sign flipped | 8 failed |

### DIFF
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -631,7 +631,9 @@ def _sole(module, suffix):
     """
     found = [
         v for k, v in vars(module).items()
-        if k.endswith(suffix) and getattr(v, "__module__", None) == module.__name__
+        if isinstance(v, type)          # `__module__` resolves through the class, so a
+        and k.endswith(suffix)          # module-level INSTANCE would otherwise qualify
+        and getattr(v, "__module__", None) == module.__name__
     ]
     assert len(found) == 1, (
         f"{module.__name__} defines {len(found)} classes ending in {suffix!r} "
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -621,3 +621,20 @@ def wire_outage_state(env, budget=0):
     env._status_unknown_this_step = False
     env._last_confirmed_read = {}
     env._max_unknown_status_steps = budget
+
+def _sole(module, suffix):
+    """The ONE class in `module` whose name ends in `suffix`.
+
+    Not `next(...)`, which returns the first match in DEFINITION order: a decoy declared
+    above the real class silently becomes the subject, and a guard then passes green on a
+    class nobody meant to check.
+    """
+    found = [
+        v for k, v in vars(module).items()
+        if k.endswith(suffix) and getattr(v, "__module__", None) == module.__name__
+    ]
+    assert len(found) == 1, (
+        f"{module.__name__} defines {len(found)} classes ending in {suffix!r} "
+        f"({[c.__name__ for c in found]}); the guard would have silently picked the first"
+    )
+    return found[0]
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -79,6 +79,45 @@ FUTURES_ENVS = [
     if issubclass(c, TorchTradeFuturesLiveEnv) and c is not TorchTradeFuturesLiveEnv
 ]
 
+# Split by module, and resolved from the registry rather than by scanning a module
+# namespace for a name ending in "TorchTradingEnv": `next(...)` returns the FIRST match in
+# definition order, so a second class defined above the real one silently becomes the
+# subject. Not hypothetical -- with a decoy that INHERITS the shared `_step` (and so
+# satisfies an identity check vacuously), the guard below passed green while the real
+# class's `_step` was re-forked to raise.
+PLAIN_FUTURES_ENVS = [c for c in FUTURES_ENVS if c.__module__.endswith(".env")]
+SLTP_FUTURES_ENVS = [c for c in FUTURES_ENVS if c.__module__.endswith(".env_sltp")]
+assert len(PLAIN_FUTURES_ENVS) == len(SLTP_FUTURES_ENVS) == 4, (
+    "a list that SHRINKS is the hazard, not one that empties: pyproject's "
+    "empty_parameter_set_mark already turns an empty list into a collection error, but "
+    "4 -> 3 is silent, and that is the shape #411's loss actually took"
+)
+
+
+def _sole(module, suffix, **require):
+    """The ONE class in `module` whose name ends in `suffix`.
+
+    Not `next(...)`: that returns the first match in DEFINITION order, so a second class
+    declared above the real one silently becomes the subject. Verified by mutation -- a
+    decoy that inherits the shared `_step` (and so satisfies an identity check vacuously)
+    let a re-forked `_step` on the real class pass green, in the very guard written to
+    catch it. Asserting the count converts that into a failure that names both classes.
+
+    `require` takes predicates applied to each candidate, for the call sites that need a
+    concrete or a dataclass one.
+    """
+    found = [
+        v for k, v in vars(module).items()
+        if k.endswith(suffix)
+        and getattr(v, "__module__", None) == module.__name__
+        and all(pred(v) for pred in require.values())
+    ]
+    assert len(found) == 1, (
+        f"{module.__name__} defines {len(found)} classes ending in {suffix!r} "
+        f"({[c.__name__ for c in found]}); the guard would have silently picked the first"
+    )
+    return found[0]
+
 
 @pytest.mark.parametrize("done_on_bankruptcy,portfolio_value,expected", [
     (True, 50.0, True),    # below 10% of the 1000 initial -> bankrupt
@@ -161,13 +200,24 @@ def test_discovery_covers_every_live_exchange():
     """
     exchanges = {cls.__module__.split(".")[-2] for cls in LIVE_ENVS} - {"shared"}
     assert exchanges == {"alpaca", "binance", "bitget", "bybit", "okx"}
-    # NON_SLTP_ENVS drives the call-site guard below, and an empty parametrize SKIPS rather
-    # than fails -- a module rename would silently retire that guard.
+    # NON_SLTP_ENVS drives the call-site guard below. An EMPTY list is already a collection
+    # error (pyproject's empty_parameter_set_mark); a list that merely shrinks is not, and
+    # that is the failure this pins.
     assert len(NON_SLTP_ENVS) == 5
+    # FUTURES_ENVS drives the observation re-fork guard and is NOT covered by the exchange
+    # set above: a venue that stops inheriting TorchTradeFuturesLiveEnv keeps its exchange
+    # name here and silently drops its cases from that guard.
+    assert len(FUTURES_ENVS) == 12 and len(LIVE_ENVS) == 16
 
 
 @pytest.mark.parametrize("method", [
     "_check_termination", "_sync_action_level_after_reset", "_build_observation_specs",
+    # Everything the shared `_step` calls. Guarding `_step` itself is not enough: a venue
+    # that re-forks one of ITS callees still passes the `_step` identity check while the
+    # "a shared fix lands once" guarantee is already gone. Verified -- a byte-identical
+    # copy of a callee on one venue failed zero of 1076 tests.
+    "_record_position_after_trade", "_resolve_action_level",
+    "_wait_for_next_timestamp", "_finalize_step_flags",
 ])
 @pytest.mark.parametrize("env_cls", LIVE_ENVS, ids=lambda c: c.__name__)
 def test_no_live_env_overrides_shared_method(env_cls, method):
@@ -194,6 +244,10 @@ def test_no_live_env_overrides_shared_method(env_cls, method):
         # what converts a failed fetch into a type `_halting` catches.
         "_current_mark_price",
         "_halting",
+        # The shared `_step`'s two state reads. Both apply the halt policy (#295); a venue
+        # copy of either is where that policy stops being one policy.
+        "_acquire_pre_trade_state",
+        "_acquire_post_bar_state",
     ],
 )
 @pytest.mark.parametrize("env_cls", FUTURES_ENVS, ids=lambda c: c.__name__)
@@ -2432,8 +2486,7 @@ def test_the_size_an_env_reports_actually_reaches_the_position(exchange):
     import importlib
 
     mod = importlib.import_module(f"torchtrade.envs.live.{exchange}.env")
-    cls = next(v for k, v in vars(mod).items()
-               if k.endswith("TorchTradingEnv") and v.__module__ == mod.__name__)
+    cls = _sole(mod, "TorchTradingEnv")
     call = next(
         (n for n in ast.walk(ast.parse(inspect.getsource(cls._step).lstrip()))
          if isinstance(n, ast.Call) and isinstance(n.func, ast.Attribute)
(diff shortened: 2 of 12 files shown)


BY571:
@CharlieHelps could you review this one? It's the third and last fold in the #288 series, and the one I'd most like a second pair of eyes on — because the first two taught me that a fold's real risk isn't the code, it's the tests that quietly stop testing anything.

## What it does

Four plain futures `_step` copies (binance/bitget/bybit/okx), 100%/96% identical, collapse onto `TorchTradeFuturesLiveEnv._step`. The 4% was a trailing comment on one line.

`+177/−194` (net −17). Modest, and the reason it isn't more negative is worth stating plainly: the deletion is 188 lines of source, but 132 lines of the addition are tests — the structural guards below. I'd rather carry those than have this fold be reversible by accident.

## A pre-existing bug fell out of the fold

Diffing the four copies surfaced a handler that could never have run:

```python
logger.warning("post-bar mark unavailable for %s; ...", self.symbol)
```

These envs have no `symbol` attribute — it's `self.config.symbol`. So the graceful degradation path the comment above it promises raised `AttributeError` instead of degrading, in all four copies, whenever a post-bar mark price was unavailable. Fixed in the shared copy.

I'd flag this as the strongest argument for the whole #288 series: the bug was invisible for as long as the code was duplicated, and obvious the moment it wasn't.

## What I'd like you to look at hardest

**1. The MRO reasoning.** SLTP envs must NOT get this `_step` — `SLTPMixin` precedes this class in their MRO and owns a `_step` that resolves a bracket tuple rather than an action level. I believe the MRO makes that automatic, and there's a guard asserting SLTP envs don't resolve to the shared one. But MRO bugs are exactly the kind that pass tests and lose money, so please check my reasoning rather than my test.

**2. Ordering.** The pre-trade read must precede the duplicate-action guard (invariant 2 — the guard reads the position it syncs), the trade must consume the qty/price that read already acquired under the halt policy rather than fetching its own (#295), and `record_step` must precede the reward (the reward scores the row it just wrote). This is asserted by AST position comparison rather than call-name membership, since `ast.walk()` is breadth-first and `calls.index(a) < calls.index(b)` proves nothing about source order — a mistake I made earlier in this series.

**3. Whether the guard actually guards.** Ask the question I now ask every check I write: *what mutation does this kill?* I verified `test_no_plain_futures_env_reforks_the_shared_step` by re-forking a byte-identical copy onto bitget — without the guard all 819 tests passed, with it the guard fails. That's the mutation it kills. If you can find a re-fork it would miss, that's the finding I want.

## Process, stated honestly

Round 1 is incomplete. `torchrl-engineer` (no divergence) and `test-justifier` (3 findings, all applied) finished; the other three agents hit the session limit. Rounds 2 and 3 haven't run. So this is **not** ready to merge yet — I'll finish the rounds. Your review is additional signal, not a substitute for them.

One thing I did verify myself, because it's the failure mode this series keeps producing: comparing collected test IDs between `main` and HEAD across `tests/envs/`, **zero vanished** (3194 → 3208, the 14 added being exactly the three new tests). On #411 twelve cases disappeared while the count stayed at exactly 795 — an emptied `parametrize` list skips rather than fails. The fix from that round is what prevented the damage here.

Suite: 4213 passed / 16 skipped / 0 failed. CI green.

BY571:
@CharlieHelps thanks — but your approval landed on `6ad684e` at 06:24Z and I pushed `3d6503f` at 06:28Z, four minutes later. It's a substantial change, so please re-review the new head rather than let the approval stand on code that no longer exists.

What changed, and why it matters to your review specifically:

**Your guard-strength probe passed for a reason worth knowing.** You planted a wrapper re-fork on bitget and the identity assertion killed it — correct. But I'd asked for a variant the guard would *miss*, and there was one: a decoy class declared **above** the real one in the module. `next(v for k, v in vars(mod)...)` returns the first name-matching class in definition order, so with a decoy that *inherits* the shared `_step` (satisfying the identity check vacuously), I re-forked `BitgetFuturesTorchTradingEnv._step` to raise and the guard passed green.

We both probed the assertion. Neither of us initially probed *which class was being examined*. The assertion was strong the whole time; the subject selection was broken. Eight call sites had that idiom — all now go through a `_sole()` helper that asserts uniqueness. An off-registry decoy that previously passed now fails 9 tests.

**Four contracts in the shared `_step` scored zero.** Dropping the reward overwrite, or recording the pre-trade price, or the pre-trade qty — each failed no test at all. The price one is the sharpest: `_acquire_post_bar_state` spends twenty lines arguing `price` must be the post-bar mark, citing measurements (12 of 14 rows wrong), and nothing asserted it. The new test gives pre-trade and post-bar **different** prices, which is precisely why it stayed invisible — with one price the two spellings are indistinguishable.

**Re-forking a callee escaped the fold's guarantee entirely.** A byte-identical copy of `_acquire_post_bar_state` on one venue passed 1076 tests: `_step` really was still shared, while "a shared fix lands once" was already gone. Both re-fork lists now cover every method the shared `_step` calls.

One correction to a claim I made in my own docstring, since it bears on your MRO check: I'd written that identity beats `"_step" in vars(cls)` because absence-from-subclass misses a base-class plant. I tested it — `SLTPMixin` sits directly after the class in the MRO, nothing between them, so today the two are equivalent and the plant is shadowed either way. Rewrote it to what I verified. Your MRO finding and mine agree.

The PR went from `+177/−194` to `+285/−244` — net positive now. That's the honest trade: 100 new cases pinning contracts four duplicated copies were never asked to keep.

Suite 4213 → 4313, 0 failed. Rounds 2 and 3 still to run, so this still isn't merge-ready — same caveat you stated yourself.

BY571:
@CharlieHelps your approval is on `3d6503f`; two commits have landed since (`1a68939`, `8c7ba5b`). Please re-review `8c7ba5b`.

Round 2 ran all five agents, and the honest headline is that **both of the late agents found defects I introduced while fixing round 1's**. That is worth knowing before you read the diff, because it says something about where the risk in this PR actually sits — not in the fold, which has been stable since the first commit, but in the tests I keep rewriting around it.

**A test passed a mutation that deleted its own scenario.** `test_a_flat_bar_falls_back_to_the_pre_trade_price` is armed by a one-shot flag flipped inside `_wait_for_next_timestamp`. Remove that call from the shared `_step` and the flag never flips, the post-bar mark never fails, the fallback never runs — and the test passed green. It now asserts its own setup happened. Note this is the *same* self-unaware one-shot mock I had caught in the sibling test earlier in the same round: I fixed the instance and not the class.

Its asserted value was also ambiguous — the fixture returns 100.0 from `get_mark_price` **and** fills `base_features` with 100.0, so 100.0 could not tell the pre-trade mark from the observer's close, which is exactly the plausible wrong implementation (the SLTP path prices brackets off the close). Now 137.0.

**My venue collapse was wrong for one test, and I had the reason backwards.** Venues differ in DATA, not only code: binance ships `[-1, 0, 1]`, the others `[-1, -0.5, 0, 0.5, 1]`. Collapsing the action test to binance made a mutation flattening the level to its sign *unobservable* on the only venue left. Half-size cases restored, and the assertion now checks the value rather than the sign.

**One place the guard was wrong rather than the code — I'd like you to double-check my call.** I extended the re-fork guard to `SLTPMixin`'s callees, and adding `_dispatch_sltp_trade` failed bybit and okx immediately. I concluded that is the venue-variation hook (its own docstring: *"the ONE thing `_step` varies by venue"*) and those two legitimately override it to thread the price rather than re-read it (#295) — so I excluded it and guarded the other three. If you think that reasoning is wrong and those two really are an incomplete fold, say so; I would rather be corrected than have quietly blessed a divergence.

**Also worth your eye:** the #295 threaded `current_qty` was pinned only incidentally, by a single test in bitget's venue file. `_execute_trade_if_needed` is *not* shared, so that path runs through venue-specific code on four venues and was exercised on one. Now pinned in the contract file.

14 mutations across both shared owners, all killed. Suite 4316, 0 failed, CI green. Production is `+43/−203` (net −160); the PR total is `+332/−302` because the tests grew to cover contracts four duplicated copies were never asked to keep.

Round 3 is starting now, so this still isn't a merge signal — same caveat you've been stating.

BY571:
@CharlieHelps round 3 is done (all five agents) and I've pushed `f0ec5cd`. Please re-review — and I want to flag plainly that **round 3 was not clean**, so by this repo's own standard it does not yet clear the bar for merge.

The finding that matters is in a test I wrote in round 2 to close a gap:

**The #295 test was tautological on its own claim.** Its docstring said it proved the trade takes the *threaded* qty and price rather than re-reading them. It could not. The fixture wires `get_status` and `get_mark_price` with `return_value`, so a second read returns the identical object. I swapped the threaded price for a fresh `self._current_mark_price()` — the exact #295 shape, the re-read that opens an outage window — and the test stayed green. It was only ever proving the kwarg wasn't zeroed.

Constants cannot distinguish "read once and thread it" from "read again"; only a venue that reports something different can. The mocks now change after the first read, and the read *count* is pinned at two per step — the stricter half, since a third read is a window even when the values happen to agree. Both re-read mutations now fail.

Two more, both in round-2 code:

- **A dead assertion.** `assert recorded * want > 0` sat directly under `assert recorded == approx(float(want))`. With `want ∈ {±1, ±0.5}` the product is `want²`, so the sign check cannot fail independently of the value check above it. I added the strong assertion and left the weak one beside it, where it reads as coverage and is arithmetic.
- **A self-referential oracle.** The reward assertion compared against `rows = len(history.actions)` measured *after* the step — a count that moves with the bug. A doubled `record_step` scores 2 against 2 rows and passes; I verified it did. The literal `1.0` is uglier and doesn't move.

Two things I'd value your judgement on rather than my own:

1. **Is a fourth round warranted?** Every round so far has found a defect introduced by the previous round's fix — this is the fourth in a row. That is either converging (each round's findings are smaller and more localised, which they are) or it is a signal that I should stop rewriting these tests. I lean toward one focused pass over the round-3 diff only. Tell me if you think that's over-process.

2. **The record-then-score tail is still duplicated** across `TorchTradeFuturesLiveEnv._step` and `SLTPMixin._step` — sixteen line-for-line identical lines, differing only in the `action=` expression. That is #288's own thesis surviving inside the PR that folds `_step`, and it is why the new test needs an `sltp` axis at all. I did **not** fold it: a new shared method arriving at round 3 buys another full review cycle. I'd rather file it as the #288 follow-up. Push back if you think it belongs here.

Diff is `+354/−308`; production is `+43/−203` (net −160). The total went up this round because making the #295 test actually test its claim cost ~30 lines — I'd rather carry those than keep a test whose docstring outran its mocks.

18 mutations across both shared owners, all killed. Suite 4316, 0 failed.

BY571:
@CharlieHelps all four addressed in `11a7d87`, and you were right on the one where I'd taken the opposite advice.

**The cardinality pin was not noise.** Round 3's simplifier called `assert len(_SHARED_METHOD_OWNERSHIP) == 4*2 + 4*5` tautological given the registry length asserts, and I deleted it on that basis. Your counter is correct and I verified it: those pins guard the **env** axis, and nothing guarded the **method** axis — dropping `_reset` from the matrix passes 936 tests silently. Restored, but as a *named* `(owner, method)` set rather than a count, so it also states what is meant to be shared and adding a genuinely new shared method is a deliberate edit rather than an arithmetic fix.

**The dispatch guard is in.** binance/bitget get the pair-specific identity assertion you asked for. I also added the converse for bybit/okx: excluding them from the uniform table costs the identity check, so there's now an AST assertion that their override still *forwards* the threaded `current_price` — the only reason the override exists. Both verified by mutation (a private binance re-fork fails the first; bybit dropping the kwarg fails the second).

**The #295 test's claim now matches its scope.** You reproduced the false negative I'd already fixed — distinct mark values, `reads["mark"] == 0` pinned — but you also asked for something I hadn't done: the test stubs `_execute_trade_if_needed`, so it pins the shared **caller's** forwarding, not a venue executor's consumption. The docstring says that now, and points at the SLTP grace-bar tests for the consumer half, rather than implying the broader guarantee.

**And the comment miscounted** — "one status read per step" is one pre-trade read plus the post-bar one, exactly as my own test asserts (`reads["status"] == 2`). Fixed in both places.

One disagreement, and it's not mine to settle: **you asked me to defer the record-then-score fold, and the user directed otherwise**, so it shipped in `576fd0e`. For the record it turned out to be where the two gaps I'd been *documenting* lived — alpaca's two copies of that tail were unpinned (zeroing the reward, dropping the overwrite, or scoring before recording each survived the whole suite there). There is no alpaca tail left to mutate now; the gap is closed by construction rather than by another test. The done-family guard failed on all ten envs when I did it, which was correct — I widened it to accept the one hop *and* pinned the delegate, so it now guards the chain rather than one link.

Worth flagging against your "expands production scope" concern: production went **more** negative, `+43/−203` → `+68/−260`.

One process note, since it bears on how much to trust my verification: mid-way through fixing your items I ran `git checkout` on the test file to undo a mutation and wiped all four fixes. I caught it only because the collected count dropped 944 → 940. Undo is by file copy now.

Suite 4334, 0 failed.

BY571:
@CharlieHelps thanks — and it was two wrong SHAs, not one. Corrected both in that comment in place:

- `1a68939` → **`576fd0e`** for the production fold, as you flagged.
- `2f53a83` → **`11a7d87`** for the commit addressing your four items — I opened the comment with a SHA from two commits earlier.

Both were stale values I had in context rather than values I read back from `git log`. That is the second time in this PR I have put an unverified SHA in a comment, and the first time it happened I wrote the correction *inside a paragraph about not making unverified claims*. So the rule I am actually adopting: no SHA goes in a comment that did not come from a `git log`/`git rev-parse` in the same action that wrote it.

Worth noting the shape of the error, since it is the same one this PR keeps finding in the tests: a value that was true earlier, carried forward, and never re-read. `1a68939` really was the commit I meant at the time I drafted that paragraph; the fold moved to its own commit afterwards and the reference did not follow. That is `rows = len(history.actions)` measured after the step, in prose.

Recording your approval on `11a7d87` and the verification behind it (179 targeted, 70 failsafe, 3320 env suite, CI green). No open items from your side.

PR #413, opened 2026-08-26
refactor(288): fold the fractional sizing, and the fee that made "identical" a lie

BY571:
Slice 4 of #288. Four `_calculate_fractional_position` bodies (54–100% similar) fold onto `TorchTradeFuturesLiveEnv`; binance extends via `super()`.

**Production `+76/−202` (net −126).** Suite 4344, 0 failed.

## The reason this slice was not a mechanical fold

The four bodies are **byte-identical text and different behaviour**. `TAKER_FEE` resolved to a different value in each module:

| binance | bitget | bybit | okx |
|---|---|---|---|
| 0.0004 | 0.0006 | 0.00055 | 0.0005 |

Folding them without lifting the fee would bind one constant in the shared module and silently re-price three venues. Sizing is `1 + leverage × fee`, so at 125× binance and bitget size a position **~2.3% apart**.

Worth stating plainly: my own similarity measurement called bybit and okx **100% identical**. They differ by 10% in fee. The metric was measuring text; the money was in the name resolution.

The fee is now a per-venue class attribute the shared body reads — pinned by value, plus a test that four different fees produce four different notionals, so "declared on the class" cannot decay into "declared and ignored".

## binance extends rather than inherits, and that is correct

binance refuses below min-**notional** and rounds the **target**. The other three floor the **delta** against min-qty inside `_execute_fractional_action`. Different placement in the pipeline, not a gap in three venues — I verified that before folding, because a fold is exactly where that distinction gets flattened by accident.

## Behaviour is unchanged, proved not asserted

144 sizings — 4 venues × 6 actions × 2 prices × 3 leverages — computed on this branch and on `main`. **Zero differ.**

## What the review rounds should look at

- The `not (x > 0)` guard failed on all four venues when the code moved: it was a **file scan** of `live/<venue>/env.py` and could no longer find its own subject. Third time a file-scan guard has broken on a fold in this series, and the third time it was right to. It now reads the resolved method and follows one hop through `super()`.
- **Two of my own new assertions were too weak**, and the mutation battery caught both: grepping the body for `_get_min_notional` passed with the `if` that uses it neutered, and the 2% buffer test compared two of the env's *own* sizings — a ratio invariant to the buffer, so it passed with the buffer deleted. Both now assert behaviour against an independent expected value.

7 mutations across the fold, all killed. Review rounds to follow.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -669,7 +669,6 @@ def test_an_outage_stops_the_step_before_it_can_trade(exchange, module):
     and the trade path by hand, in an order _step cannot produce, and so proved nothing
     about production -- _step raises on the status read before it ever reaches the sync.
     """
-    import importlib
     env_mod = importlib.import_module(f"torchtrade.envs.live.{exchange}.{module}")
     # By NAME, not by `"_step" in vars(c)`: #288 moved the SLTP `_step` onto the mixin,
     # and a discovery predicate keyed on where the method LIVES silently stops finding
@@ -2399,7 +2398,6 @@ def test_every_live_config_validates_its_action_levels(exchange):
     """The WIRING, per exchange. Crossing the two dimensions proved the same thing five
     times over: a missing call kills every value row for that exchange together, so the
     values never distinguish anything the rule test above does not already cover."""
-    import importlib
 
     module = importlib.import_module(f"torchtrade.envs.live.{exchange}.env")
     config_cls = _sole(module, "Config")
@@ -2457,7 +2455,6 @@ def test_the_size_an_env_reports_actually_reaches_the_position(exchange):
     # TorchTradeFuturesLiveEnv, so a file scan reads a module that no longer contains the
     # call. And a substring cannot tell `trade_info` from `trade_info["qty"]` -- which is
     # exactly the half-passing this guard exists to reject.
-    import importlib
 
     mod = importlib.import_module(f"torchtrade.envs.live.{exchange}.env")
     cls = _sole(mod, "TorchTradingEnv")
@@ -2501,7 +2498,6 @@ def test_a_real_step_lands_both_size_values_on_the_position(exchange, min_qty):
     existed, the other that the call site named the dict. Only reading the value off the
     position after a real step can tell you it arrived, so this drives `_step` and looks.
     """
-    import importlib
 
     module = importlib.import_module(f"torchtrade.envs.live.{exchange}.env")
     env_cls = _sole(module, "TorchTradingEnv")
@@ -2652,7 +2648,6 @@ def test_the_pre_trade_read_halts_like_the_post_bar_one(exchange, module):
     it. Resolution is the stronger check anyway -- a venue that re-forks `_step` without
     the helper still fails, because `inspect.getsource` then returns ITS copy.
     """
-    import importlib
 
     mod = importlib.import_module(f"torchtrade.envs.live.{exchange}.{module}")
     cls = _sole(mod, "TorchTradingEnv")
@@ -4028,7 +4023,6 @@ def test_each_venue_builds_its_observer_with_the_arguments_it_needs(venue, expec
     `client` is asserted as a CATEGORY, not a value: bybit shares one session between the
     two roles and the other three must not, which is the actual contract.
     """
-    import importlib
 
     # The concrete plain env, not the abstract exchange base: the base still declares
     # `_step`/`_execute_trade_if_needed` abstract, so it cannot be instantiated at all.
@@ -4108,7 +4102,6 @@ def test_init_trading_clients_wires_each_venue_as_before(
     session, so it must be built second; okx must NOT, because its trader client is a
     Trade API and klines live on the MarketData one -- the round-1 regression.
     """
-    import importlib
 
     module = importlib.import_module(f"torchtrade.envs.live.{venue}.env")
     cls = _sole(module, "TorchTradingEnv")
@@ -4244,10 +4237,7 @@ def test_binance_extends_the_shared_sizing_rather_than_replacing_it():
         cls._calculate_fractional_position
     ), "binance's sizing no longer delegates to the shared implementation"
 
-    # By BEHAVIOUR, not by grepping the body for `_get_min_notional`: neutering the `if`
-    # that uses it left the name in the source and the string check passed.
     env = _sizing_stub(cls, min_notional=1e12)   # nothing can clear this
-    env.trader.round_quantity.side_effect = lambda q: round(float(q), 3)
     assert env._calculate_fractional_position(1.0, 100.0) == (0.0, 0.0, "flat"), (
         "binance opened a position whose notional is below the venue minimum; the "
         "exchange rejects it and the env then believes it holds one"
@@ -4255,9 +4245,7 @@ def test_binance_extends_the_shared_sizing_rather_than_replacing_it():
 
     env._get_min_notional = lambda: 0.0
     env.trader.round_quantity.side_effect = lambda q: 0.123456
-    # Signed, and both directions. `abs(size)` discarded precisely the expression under
-    # test -- `position_qty * (1 if position_size > 0 else -1)` -- and only long was run,
-    # so the short branch was never entered at all.
+    # Signed, and both directions.
     for action, expected in ((1.0, 0.123456), (-1.0, -0.123456)):
         size, _, _ = env._calculate_fractional_position(action, 100.0)
         assert size == pytest.approx(expected), (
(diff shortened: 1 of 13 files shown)


BY571:
@CharlieHelps could you review this one? Slice 4 of #288, and the interesting part is not the fold.

**The four bodies were byte-identical text and different behaviour.** `TAKER_FEE` resolved to a different value in each module (0.0004 / 0.0006 / 0.00055 / 0.0005), so the obvious fold binds one constant in the shared module and silently re-prices three venues. My own similarity check called bybit and okx **100% identical** — they differ by 10% in fee. That is the thing I would most like a second pair of eyes on: whether the class-attribute approach actually keeps the fee venue-specific everywhere it matters, or whether I have moved the same trap somewhere less visible.

**What I would like you to attack, in order:**

1. **The fee.** `TAKER_FEE` is now a class attribute each venue sets from its own module constant, read as `self.TAKER_FEE` in the shared body. Is there any path — subclassing, the SLTP variants, `__init__` ordering — where a venue ends up reading someone else's fee? I have a test pinning the four values and a test that four fees produce four different notionals, but both were written by the person who wrote the bug they are guarding.

2. **binance extending rather than inheriting.** binance refuses below min-**notional** and rounds the **target**; the other three floor the **delta** against min-qty in `_execute_fractional_action`. I concluded that is a real placement difference, not a gap in three venues. If you think the other three are actually missing a min-notional check, say so — that would be a live bug, not a refactor finding.

3. **My equivalence proof.** 144 sizings across 4 venues × 6 actions × 2 prices × 3 leverages, this branch vs `main`, zero differ. Tell me what that grid is blind to. It does not cover a flat account, an unreachable venue, or a `round_quantity` that returns something surprising.

**Two of my own new assertions were too weak** and my mutation battery caught them, which is worth knowing when judging how much the rest of my tests are worth: grepping the body for `_get_min_notional` passed with the `if` that uses it neutered, and the 2% buffer test compared two of the env's *own* sizings — a ratio that is invariant to the buffer, so it passed with the buffer deleted. That second one is the self-referential-oracle mistake I flagged and fixed on #412 and then made again here.

Suite 4344, 0 failed. Review rounds starting now, so this is not a merge signal yet.

BY571:
@CharlieHelps thanks — acting on all three, and note your review was on `e534e40`; two commits have landed since.

**Your point 1 is now moot, because it was worse than you found it.** You wrote that the SLTP classes "still read their own module-level `TAKER_FEE`" so there is no cross-venue leak. That was true of the *call sites you looked at*. But the fold also made the SLTP classes **inherit** `_calculate_fractional_position` from the futures base while `TAKER_FEE` sat on the plain leaf — which their MRO never touches. Calling it on any of the four raised `AttributeError`. Latent, since nothing reaches it yet, which is exactly why no test caught it. The fee is now on `<venue>/base.py`, which both siblings inherit, and `_finish_futures_init` refuses to construct an env without one.

**Your point 3 was the most useful and I have fixed it.** You are right that the 144-case grid lived in `/tmp` and evaporated — a refactor's central claim backed by evidence nobody else could re-run. The eight pre-fold numbers are committed now (four venues × 1x and 125x, to the last float bit). At 125x they diverge ~2.3%, so the test doubles as visible proof the venues are still priced apart, and I verified it bites: retyping bybit's fee to okx's fails two cases by name.

I did not turn the other axes you named — SLTP, real exchange filters, direction switches, revalidation after rounding — into differential evidence, and I am not claiming otherwise.

**Your point 2 findings are recorded on #288**, not fixed here, per your own recommendation. All three are the same shape: an order the env believes it placed and the venue rejected, which is invariant 2 inverted. Worth saying that the comparison across four venues is what made them visible at all — an argument for the deduplication rather than against it.

Round 1 also turned up something you and I both missed: my balance guard was defended by **two string greps**, and four mutations kept both strings intact and passed — including `x <= 0` → `x < 0`, which accepts **zero equity**, the liquidation moment the guard's own comment names as the worst one to keep trading through. It is behavioural now, asserting `LiveObservationHalt` rather than `ValueError`, because `_halting` only converts the error if it is raised inside the closure — so the exception *type* is what catches a hoisted guard.

Rounds 2 and 3 still to run, so not a merge signal. Suite 4370, 0 failed.

BY571:
@CharlieHelps three review rounds are done (round 3 is 3 of 5 — two agents died to a session limit, resuming at 17:40). Please review the current head; your last review was on `e534e40`, five commits back.

**The finding I'd most like you to check, because two reviewers got it wrong and I nearly acted on them.**

The shared `_calculate_fractional_position` opens with `if action_value == 0.0: return 0.0, 0.0, "flat"`. Two agents across two rounds called it a redundant third copy — the four callers short-circuit on zero first, and the free function guards zero again underneath. One verified that the free function returns `(0.0, 0.0, "flat")` without it and recommended deletion.

Measuring instead of reasoning:

```
with guard,    flat action, NaN balance -> (0.0, 0.0, "flat"),  balance NOT read
without guard, flat action, NaN balance -> ValueError,          balance READ
```

The guard's purpose is skipping a network read, not producing `"flat"`. An equivalence argument that compares only return values is blind to that. Deleting it would have been a behaviour change on the liquidation path #295 exists for, in a PR whose entire claim is that behaviour did not change — and the golden table only exercises `action=1.0`, so it would have stayed green. It is kept and now pinned by a test asserting the balance is not read. **If you think that reasoning is wrong, say so** — it is the one production line in this PR where I overruled two reviewers.

**Two corrections to things I told you earlier.**

1. **My equivalence proof was invalid when I first gave it to you.** `python /tmp/probe.py` sets `sys.path[0]` to the script's directory, so `import torchtrade` resolved to the editable install — the main repo tree. Every comparison I ran from a worktree was branch-versus-itself. Redone via `-c` with `assert torchtrade.__file__.startswith(expected_tree)`: true pre-fold `main` vs HEAD is 0 of 144. The conclusion held; the method did not, and I had stated it as verified four times.

2. **The golden table you asked me to commit was blind to binance** — the one venue whose sizing the fold does not fully absorb. It asserted `notional_value`, and binance's override transforms only `position_size`; deleting that override entirely left it 8/8 green. It now pins both, and binance's rows are lot-rounded where the other three are not, which is the override visible in the data.

**Also fixed across the rounds:** the balance guard was defended only by string greps and let four mutations through (including `x <= 0` → `x < 0`, accepting zero equity); an `available_balance` fail-open survived every test because the fixture set both balances equal; the SLTP classes inherited the sizing but not the fee, so four of eight classes would `AttributeError`; and a fee test pinned an attribute no SLTP code path reads.

Production is `+80/−231` (net −151). Suite 4356, 0 failed.

Your three live findings from last time (binance revalidating min-notional on the rounded delta, binance SLTP having no min-notional guard, bitget/bybit ignoring their exchanges' minimum-notional fields) are recorded on #288 and remain out of scope here, per your own recommendation.

BY571:
@CharlieHelps round 3 is complete (5 of 5 — the two agents that hit the session limit finished after it reset). Two corrections to what I told you three comments ago, both against my own work.

**1. I overclaimed the flat-guard finding, in the same message where I told you I had measured rather than reasoned.** I said deleting the guard "changes behaviour on the liquidation path #295 exists for." The measurement was right; the scope was not. All four callers pre-filter `action_value == 0.0` before the sizing is reached — I checked the source positions in each — so **no production path reaches that guard at all.** It defends a caller that stops pre-filtering, not a live one. Keeping it is still right (it is one line, and now pinned), but the justification I gave you was broader than the code's reach. Corrected at the guard and in the test.

Measuring protects a claim's truth value, not its scope. Scope is where my prose keeps drifting, and this is the third time in this PR.

**2. The test I added to pin that guard was blind one venue over.** Deleting binance's *own* `if side == "flat"` early return survived the entire suite — a flat sizing call then issues `futures_exchange_info()`, an unhalted HTTP round-trip down the one path the test exists to keep exchange-free. Parametrising over four venues does **not** fix it: `_sizing_stub` stubbed `_get_min_notional`, which swallowed exactly that call. The stub is optional now and the assertion counts every trader call, not just the balance read. Both mutations die.

That is the second fixture in this PR that masked the distinction under test, after the equal-balance fixture that hid the `available_balance` fail-open. A stub that answers a call the code was supposed to avoid making cannot see whether it avoided it.

**Where that leaves the PR.** Round 3's harness: 34 of 39 mutations die. The five survivors are one `<` vs `<=` boundary where HEAD is correct and the mutant wrong, one pre-existing binance delta-rounding gap that #271 left unpinned, and two equivalent mutants. Equivalence against true pre-fold `main` is 0 of 144, independently reproduced by two reviewers with their own harnesses after mine turned out to be measuring the wrong tree.

Both reviewers who ran correctness checks call it merge-ready. Production is `+83/−231` (net −148); suite 4359, 0 failed.

Your three live findings and two new pre-existing ones (binance's unpinned delta rounding; the SLTP fee's silent fallback) are recorded on #288.

BY571:
@CharlieHelps thanks — your approval is on `0a8b26f` and HEAD is now `128cd6f`, pushed after it. One commit, and it changes something you explicitly verified, so please re-review rather than let the approval carry.

**What changed, and why it matters to your review.** You wrote that you "reproduced the no-read behavior across all eight plain/SLTP venue classes." That is right for the *shared* guard. But binance has a **second** flat branch of its own — `if side == "flat": return ...` inside its override — and deleting that one survived the entire suite. A flat sizing call then issues `futures_exchange_info()`: an unhalted HTTP round-trip on exactly the path the no-read property is about.

My test did not catch it either, and the reason is worth stating: `_sizing_stub` stubbed `_get_min_notional`, which *swallowed that very call*. Parametrising the test over four venues still passed. The stub is optional now and the assertion counts every trader call rather than the balance read alone. Both mutations die — the shared guard on all four venues, binance's own branch on binance.

So the no-read property you reproduced held at the method boundary and leaked one layer in, at the only venue that adds code after the shared guard.

**Also in that commit: I narrowed a claim you were more careful about than I was.** I had justified keeping the guard as "changes behaviour on the liquidation path #295 exists for." Your phrasing — "changes observable behavior at the method boundary" — is the accurate one. All four callers pre-filter `action_value == 0.0` before the sizing is reached, so no production path reaches that guard at all; it defends a caller that stops pre-filtering. I have corrected the guard comment and the test docstring to say that.

**And a decision stated rather than left implicit,** since a reviewer asked me to confirm it was deliberate: round 1 removed the `_finish_futures_init` TAKER_FEE boundary check. That cuts against CLAUDE.md invariant 4. The mitigation is that `test_every_class_that_inherits_the_shared_sizing_can_run_it` is derived from a length-pinned registry and *executes* the sizing, so a venue that forgets its fee cannot reach green CI — it fails in CI rather than at construction. I am content with the trade and want it on the record as a choice.

Round 3 finished 5 of 5. Harness: 34 of 39 mutations die; survivors are one `<` vs `<=` where HEAD is correct and the mutant wrong, one pre-existing binance delta-rounding gap (#271 pinned the target and left the delta — recorded on #288), and two equivalent mutants.

Suite 4359, 0 failed. Equivalence vs true pre-fold `main`: 0 of 144, reproduced independently by two reviewers after my own probe turned out to be reading the wrong tree.

Issue #414, opened 2026-08-27
Live: four order-size validation gaps that let the env believe it placed a rejected order


BY571:
Split out of #288 so that issue can close on deduplication alone. All four were surfaced by comparing the four venue stacks side by side during the folds (#410–#413) — they are pre-existing, none was introduced by that work, and none is a refactor finding.

They share one shape: **an order the env believes it placed and the venue rejected.** That is invariant 2 inverted — the cache says a position exists while the account is flat — and this repo has already paid for that failure class twice.

### 1. binance revalidates min-notional on the wrong quantity

A direction switch validates min-notional on the rounded **delta**, then closes and opens `abs(target_qty)`. The rounded target can sit below `MIN_NOTIONAL` while the larger switch delta passes.

Worked example: price `99.99`, minimum `100`, raw target `1.0002` floors to `1.0` → notional `99.99` (below the minimum), while a `1.01` delta passes the check. **The opening quantity is what the exchange sees, and it is not the quantity that was validated.**

### 2. binance SLTP has no min-notional guard at all

Separate execution path; it submits fractional quantities and its executor enforces only quantity constraints.

### 3. bitget and bybit never read their exchanges' minimum-notional fields

Both consume min-qty and step but ignore `minTradeUSDT` (bitget) and `lotSizeFilter.minNotionalValue` (bybit). No equivalent concrete omission for OKX derivatives, where `minSz`/`lotSz` are the binding constraints.

### 4. binance's delta lot-rounding is unpinned

`torchtrade/envs/live/binance/env.py`, in `_execute_fractional_action`:

```python
delta = self.trader.round_quantity(abs(delta)) * sign
```

Replacing that entire line with `delta = delta` **passes all 3,168 tests**. #352 fixed the **target** rounding for #271 and left the **delta** rounding untested. Consequence is an un-lot-rounded quantity sent to the venue — same rejection class as the three above.

### Suggested order

(4) first: it is a one-line behavioural test against code that already works, and it pins the thing the other three fixes will need to not break. Then (1), which is a real bug with a worked example. (2) and (3) are venue-parity work.

### Credit

(1)–(3) found by @CharlieHelps while reviewing #413. (4) found by a mutation harness in the same review round.

PR #415, opened 2026-08-27
refactor(288): fold the SLTP bracket sizing, which #413 left inlined four times

BY571:
Slice 5 of #288. `_resolve_bracket_quantity` on `TorchTradeFuturesLiveEnv` replaces the trade_mode dispatch, halted balance read, fee resolution and sizing call that each `<venue>/env_sltp.py` carried its own copy of.

**Production `+82/−216` (net −134).** Suite 4359, 0 failed.

## What the four copies actually were

Measured, not assumed: 85 of binance's 197 lines were shared with all three others, and the largest shared run is this 53-line block.

- **bybit vs binance** — differs by a two-line *comment*. Code byte-identical.
- **okx** — that same comment, plus a real five-line min-qty short-circuit. That one stays in okx.

So the sizing was code-identical four ways, and the venue difference was one guard rather than a signature. It reads `self.TAKER_FEE` rather than the module constant — the class attribute #413 put on `<Venue>Base` is inherited by the SLTP siblings, which is the plumbing that slice set up for this one.

## Equivalence, proven

**864 cases per tree** — 4 venues × 4 trade modes × 2 prices × 3 leverages × 3 balances (including `0.0` and `NaN`) × 3 `trader.transaction_fee` values (absent, valid, unusable) — comparing both the returned quantity **and** the exact params reaching `calculate_fractional_position`. **Zero differ**, with `assert torchtrade.__file__.startswith(tree)` on both sides so the probe cannot read the wrong one.

## Two things worth your attention in review

**The dead imports masked my own verification.** The fold left `TAKER_FEE`, `calculate_fractional_position` and `PositionCalculationParams` imported-but-unused in all four venue files — 12 dead names. My probe patches `calculate_fractional_position` *wherever it is bound*, found the dead import in the venue module, and patched that one — while the shared method used its own. The spy recorded nothing and 144 rows read as divergent when the quantities were identical in all of them. Removing the dead imports fixed the probe as well as the code.

**The SLTP finiteness guard failed on all four venues** when the code moved — it was a file scan of `env_sltp.py`, and the balance check now lives in the shared base. Third file-scan guard in this series to break on a fold, and right every time. It resolves the method and follows the delegation now; the `current_price` half still reads the venue body, which is where that check still lives for binance and bitget.

## Still open after this

The SLTP fee's `trader.transaction_fee` fallback degrades on an unusable value with only a `logger.warning`, and that branch has no test in either direction. It is now in **one** place rather than four, so it is finally cheap to pin — but it is behaviour this PR preserves rather than changes, so I have not touched it here.

Review rounds to follow.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -4285,16 +4285,6 @@ def _sizing_stub(cls, *, leverage=10, balance=10_000.0, min_notional=0.0,
     return env
 
 
-# Derived from the registry so a fifth venue fails here rather than on its first live
-# trade. The values stay literal: the failure guarded against is every venue reading ONE
-# fee, which a derived expected value would reproduce.
-_EXPECTED_TAKER_FEES = {
-    "binance": 0.0004, "bitget": 0.0006, "bybit": 0.00055, "okx": 0.0005,
-}
-assert set(_EXPECTED_TAKER_FEES) == {c.__module__.split(".")[-2]
-                                     for c in PLAIN_FUTURES_ENVS}
-
-
 @pytest.mark.parametrize("cls", PLAIN_FUTURES_ENVS + SLTP_FUTURES_ENVS,
                          ids=lambda c: c.__name__)
 def test_every_class_that_inherits_the_shared_sizing_can_run_it(cls):
@@ -4336,14 +4326,20 @@ _GOLDEN_SIZINGS = [
 # Pre-fold values, read off `main` before #415 moved the block. One row per venue at a
 # single leverage: unlike the plain golden table, nothing downstream of this method varies
 # by venue except `TAKER_FEE`, so a leverage axis would be decoration rather than evidence.
-PLAIN_VENUES_SLTP = [c.__module__.split(".")[-2] for c in SLTP_FUTURES_ENVS]
-
+# Completeness pinned, not just the values: a fifth venue would otherwise drop out of this
+# table silently while the other SLTP tests picked it up. Shrinkage is the hazard, not
+# emptiness -- an empty parametrize is already a collection error here.
 _GOLDEN_BRACKET_QUANTITIES = [
     ("binance", 489.02195608782426),
     ("bitget", 488.5343968095713),
     ("bybit", 488.6561954624782),
     ("okx", 488.77805486284296),
 ]
+
+
+assert {v for v, _ in _GOLDEN_BRACKET_QUANTITIES} == {
+    c.__module__.split(".")[-2] for c in SLTP_FUTURES_ENVS
+}
 
 
 def _bracket_stub(cls, *, balance=10_000.0, leverage=5, fee=...):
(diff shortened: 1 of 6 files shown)


BY571:
@CharlieHelps three review rounds are complete (5 agents each, 15 reports). Please review.

**What the fold is.** `_resolve_bracket_quantity` on `TorchTradeFuturesLiveEnv` replaces a 53-line block each of the four `<venue>/env_sltp.py` carried. Measured before touching anything: 85 of binance's 197 lines were shared with all three others; bybit differed from binance by a two-line **comment** and nothing else; okx by that comment plus a real five-line min-qty short-circuit, which stays in okx.

**Equivalence.** 864 cases per tree (4 venues × 4 trade modes × 2 prices × 3 leverages × 3 balances incl. 0.0/NaN × 3 `transaction_fee` values), comparing the returned quantity **and** the exact params reaching `calculate_fractional_position`, with `assert torchtrade.__file__.startswith(tree)` on both sides. Zero differ. Independently re-derived by two reviewers with their own harnesses — one at 51,840 cases, one at 7,040 by hand-transcribing the pre-fold formula from the diff rather than importing it.

**What I'd like you to attack:**

1. **The `None` contract.** An unusable balance returns `None` and each caller reports a failed trade; the plain path raises inside `_halting` and becomes a halt. I preserved that divergence rather than fixing it — filed as #416. A reviewer confirmed refusing is arguably safer for opening a *new* bracket from flat, since the function returns before touching any existing position. Tell me if preserving it is the wrong call.

2. **`abs()` removed from the sized quantity.** Two reviewers split; the third traced it: with `abs()`, a corrupted `position_fraction = -0.5` returns `244.5` — a plausible positive quantity flowing into `trader.trade()`. Without it, `-244.5` gets refused by the venue. I took "laundering, not insurance". It is the one behavioural line in the diff that is not a pure move.

3. **The `Unsupported trade_mode` raise.** Two reviewers said delete it as unreachable (the config validates at `__post_init__`); one argued the opposite and then withdrew in my favour. `fractional` is the *fall-through*, so deleting it sizes an unknown mode against real equity. Kept and now pinned.

**Three defects the rounds found in my own work**, worth knowing when judging the rest: `math.isfinite(candidate)` in the fee check was dead (the range test already rejects NaN and both infinities); my round-1 guard rewrite silently gave up the ability to catch a venue re-inlining the sizing (`inspect.getsource` follows inheritance, so it returned the shared helper for a venue that had stopped calling it); and three prose cuts I reported making had silently not happened, because the edit used `if needle in text` rather than asserting.

Production `+72/−238` (net −166). Suite 4394, 0 failed. #416 tracks the two deferred behaviour questions; #414 tracks four live order-validation bugs the comparison surfaced.

BY571:
Correction to my summary above, found by the last of the fifteen agent reports.

**The commit whose subject is "three cuts I reported making had silently not happened" contained another silent edit failure.** Fixing a variable alias, I scoped `replace("override", "fee")` to the function body — and `override` is in the function *name*. The test became `test_a_usable_trader_fee_fees_the_venue_constant`, with a docstring reading *"-- So the fee fee has been proven only in the direction that ignores it."* I asserted the anchor matched and then did an unanchored substring replace inside it, which is the same defect one level down.

It also deleted the comment explaining why `0.0` is in that parametrize — it is the **accept**-side boundary (`0 <= fee`, not `0 < fee`), and that case is the only thing standing between a fee-free replay executor and #278's shape. Restored into the docstring, where the next edit cannot orphan it.

**And the completeness assert I added in round 2 was pinned at the wrong granularity.** `{v for v, _, _, _ in _GOLDEN_SIZINGS}` compares venue *names*, so dropping one of a venue's two rows leaves the set identical and loses a test silently — precisely the shrinkage the assert exists to prevent. Verified: dropping okx's 1x row passed. It compares `(venue, leverage)` pairs now, and dropping either row is a collection error.

Both are fixed in `2093e05`. Neither changes behaviour; both are the same failure mode this PR keeps surfacing — an edit or a check that runs, reports success, and is not looking at what I think.

Final matrix from that reviewer: **24 of 32 mutations killed**, zero test IDs removed against `main` (4373 → 4408). All eight survivors are either documented equivalents (the `success` flag is unreadable when `executed` is False; okx's `None` absorbed by its own min-qty check), provably equivalent (`abs()` over the reachable domain), or pre-existing gaps in code this PR does not touch.

BY571:
@CharlieHelps all three blocking findings addressed in `HEAD`, and your second one corrects the reasoning I used to justify the change in the first place.

**You were right that my fail-closed claim does not hold across venues.** I removed `abs()` arguing a negative quantity would reach the venue and be refused loudly. Measured, per your note:

```
okx    _format_size(-244.5) -> floor -244.500 -> clamped UP to min_qty -> 0.001  A LONG
bitget forwards the negative, relies on the exchange
bybit  sends str(quantity) raw
binance raises below the step
```

So removing `abs()` was right — it laundered corruption into a plausible positive — but "the venue will catch it" was wrong for the one venue that turns it into an order. Taking your preference: **the shared helper now refuses a non-positive quantity itself**, which is the only outcome identical on all four. Pinned by a test that re-adding `abs()` fails.

**The delegation guard is an AST call check now.** You are right that a comment mentioning the name satisfied a substring search. A round-3 reviewer rated this 4 and I chose to document the limitation instead of closing it — documenting a bypassable guard is not the same as having one. Verified: okx inlining the sizing with the name left in a comment now fails 5 tests, and previously failed none.

**okx's min-qty branch is pinned.** Deleting it left the whole okx SLTP suite green, and it is not redundant caution — okx is the only venue where a sub-minimum quantity becomes a *larger* position rather than a rejected one, by the same clamp-up. A below-minimum bracket now asserts no order is sent.

All three verified by mutation rather than by reading:

| mutation | before | after |
|---|---|---|
| okx inlines, comment names the helper | 0 failures | **5** |
| okx min-qty branch deleted | 0 failures | **1** |
| `abs()` restored | 0 failures | **1** |

Equivalence for well-formed inputs is unchanged: **0 of 864** against pre-fold `main`. The refusal is unreachable from a validated config (`position_fraction` is `(0, 1.0]`), so it alters no behaviour any caller can currently produce — it pins the one thing this PR deliberately changed.

Suite 4397, 0 failed. Thank you for the venue-by-venue trace on the negative quantity; that is the part I had assumed rather than checked.

PR #417, opened 2026-08-27
fix(288): the close action that two of four SLTP venues ignored

BY571:
## The bug

`include_close_action` puts `("close", None, None)` in the action map for all four futures SLTP venues — the field lives on the shared `BaseFuturesSLTPConfig`. **Only bybit and okx ever handled it.** On binance and bitget the tuple fell past both `if side == "long"` and `elif side == "short"` and returned `executed=False`.

Measured on all four, before the fix:

| venue | `close_position` called? | executed | position after |
|---|---|---|---|
| binance | **False** | **False** | **1** ← close ignored |
| bitget | **False** | **False** | **1** ← close ignored |
| bybit | True | True | 0 |
| okx | True | True | 0 |

The policy asks to close, the position stays open, and nothing says so.

A quieter version of the same silence: a close the venue **refuses** (`close_position()` returns `False` — a real path; both bybit's and okx's executors return it when the re-query shows the position still open) or that **raises** left `success=None`, which is exactly what HOLD returns. Both now report `success=False` on all four venues.

To be precise about what that second fix is worth: **it changes no behaviour today.** Every reader of `success` gates on `executed` first, and on a failed close `executed` is already `False`, so nothing downstream sees the new value. The real protection is that the code does *not* mark itself flat on failure, plus `_sync_position_from_exchange()` at the top of the next step. It is a diagnostic contract made uniform, not a money bug fixed — worth the lines because the divergence it removes is the kind that gets built on.

## The fold is what made it visible

Replacing binance's hand-rolled bracket arithmetic with `calculate_bracket_prices` — which bitget already called — left the two `_execute_trade_if_needed` bodies **100% identical on code lines**. One copy now lives on `SLTPMixin`.

bybit and okx keep their overrides. The four venues split by **signature**, not by exchange: those two take the mark threaded in by the caller (`*, current_price`, #295) rather than pricing off their own candle.

**The close itself was three copies, not two.** The mixin, bybit and okx each carried the same 22 non-comment lines, so folding binance and bitget together left three. `SLTPMixin._close_action` is now the single owner for all four, and it goes in the uniform `_SHARED_METHOD_OWNERSHIP` table that `_execute_trade_if_needed` cannot use. The `success` divergence above is now structurally unreachable rather than test-enforced.

Also folded, each confirmed unreachable or dead before removal:
- The per-side test in the direction-switch branch. By then `side` is long or short and the duplicate guard has returned on a same-direction position, so a non-zero position **is** the opposite one. bybit had already dropped it.
- The long/short branches, which carried identical arithmetic differing only by the literal — the shape bybit and okx already ship.
- Dead `logging` imports in both stripped files (0 call sites, 0 external readers), `Dict`/`Tuple`, and an `import torch` unused since before this PR.
- Two test classes my own deletions emptied, plus their unrequested `env_with_close` fixtures.

## The traps, and how each is pinned

**Carrying bybit's guard across needs its early return, not just its side derivation.** `"buy" if side == "long" else "sell"` alone turns the silent no-op into a **wrong BUY on an open short**, doubling the position instead of flattening it. The test is parametrised over both directions because only the short case kills that.

**The branch collapse is safe on an ordering, not a type.** `calculate_bracket_prices` raises on any side that is not long/short and runs *before* `trade()`, so an unrecognised side raises there rather than reaching the venue as the `else` sell. Reverse those two lines and identical-looking code sends a real order.

**The first patch for the refusal made a smaller copy of the bug it was fixing** — `success=False` landed on the mixin only, so binance/bitget would have disagreed with bybit/okx in a shared field. Caught in round 2, and the reason the close is now one copy rather than three.

**The fold had no ownership guard.** `_execute_trade_if_needed` can't go in the uniform table (bybit/okx legitimately override), so it gets the split form `_dispatch_sltp_trade` already has. The gap was real: appending a faithful private copy back onto binance **passed all 1202 tests**, because the only thing watching was a sizing test that greps the resolved source for `_resolve_bracket_quantity` — which a re-fork still contains.

## Tests

Four shared parametrized tests over all four venues, replacing six per-venue ones that covered two:

| behaviour | previously |
|---|---|
| close actually closes (long + short) | binance/bitget: **no coverage** |
| refused **and** raised close not reported as a close | **no coverage on any venue** |
| close on a flat account doesn't reach the venue | bybit/okx only |
| `lock_position_until_sltp` outranks close | bybit/okx only |
| the folded method isn't re-forked | **no coverage** |

Every assertion kills a distinct mutant. Removing the shared close branch fails binance+bitget only (bybit/okx survived on their own overrides), which pinned the test to the fix rather than to something incidental.

Two test defects found by mutation and fixed along the way: bracket-price assertions at `rel=1e-2` (which accepted a stop mispriced by up to **half the stop distance** — a +0.5% skew survived at both venues), and three binance tests whose only assertions sat inside `if trade_info["executed"]:`, so a fixture that stopped executing would have made them pass silently.

## Known and not fixed here

**alpaca is a fifth SLTP env with the same bug by a different mechanism** — `create_alpaca_sltp_action_map` drops the side marker, so `include_close_action=True` yields a close slot byte-identical to HOLD. It needs a 2-tuple format decision rather than a guard, so it is **#418**.

Suite: **4421 passed**, 16 skipped. Net **−109** lines. Part of #288.


BY571:
@CharlieHelps this PR changed shape substantially since your last look — please re-review.

It started as a 6-line helper swap. Two review agents independently caught that the swap itself drove binance's and bitget's `_execute_trade_if_needed` to **100% identical on code lines**, which made a real bug visible: `include_close_action` is on the shared config for all four venues, but only bybit/okx handle `side == "close"`. On binance and bitget a close action silently no-ops and leaves the position open.

Two things worth your attention specifically:

1. **The early return in the close branch.** bybit derives the side as `"buy" if side == "long" else "sell"`. Carried across naively that turns the silent no-op into a *wrong SELL* on an open short — bybit is only safe because of its `current_position == 0` early return. I kept the return; please confirm the ordering is right (it sits after the lock guard, before the duplicate-action guard, matching bybit).

2. **Is folding onto `SLTPMixin` the right home**, given bybit/okx keep overrides due to the `*, current_price` signature fork? The alternative is leaving two copies and fixing both.

Mutation-verified: removing the shared branch fails binance+bitget only, 4 of 8 cases.

BY571:
@CharlieHelps two more commits since your approval at `76e4969` — final look please.

`5adfa07` is test-only. `0a1e4f2` is the one worth your time, because it fixes a defect **in my own fix**:

1. **My `success=False` patch for a refused close landed on the mixin only.** That would have made binance/bitget disagree with bybit/okx in a shared field, and within the mixin a *refused* close still reported differently from a *raised* one. Now uniform across all four venues and both failure paths. This is the same "fix landed on some copies" shape the PR exists to remove, so I'd like it checked rather than trusted.

2. **I collapsed the mixin's long/short branches into one block** (68 lines → 42), matching the shape bybit and okx already ship. My safety argument is that `calculate_bracket_prices(side, ...)` raises on any side that is not long/short and runs *before* `trader.trade(...)`, so an unrecognised side can never reach the venue as the `else` "sell". Previously such a side fell through both branches as a silent no-op; now it raises. **Is raising correct there, or does it break a caller that relied on the no-op?** That is the question I am least sure about.

3. **Ownership guard added.** A faithful private re-fork of the folded method appended to binance passed all 1202 tests before; it now fails.

Also filed #418: alpaca is a fifth SLTP env with the same bug by a different mechanism (its 2-tuple action map drops the "close" marker, so the slot is byte-identical to HOLD). Stated rather than fixed, since it needs a format decision.

BY571:
@CharlieHelps blocker addressed in `0fc4859` — and you found a case I had missed, plus one I had only half-fixed.

**On the blocker.** You're right that `calculate_bracket_prices` was the wrong place for it. I had hoisted that call above the direction-switch flatten, which fixed the money hazard but left validation exactly as path-dependent as you describe. Measured across all four venues before this commit:

| venue | unsizeable + bad side | open position + bad side |
|---|---|---|
| binance | returned quietly | ValueError, no venue call |
| bitget | returned quietly | ValueError, no venue call |
| bybit | returned quietly | **ValueError, position closed first** |
| okx | returned quietly | **ValueError, position closed first** |

So my hoist had itself landed on one of three copies — bybit and okx were still flattening a live position on the way to failing. That's the third time this PR has produced a fix that reached some copies and not others, which is a decent argument for the follow-up fold of the rest of `_execute_trade_if_needed`.

`if side not in self.SIDE_DIRECTION: raise` now sits immediately after the `None`/`"close"` handling, above every venue call, on all three implementations — where you asked for it. The hoist is reverted with it: with the guard up front it protects nothing, and pricing reads better next to the trade that uses it.

**On the regression test.** Exactly the shape you specified — open position, assert `ValueError`, no `close_position()` call, position unchanged — parametrised over all four venues **and** both sizing paths, since a single path is what let this hide in the first place. Removing the guard from the mixin alone fails 4 of its 8 cases while bybit and okx keep passing on theirs, so the per-copy landing is pinned too.

**On your non-blocking follow-up** — yes, that contract was intended, and it was already fixed in `4e420e3`: a refused or raised direction-switch flatten now reports `success=False` on all four venues, and asserts that the entry which follows it is abandoned. Round 3's mutation pass had found that same `close_position()` call was unguarded everywhere (`if not close_success: return` → `if False:` survived all 3407 tests), which is the more serious half — the venue would hold both sides.

Suite: 4448 passed, 16 skipped.

Commits since your review: `eaae1c7` (the close was three copies, not two — extracted to `SLTPMixin._close_action`), `10d48f5` + `3528600` (the hoist, and restoring a line a review agent's mutation harness had deleted from my working tree mid-commit), `4e420e3` (your follow-up + the flatten-failure gap + three coverage holes), `0fc4859` (this).

PR #419, opened 2026-08-27
refactor(288): fold the last two SLTP executors; the split was one line

BY571:
The last of #288's SLTP duplication. bybit and okx each kept a ~110-line `_execute_trade_if_needed` that differed from the shared one in **exactly one thing**: where the entry price comes from. Everything downstream — guards, close, duplicate check, sizing, flatten, bracket, trade, exception handling — was already identical.

## The hook

`_bracket_entry_price` is that one thing. binance and bitget read their own candle close under the halt policy; bybit and okx use the mark `_step` already acquired and must **not** re-read it (#295, where re-reading inside the trade path bypassed the policy). Both venues pick the alternative with one line rather than a copy:

```python
_bracket_entry_price = SLTPMixin._validated_mark_price
```

That line exists because my first pass gave bybit and okx a 14-line hook each — byte-identical to one another. Two copies of a thing is how this method reached two 110-line copies, so it didn't survive its own diff.

okx's sub-minimum refusal moves to a `_resolve_bracket_quantity` override that delegates via `super()` — the same shape binance's sizing already uses, exempted from the identity guard the same way. Keyed on the overriding **class**, not the venue: keyed on the venue it also swallowed okx's two plain envs, which override nothing and would have passed while asserting nothing about them. Whether the other three should refuse sub-minimum quantities too is **#414's** question, not this fold's.

## Equivalence

324 cases per tree — 2 venues × 3 positions × 3 sides × sizeable/not × {trade ok, refused, raised} × 3 `bracket_status` shapes — run differentially against `main`:

| | cases | differing |
|---|---|---|
| real-trader (`bracket_status` absent) | 108 | **0** |
| total | 324 | **4** |

All four differences are okx with an **injected** `bracket_status` reporting a rejected leg, which no live okx trader produces — its executor never sets the attribute, so the all-True default applies and behaviour is identical.

**Correction (round 1).** I originally wrote that this *fixes* `ReplayOrderExecutor`. It does not, and the torchrl reviewer was right to refute it: replay sets `sl_placed = stop_loss is not None`, and `calculate_bracket_prices` always returns two floats, so replay reports all-True on this path too. **No trader in the repo can currently report a rejected leg here** — the only producer is the injected stub in the test.

The gate is still worth having, because binance and bitget *do* place the legs separately, and a venue that starts doing so should not need the test written for it afterwards. But it pins a contract rather than fixing an observed failure, and the test docstring now says so.

(An earlier version of this probe monkeypatched `_resolve_bracket_quantity` to force a sub-minimum quantity and reported 40 differences. That method is exactly where okx's check moved, so replacing it deleted the behaviour under comparison. The corrected probe drives a real sub-minimum quantity through real sizing.)

## One test needed repointing

`test_sltp_sizing_rejects_a_non_finite_price_or_balance` greps the resolved source for the price guard — which moved to `_bracket_entry_price`. That's the **second** time this test's subject has moved out from under it (the first was #288 folding the sizing onto the shared base; its own comment records that). It now resolves-then-scans, and pins both halves of the split: that binance/bitget read the observer, and that bybit/okx do not.

## What's left in #288 after this

| target | copies | lines | similarity |
|---|---|---|---|
| plain `_execute_trade_if_needed` | 4 | 53 | bybit/okx 100%, binance/bitget 77% |
| plain `__post_init__` | 4 | 56 | 61–80% |

Plus the test-side consolidation, [measured here](https://github.com/TorchTrade/torchtrade/issues/288#issuecomment-5437804225).

## Round 1 (five agents) — what it changed

Two agents independently found that **the guard did not follow the fold**: `_execute_trade_if_needed` became shared by all four venues while the test protecting it was still keyed on the pre-fold split. A faithful copy of the shared executor pasted back onto bybit **passed all 3445 tests** — the un-folding of the exact method this PR folds was invisible, on the two venues it just changed.

The fix was a deletion: both methods moved into the uniform `_SHARED_METHOD_OWNERSHIP` table and the 72-line `_DISPATCH_INHERITORS`/`_DISPATCH_OVERRIDERS` block went with them, since its premise ("bybit and okx legitimately override") stopped being true here. The variation moved rather than vanished, so a new identity guard pins `_bracket_entry_price` at its new home.

Three mutants that were surviving the whole suite now die:

| mutant | was | now |
|---|---|---|
| a **refused** flatten also clears the brackets | 3445 passed | 4 failures |
| a **successful** switch arms nothing | 3445 passed | 15 failures |
| `isfinite` dropped from either price guard | killed by a *grep* only | 4 failures, behavioural |

Also: `_mark_flat()` folds the flatten triplet that `_close_action` and the direction-switch flatten were doing 140 lines apart — and had **already drifted**, which is precisely the regression Charlie caught.

Net **−51**. Suite: **4485 passed**, 16 skipped.

### DIFF
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -282,7 +282,7 @@ def test_non_sltp_step_syncs_before_it_trades(env_cls):
     """
     step = env_cls._step
     sync = _first_call_position(step, {"_sync_position_from_exchange"})
-    # NOT `_dispatch_sltp_trade`: this is parametrized over NON_SLTP_ENVS, which cannot
+    # NOT the SLTP executor: this is parametrized over NON_SLTP_ENVS, which cannot
     # contain an SLTP env, so adding it read as coverage that could never bind.
     trade = _first_call_position(step, {"_execute_trade_if_needed"})
 
@@ -328,13 +328,13 @@ def test_every_live_step_validates_its_action_before_it_trades(env_cls):
     step = env_cls._step
     resolve = _first_call_position(step, {
         "_resolve_action_level", "_resolve_action_index", "_resolve_action_tuple"})
-    # `_dispatch_sltp_trade` is the SLTP trade call as of #288: the shared `_step` hands
+    # `_execute_trade_if_needed` is the SLTP trade call: the shared `_step` hands
     # off through it so bybit/okx can thread the mark. Naming only the executor made this
     # guard find no trade call at all on the four SLTP venues, which is a TypeError rather
     # than a missed ordering -- loud, but only because the list that feeds it was fixed
     # first. Keyed on the wrong name it would simply have stopped constraining anything.
     trade = _first_call_position(
-        step, {"_execute_trade_if_needed", "_dispatch_sltp_trade"}
+        step, {"_execute_trade_if_needed"}
     )
 
     assert resolve is not None, (
@@ -2694,7 +2694,7 @@ def test_the_pre_trade_read_halts_like_the_post_bar_one(exchange, module):
     # this already existed two hundred lines up; I reached for `in source` anyway.
     acquire = _first_call_position(cls._step, {"_acquire_pre_trade_state"})
     trade = _first_call_position(
-        cls._step, {"_execute_trade_if_needed", "_dispatch_sltp_trade"}
+        cls._step, {"_execute_trade_if_needed"}
     )
     assert acquire is not None, (
         f"{exchange}/{module}'s resolved _step reads the venue before trading without "
@@ -4242,10 +4242,9 @@ def test_dependency_injection_still_skips_construction_entirely():
 # would hand a spot env the futures step.
 # Per OWNER, because the two own different method sets.
 #
-# `_dispatch_sltp_trade` is deliberately ABSENT. It is the venue-variation hook -- its own
-# docstring calls it "the ONE thing `_step` varies by venue" -- and bybit and okx really do
-# override it, to thread the price rather than re-read it (#295). Guarding it would demand
-# a fold of the one method that exists not to be folded.
+# `_dispatch_sltp_trade` is gone: once `_bracket_entry_price` took over the venue
+# split it was a forwarder with one caller and zero overrides, so `_step` calls
+# the executor directly now.
 _SHARED_METHOD_OWNERSHIP = [
     # `_record_and_score` is owned by TorchTradeLiveEnv and reached by all TEN stepping
     # envs, alpaca included -- the only tail in this file that alpaca shares (#288).
@@ -4268,7 +4267,7 @@ _SHARED_METHOD_OWNERSHIP = [
         # bybit+okx split that no longer exists -- so they guarded nothing on the two
         # venues the fold had just changed. A faithful copy of the executor pasted back
         # onto bybit passed all 3445 tests while those guards were still in place.
-        "_dispatch_sltp_trade", "_execute_trade_if_needed",
+        "_execute_trade_if_needed",
     )),
 ]
 # By NAME, not by count. Dropping `_reset` from the matrix above passed 936 tests: the
@@ -4286,7 +4285,6 @@ assert {(owner.__name__, method) for _, owner, method in _SHARED_METHOD_OWNERSHI
     ("SLTPMixin", "_record_sltp_position"),
     ("SLTPMixin", "_reset_sltp_state"),
     ("SLTPMixin", "_close_action"),
-    ("SLTPMixin", "_dispatch_sltp_trade"),
     ("SLTPMixin", "_execute_trade_if_needed"),
 }
 
@@ -4819,7 +4817,7 @@ def test_a_refused_sltp_bracket_does_not_write_a_phantom_position(trade_info,
     duplication costs, in the gate rather than in the code.
     """
     env, trader = _real_futures_env(budget=0, venue="binance", sltp=True)
-    env._dispatch_sltp_trade = lambda action_tuple, current_price: trade_info
+    env._execute_trade_if_needed = lambda action_tuple, current_price=None: trade_info
 
     td = env.reset()
     assert env.position.current_position == 0, "setup: expected to start flat"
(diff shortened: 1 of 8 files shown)


BY571:
@CharlieHelps review please — this is the fold you'd expect after #417, and it closes most of what's left in #288.

bybit and okx each kept a ~110-line `_execute_trade_if_needed` that differed from the shared copy in **exactly one thing**: where the entry price comes from. That's now a `_bracket_entry_price` hook.

Three things I'd particularly like challenged:

1. **The equivalence claim.** I ran 324 differential cases per tree against `main` (2 venues × 3 positions × 3 sides × sizeable/not × {trade ok, refused, raised} × 3 `bracket_status` shapes). 108/108 real-trader cases identical; 4 differ, all okx with an *injected* `bracket_status` reporting a rejected leg. I claim that's unreachable with a live okx trader — its executor never sets the attribute, so the all-True `getattr` default applies — and that it's a **fix** for `ReplayOrderExecutor`, which does set it and which okx was ignoring while recording a stop as live. Both halves worth attacking.

2. **`_bracket_entry_price = SLTPMixin._validated_mark_price`** as a class-level assignment on bybit and okx. It exists because my first pass gave each venue a 14-line hook that was byte-identical in both — which is how this method got to two 110-line copies in the first place. Is the assignment clear enough, or too clever versus a 3-line method?

3. **#295 specifically.** The reason bybit/okx take a threaded mark is that re-reading the price inside the trade path bypassed the halt policy. Please confirm the fold hasn't reintroduced a second read for them, and that binance/bitget still read under `_halting` with their own cache slot.

Also: okx's sub-minimum refusal moved to a `_resolve_bracket_quantity` override delegating via `super()`, mirroring binance's existing sizing override. I deliberately did **not** promote it to all four venues — #414 says it was split out of #288 so this issue can close on deduplication alone. Tell me if you'd rather see that decided here.

Net −138. Suite 4451 passed.

BY571:
@CharlieHelps blocker fixed in `b3992d3`. You were right, and the way my verification missed it is worth recording.

**The regression.** Confirmed exactly as you described:

| | bybit | okx |
|---|---|---|
| `main` | sl=90.0 tp=110.0 | **sl=0.0 tp=0.0** |
| the fold | sl=90.0 tp=110.0 | **sl=90.0 tp=110.0** ← regressed |
| now | sl=0.0 tp=0.0 | sl=0.0 tp=0.0 |

Cleared in the shared path after the confirmed flatten, which restores okx and closes the same gap for the other three — they never had it.

**Why 324 differential cases didn't catch it.** The probe initialised `active_stop_loss`/`active_take_profit` to `0.0` — the same value a correct clear produces. So "cleared" and "never set" were indistinguishable in every case it ran, and it reported a clean result while the regression was sitting there. A probe that seeds state with the value the code under test would write cannot observe that write. Fixed to seed 90/110 and re-ran:

- real-trader cases: 108, differing 8
- total differing: 28 — bybit 24, okx 4
- **fields that ever differ: `sl`, `tp`, and nothing else.** Position, trade calls, trade kwargs, `trade_info` and the balance cache are identical across all 324.

The 24 bybit cases are this fix. The 4 okx cases are the rejected-leg fix you already confirmed.

**The regression test** seeds non-zero brackets — that being the whole point — and covers a successful flatten followed by a refused *or* raised entry, on all four venues. Removing the two clears fails all 8 of its cases.

Thanks for catching it by reading; the differential harness was never going to.

Suite 4459 passed.

BY571:
@CharlieHelps please re-review at `9d804eb`. You approved `b3992d3`; six commits have landed since, from two full review rounds. Several of them change things you already looked at, so I'd rather have a fresh pass than lean on the old approval.

**What changed, and why each mattered**

1. **`9c34d96` — the guard did not follow the fold.** `_execute_trade_if_needed` became shared by all four venues while the test protecting it stayed keyed on the pre-fold split. A faithful copy of the shared executor pasted back onto bybit **passed all 3445 tests**. Both methods went into the uniform `_SHARED_METHOD_OWNERSHIP` table and the 72-line `_DISPATCH_*` block was deleted, its premise having died in this PR.

2. **`e43fa4b` — `_dispatch_sltp_trade` deleted.** Once `_bracket_entry_price` owned the venue split, it was a forwarder with one caller and zero overrides. `_step` calls the executor directly.

3. **`6cd82f7` — the binding had a footgun, and a fail-open guard.** `_bracket_entry_price = SLTPMixin._validated_mark_price` binds the *function*, so a subclass of bybit's env calling `super()._bracket_entry_price()` silently got the **candle** version — #295, via the idiom introduced to prevent it. Replaced with a `_PRICES_OFF_THREADED_MARK` class attribute, which also lets the method sit in the uniform table instead of needing a bespoke guard. Separately, `_sync_position_from_exchange` became `_mark_flat`'s third caller, which exposed a `getattr(self, "_last_confirmed_read", {})` fail-open whose only beneficiary was a test stub.

4. **`1ff4310` / `9d804eb`** — `_mark_flat` was itself unguarded; the balance pop was pinned from one of its two call sites; and the `bracket_status` getattr default — **which is the production path for bybit and okx, since neither executor sets the attribute** — had never been executed by any test, because `MagicMock` auto-creates whatever it is asked for.

**Two corrections to things I told you earlier, both refuted by review:**

- The okx `bracket_status` change is **not** a `ReplayOrderExecutor` fix. Replay sets `sl_placed = stop_loss is not None` and `calculate_bracket_prices` always returns two floats, so replay reports all-True too. No trader in the repo can produce a rejected leg on this path. It pins a contract; it does not fix an observed failure.
- `active_stop_loss`/`active_take_profit` have **zero production readers**. The bracket-clear fix you asked for is state hygiene, not a money-path fix — worth having, but I do not want the PR reading as more than it is.

**What I'd most like challenged:** the `_PRICES_OFF_THREADED_MARK` reshape (item 3) is the only behavioural-shape change since your approval, and #295 is what it touches. Differential sweep against main is unchanged at 300/19,440 differing, all of them `sl`/`tp` only — 216 the bracket-clear fix, 84 okx's per-leg gate under an injected stub.

Suite **4493 passed**, 16 skipped. Net −39.

PR #420, opened 2026-08-31
fix(414): four ways an order got submitted that the venue would reject

### LINKED ISSUE #414: Live: four order-size validation gaps that let the env believe it placed a rejected order

BY571:
Closes #414. All four gaps share one shape: **an order the env believes it placed and the venue rejected** — invariant 2 inverted, which this repo has already paid for twice.

## 1. binance validated one quantity and sent another

The min-notional check ran on the **delta**; the direction-switch branch submits `abs(target_qty)`. On a switch the signs oppose, so `|delta| = |target| + |current|` — the larger number clears a minimum the smaller one does not.

Reproduced from @CharlieHelps' worked example (price 99.99, minimum 100, short 1.0 → long 1.0):

| | value | notional | |
|---|---|---|---|
| delta **validated** | 2.0 | 199.98 | passes |
| amount **sent** | 1.0 | 99.99 | **venue rejects** |
| env believes | | | `executed=True success=True` |

The close has already happened at that point, so the account is flat while the env reports a position.

The fix is an **ordering** change, not a second check: decide what will be sent, validate *that*, then perform the switch's close. Validating after the close is too late to refuse — the position is already gone and "not executed" would describe an account that just changed.

## 2 & 3. No notional check on the SLTP path, on any venue

`_get_min_notional` lives on binance's **plain** env; its SLTP sibling doesn't inherit it and the other three never had one. And bitget and bybit read no notional field at all despite having it in hand — CCXT normalises bitget's `minTradeUSDT` to `limits.cost.min`, and bybit's `minNotionalValue` is in the same `lotSizeFilter` it already parses.

One shared guard rather than four venue copies — four copies of a money check is what #288 spent six PRs undoing. binance gains the `get_lot_size` accessor the other three already had (from filters it already caches; **no new API call**), and okx reports `0.0` because its derivatives bind on `minSz`/`lotSz` — stated rather than absent, so the guard reads one key everywhere.

It **refuses**, never rounds up to the minimum. Rounding up allocates more than the action asked for, which is how okx's formatter turns a sub-minimum size into a real position.

## 4. The delta lot-rounding had no test

Replacing `delta = self.trader.round_quantity(abs(delta)) * sign` with `delta = delta` passed **all 4493 tests**. #352 fixed the *target* rounding for #271 and left the *delta* rounding unpinned. Invisible because every other test rounds to a step the raw value already sits on; the new test uses a 0.5 step and a 1.7 target.

## Found while fixing

The `notional` and `quantity` trade modes **returned before the check** — two of three modes validated nothing at all, the same "sent something other than what was validated" shape as bug 1. All three now pass through it.

## Verification

Seven mutants, every one of which survived the suite before this PR:

| mutant | now |
|---|---|
| restore the delta-based notional check | 1 failed |
| remove the delta lot-rounding | 1 failed |
| disable the notional refusal | 4 failed |
| round **up** to the floor instead of refusing | 4 failed |
| bitget / bybit / binance each drop its notional field | 1 failed each |

The venue mocks had to model the notional fields first — **none of them carried it**, so nothing could tell whether an executor read the value or hardcoded `0.0`.

Suite: **4504 passed**, 16 skipped.

### DIFF
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -268,7 +268,7 @@ class TestBinanceFuturesOrderClass:
         good = mock_client.futures_exchange_info
         mock_client.futures_exchange_info = MagicMock(side_effect=Exception("503 blip"))
         executor = BinanceFuturesOrderClass(symbol="BTCUSDT", client=mock_client)
-        assert executor.get_lot_size()["min_notional"] == 0.0   # nothing was ever read
+        assert executor.get_lot_size()["min_notional"] is None   # never read != no floor
 
         mock_client.futures_exchange_info = good                # the venue comes back
         assert executor.get_lot_size()["min_notional"] == 5.0, (
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -77,6 +77,15 @@ class TestBinanceFuturesOrderClass:
                     # whether the executor read it or hardcoded 0.0 (#414).
                     {"filterType": "MIN_NOTIONAL", "notional": "5"},
                 ],
+            }, {
+                # A SECOND symbol, with a different floor. With one symbol nothing could
+                # tell whether MIN_NOTIONAL is scoped to this symbol or whichever one the
+                # payload happens to list last (#414).
+                "symbol": "ETHUSDT",
+                "filters": [
+                    {"filterType": "LOT_SIZE", "stepSize": "0.01", "minQty": "0.01"},
+                    {"filterType": "MIN_NOTIONAL", "notional": "20"},
+                ],
             }]
         })
 
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -247,6 +247,26 @@ class TestBinanceFuturesOrderClass:
         assert order_executor._tick_size == 0.1
         assert order_executor._tick_decimals == 1
 
+    def test_get_lot_size_recovers_after_a_transient_exchange_info_failure(self, mock_client):
+        """binance was the only venue whose `get_lot_size` never re-fetched.
+
+        `_fetch_symbol_filters` returns early on a transient failure -- BEFORE its own
+        fail-closed "no LOT_SIZE" raise -- so one blip at construction left every cache
+        empty and nothing repaired them. Both floors were then silently 0.0 for the life
+        of the process, which is the fail-open this whole change exists to remove (#414).
+        """
+        from torchtrade.envs.live.binance.order_executor import BinanceFuturesOrderClass
+        good = mock_client.futures_exchange_info
+        mock_client.futures_exchange_info = MagicMock(side_effect=Exception("503 blip"))
+        executor = BinanceFuturesOrderClass(symbol="BTCUSDT", client=mock_client)
+        assert executor.get_lot_size()["min_notional"] == 0.0   # nothing was ever read
+
+        mock_client.futures_exchange_info = good                # the venue comes back
+        assert executor.get_lot_size()["min_notional"] == 5.0, (
+            "the floor never repaired itself; one transient failure at startup disables "
+            "the notional guard permanently"
+        )
+
     def test_get_lot_size_reports_the_venue_minimums_from_the_cached_filters(self, mock_client):
         """binance was the only executor without `get_lot_size`, so the shared sizing
         could not ask it for a notional floor and the SLTP path submitted whatever it
--- a/tests/envs/binance/test_futures_order_executor.py
+++ b/tests/envs/binance/test_futures_order_executor.py
@@ -73,6 +73,9 @@ class TestBinanceFuturesOrderClass:
                 "filters": [
                     {"filterType": "PRICE_FILTER", "tickSize": "0.10"},
                     {"filterType": "LOT_SIZE", "stepSize": "0.001", "minQty": "0.001"},
+                    # Real payloads carry this; the mock omitted it, so nothing could tell
+                    # whether the executor read it or hardcoded 0.0 (#414).
+                    {"filterType": "MIN_NOTIONAL", "notional": "5"},
                 ],
             }]
         })
@@ -243,6 +246,31 @@ class TestBinanceFuturesOrderClass:
         """Tick size should be cached from exchange info at init."""
         assert order_executor._tick_size == 0.1
         assert order_executor._tick_decimals == 1
+
+    def test_get_lot_size_reports_the_venue_minimums_from_the_cached_filters(self, mock_client):
+        """binance was the only executor without `get_lot_size`, so the shared sizing
+        could not ask it for a notional floor and the SLTP path submitted whatever it
+        had computed (#414).
+
+        The values come from filters `_fetch_symbol_filters` already walks, so this adds
+        no API call -- `futures_exchange_info` is still called exactly once, at
+        construction.
+        """
+        from torchtrade.envs.live.binance.order_executor import BinanceFuturesOrderClass
+        executor = BinanceFuturesOrderClass(symbol="BTCUSDT", client=mock_client)
+        calls = mock_client.futures_exchange_info.call_count
+
+        lot = executor.get_lot_size()
+
+        assert lot["min_notional"] == 5.0, (
+            "MIN_NOTIONAL was not captured; the shared guard would read a 0.0 floor and "
+            "refuse nothing"
+        )
+        assert lot["min_qty"] == 0.001
+        assert lot["qty_step"] == 0.001
+        assert mock_client.futures_exchange_info.call_count == calls, (
+            "get_lot_size re-fetched exchange info instead of using the cached filters"
+        )
 
     def test_round_price_without_precision(self, mock_client):
         """When tick size fetch fails, prices pass through unmodified."""
--- a/tests/envs/binance/test_torch_env_futures_sltp.py
+++ b/tests/envs/binance/test_torch_env_futures_sltp.py
@@ -507,8 +507,7 @@ class TestMultipleSteps:
         mock_observer.window_sizes = [10]
 
         mock_trader = MagicMock()
-        # A realistic lot step: `float(MagicMock())` is 1.0, which reads as a 1.0-BTC
-        # lot and floors every real quantity to zero in the shared sizing guard.
+        # Realistic lot step (float(MagicMock()) is 1.0 -- see the note above).
         mock_trader.get_lot_size = MagicMock(
             return_value={"min_qty": 0.001, "qty_step": 0.001, "min_notional": 0.0})
         mock_trader.cancel_open_orders = MagicMock(return_value=True)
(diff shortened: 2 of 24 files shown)


BY571:
**Differential sweep of the binance reordering, against `main`.** Posting it because reordering a money path is the riskiest part of this PR, and "the tests pass" is not evidence that nothing else moved.

2520 cases over the full cross-product of `current_qty` × `target_qty` × `price` × `min_notional` × `lot_step` × `close_ok`, recording the complete ordered sequence of venue calls (`close_position`, `market_order` with side and amount) plus the returned `trade_info`.

| | |
|---|---|
| cases compared | 2520 |
| differing | **28 — all of them direction switches** |
| non-switch cases changed | **0** |

Of those 28:

| | main | branch |
|---|---|---|
| closed the position | **28** | 0 |
| submitted an order | 14 | 0 |
| submitted orders that were *above* the floor | **0** | — |

Two things that matters for:

1. **No false refusals.** Every order `main` submitted in the differing set was genuinely below the venue floor, so nothing that should have traded is now refused.
2. **`main` closed the position in all 28**, including the 14 where it then submitted an order the venue rejects. That is the bug in its worst form — the close is irreversible, so refusing afterwards would describe an account that had just changed. The branch refuses before touching the venue.

A representative row (`current=-0.01, target=1.0, price=99.99, floor=100.0`):

```
main  : [close -0.01]  [order buy 1.0]     -> 1.0 × 99.99 = 99.99, below the 100.0 floor
branch: []                                 -> refused before the close
```

Sweep harness: `sweep414.py`, run against a detached `main` worktree with `PYTHONPATH` pinned and `torchtrade.__file__` asserted inside it.

BY571:
@CharlieHelps ready for review at `b02c8d8`. Closes #414 — you found three of its four gaps while reviewing #413.

**What it fixes**

1. **binance validated the delta and sent the target on a direction switch.** Your worked example (99.99 / floor 100) reproduces exactly: validated 2.0 at 199.98, sent 1.0 at 99.99, venue rejects, and the close had already happened — so the account is flat while the env reports a position.
2. **No venue checked notional on the SLTP path**, and bitget/bybit read no notional field at all despite having it in hand.
3. **The plain path on bitget, bybit and okx had no check either** — found in round 2, on the two venues this PR had just taught to fetch the number.
4. **The delta lot-rounding had no test**: replacing it with `delta = delta` passed all 4493.

**What I'd most like you to attack**

The review rounds found that **four separate times, my fix carried the defect it was fixing.** Each was caught by a round, never by the suite:

| round | the fix | the defect in it |
|---|---|---|
| 1 | the switch reorder | returned `at_target=True`, latching the action level to a direction the account didn't hold — the agent was frozen out of that direction for the rest of the episode |
| 2 | the shared notional guard | validated the **pre-rounded** quantity while the venue receives the floored one |
| 3 | the flooring fix | used a bare `math.floor(q/step)*step`, which is a whole step below what `round_quantity` actually sends — so it **refused legal orders** ($101.50 validated as $98.00) |
| 3 | the `_get_min_notional` delegation | broke `target_tol`, which binance alone derived from the notional floor; at 0.0 every lot-rounded fill reads as a divergence |

The common cause is worth naming: **every stub reimplemented my guard's own rounding**, so the tests agreed with the bug. The repo already had the answer — `round_quantity`'s epsilon exists for exactly this float behaviour, documented, one file away.

Specific things I'd value your eye on:

- **The flooring now uses `round_quantity`'s epsilon.** Is matching it the right call versus calling the executor's own rounding through a hook? Matching duplicates a rule; calling it means four different method names (`round_quantity` / `_round_amount` / `_format_size` / none on bybit).
- **okx's plain-path guard was deleted** because its executor writes `min_notional` 0.0 at all three cache sites, so it could never fire — and the test row covering it stubbed a 5.0 floor okx cannot produce. Right call, or should okx grow a real floor instead?
- **`at_target=not switching`.** The non-switch refusal still latches deliberately, and the flat case latches too. I kept that (pre-existing, all four venues, out of #414's scope) but pinned it as an explicit assertion so it's a decision rather than an accident. Reasonable line to draw?

**Also filed:** #421, seven pre-existing branches no test can reach because the mock never models the field — including bybit's dead position-side guard, which re-opens #341 on the one venue its fix missed.

Suite **4544 passed**, 16 skipped. Diff +828/−90.

BY571:
@CharlieHelps your blocker is already fixed — your review is at `cb81404`, the first of nine commits, and rounds 2 and 3 found the same thing independently.

**Your exact case, at HEAD:**

```
quantity_per_trade = 1.0005, price 99.99, floor 100.0, step 0.001
  raw notional   : 100.039995   (passes a naive check)
  venue receives : 1.000 -> 99.99   (binance rejects)
  guard returns  : None            <- refused
```

It's now added as the regression test you asked for (`test_a_boundary_quantity_is_judged_after_the_venue_floors_it`), with your numbers as one row and the same shape at another scale as the other. Reverting to raw-quantity validation fails 4.

Worth flagging what happened between your review and now, because the first fix was wrong in the opposite direction:

| commit | |
|---|---|
| `0721825` | floored before validating — your finding |
| `47b0e26` | that used a bare `math.floor(q/step)*step`, which is **a whole step below** what `round_quantity` actually sends (`0.29/0.01` is `28.999999999999996`). So it started **refusing legal orders** — $101.50 validated as $98.00. Now uses `round_quantity`'s own epsilon. |

Since your review the PR also picked up: the plain path on bitget/bybit/okx had no notional check at all; one transient exchange-info failure at startup disabled binance's guard permanently; `_get_min_notional` had two owners with opposite fallbacks (100.0 vs 0.0); and `target_tol` broke when that delegation landed.

Re-review at `HEAD` please — CI green, suite 4534 passed.

BY571:
@CharlieHelps both blockers fixed in `1fc6cb5`. You were right on the first one about a design call I'd already been offered and declined.

**1. Copying binance's epsilon was not a contract.** Your bitget case reproduces exactly:

```
quantity 0.0499999999999995   (= 5.0 / 100.000000000001)
  guard, binance epsilon : 0.050 x 100 = 5.00  -> accepted
  CCXT truncate, bitget  : 0.049 x 100 = 4.90  -> venue rejects
```

The guard now asks the executor: `quantize_quantity`, delegating to `round_quantity` on binance, `_round_amount` on bitget, `_format_size` on okx, and the identity on bybit — which sends the raw float, so flooring it at all was wrong.

The shared epsilon is **deleted**, not fixed. A round-3 agent had asked me exactly this — match the epsilon or call the executor — and I chose to match it, reasoning that four method names was worse than one duplicated rule. Four names that are each correct beats one rule that's wrong for three venues. Your example is what that costs.

**2. "Unknown" and "no floor" were the same value.** Also confirmed, and worse than described: bitget *cached* the zero, so one failed `market()` lookup disabled the guard for the life of the process.

`None` now means "not read" and the guard refuses on it; `0.0` still means "genuinely no floor" and still trades, which keeps okx working. And a failed lookup is no longer cached, so a recovered venue is picked up on the next call rather than pinned to a bad moment.

| | |
|---|---|
| unknown floor (outage) | **refused** |
| genuine 0.0 (okx) | trades |
| real floor, cleared | trades |

Both mutants die: treating unknown as 0.0 fails 1; re-implementing the floor in shared code instead of asking the executor fails 1.

On your third point — agreed, the flat-opening latch is pre-existing and I've left it, with the behaviour pinned as an explicit assertion so it's a decision rather than an accident. Not blocking this PR.

Suite **4535 passed**. Re-review when you get a chance.

BY571:
@CharlieHelps fixed in `4e8ca1c`. You were right that I closed the wrong half — I fixed the fetch-**failed** path and left the succeeded-but-**incomplete** one, which is the more common case.

All three now report `None`:

| venue | the incomplete case |
|---|---|
| bitget | `limits.cost.min` absent — CCXT doesn't populate it for non-USDT-quoted markets, so this was *most* such markets |
| bybit | `minNotionalValue` absent from an otherwise valid `lotSizeFilter` |
| binance | a `MIN_NOTIONAL` filter **present** with an empty `notional` |

And the plain paths: you're right that `float(None)` raising out of `_step` is not the fail-closed behaviour, it's an unhandled exception on the money path. All three now return the same controlled refusal the shared guard gives.

Six mutants, all previously surviving the suite — the three metadata collapses and the three plain-path guards — each fails 1 now.

One small deviation worth flagging: bitget's and bybit's `env.py` have no module logger, so those two refuse silently rather than acquire an import for one log line. binance logs. Say the word if you'd rather they all log.

Suite **4544 passed**, CI green.

For what it's worth on the pattern across your three reviews: each blocker was the same defect one level further in — validated-vs-sent quantity, then one venue's rounding rule applied to four, then unknown-vs-zero at fetch level, then unknown-vs-zero at field level. Worth me asking earlier what the *general* form of a finding is rather than fixing the instance you named.

BY571:
@CharlieHelps fixed in `$(git rev-parse --short HEAD)`. Both parts of your finding were right, and the second one is the more embarrassing.

`_get_min_notional()` has **two** callers in `binance/env.py` — `_execute_fractional_action`, which I guarded last commit, and `_calculate_fractional_position`, which I did not. And the regression test written to catch exactly this stubbed `_calculate_fractional_position`, so it bypassed binance's override and could never have reached the second site. The test now drives the real override; removing the new guard fails it.

I grepped for every caller this time rather than fixing the one you named: two in binance, zero elsewhere, both guarded.

That's the fourth review where the fix I shipped was narrower than the finding. Looking back at the four:

1. validated-vs-sent quantity → I fixed the switch branch
2. one venue's rounding applied to four → I'd matched binance's epsilon instead of asking the executor
3. unknown-vs-zero at the *fetch* level → left the *field* level
4. one call site → left the other

Each time the general form was available and I took the specific one. Noted, and thank you for four rounds of it.

Suite **4544 passed**, CI green.

Issue #421, opened 2026-08-31
Seven live-venue branches that no test can reach, because the mock never models the field


BY571:
Found by a branch-coverage sweep during #414 review (PR #420). All seven are **pre-existing**; none was introduced by that work.

They share one shape, and it is the reason #414's own gaps hid for so long: **the production code reads a payload field the test mock does not supply**, so the branch that reads it never executes and a mutation to it survives the entire suite. A `MagicMock` that auto-creates attributes, or a conftest that omits an optional key, makes the code look covered while the interesting leg is dead.

This is the same defect class as the four in #414 — there, no venue mock carried its notional field, so nothing could distinguish "reads the field" from "hardcodes 0.0".

### Ranked by money

| # | Branch | Why it matters | Severity |
|---|---|---|---|
| 1 | **okx's mark-price success path** — `okx/order_executor.py:533` | `markPx` *is* in the conftest, but the only test calling the real `get_mark_price()` deliberately omits it to test the raise; everything else stubs `trader.get_mark_price`. So nothing proves okx reads `markPx` rather than `idxPx`. Mark price feeds `unrealized_pnl_pct`, `distance_to_liquidation`, and fractional sizing. | 8/10 |
| 2 | **bybit's "unusable position side" guard is dead** — `bybit/order_executor.py:333-338` | The conftest always supplies `"Buy"`/`"Sell"`, so the `POSITION_UNKNOWN` return never runs and the `qty = size if side == "Buy" else -size` fallthrough is unguarded. **This is the exact defect #341 fixed**, and its two sibling tests cover bitget and okx only — invariant 3, untested on one of four venues. | 8/10 |
| 3 | **okx `posSide == "long"` never exercised** — `okx/order_executor.py:407` | `"short"` and `"net"` are covered; the parametrize has no `"long"`, and the one fixture with it supplies two positions so `POSITION_UNKNOWN` returns first. Hedge-mode direction. | 7/10 |
| 4 | **alpaca's fractionable-equity rounding leg is dead** — `alpaca/order_executor.py:159` | `tests/mocks/alpaca.py:80-82` defaults `MockAsset` to BTC/USD's crypto values, so `if min_qty and step` always wins. `_round_qty` is the quantity lattice for every Alpaca order; the US-equities leg has no test and `MAX_EQUITY_QTY_DECIMALS = 9` is asserted nowhere. | 7/10 |
[...]

PR #422, opened 2026-09-01
fix(418): alpaca's CLOSE action was a second HOLD, because the map threw the side away

### LINKED ISSUE #418: AlpacaSLTP: include_close_action produces an action slot identical to HOLD
Found while fixing the same bug class on binance/bitget in #417 (part of #288).

## The defect

`AlpacaSLTPConfig.include_close_action` exists (`alpaca/env_sltp.py:40`) and is passed to the map builder (`:127`). Setting it `True` produces:

```python
>>> create_alpaca_sltp_action_map([-0.02], [0.03], include_close_action=True)
{0: (None, None), 1: (None, None), 2: (-0.02, 0.03)}
```

Index 1 is the "close" slot. It is **byte-identical to index 0 (HOLD)**.

`create_alpaca_sltp_action_map` returns `(sl, tp)` 2-tuples rather than the futures `(side, sl, tp)` 3-tuples, so the `"close"` marker has nowhere to live and is dropped. Alpaca's executor then returns at its first guard:

```python
# alpaca/env_sltp.py:234
if action_tuple == (None, None) or self.position.current_position == 1:
    return trade_info
```

So enabling the flag widens the `Categorical` action space by one and hands the policy a slot that provably cannot do anything. The comment in alpaca's `_step` claiming "the close action's `(None, None)` targets flat" overclaims — it can never target flat.

## Why it is filed separately from #417

Same family as the binance/bitget bug that #417 fixes (`include_close_action` on the shared futures config emitted `("close", None, None)` that only bybit/okx handled), but a **different mechanism**. The futures venues had the marker and ignored it; alpaca loses the marker in the tuple format.

Fixing it is a design call, not a guard:

1. **Widen alpaca's map to 3-tuples** and give its executor a `"close"` branch — consistent with the four futures venues, but changes alpaca's action-map contract and every test that reads it.
2. **Reject the flag at the boundary.** Per CLAUDE.md invariant 4 ("a guard that absorbs a nonsense config is worse than no guard: it makes the nonsense silent"), raise in `__post_init__` if `include_close_action=True` on a config whose map cannot express it.
3. **Sell-to-flat via the existing path** — alpaca is long-only, so "close" means sell. There may be a cheaper encoding than a full 3-tuple migration.

(2) is the smallest correct step and could ship immediately; (1) is the consistent one.

## Reproduce

```python
from torchtrade.envs.utils.action_maps import create_alpaca_sltp_action_map as f
m = f([-0.02], [0.03], include_close_action=True)
assert m[0] == m[1]     # HOLD and CLOSE are the same action
```

## Scope note

`SLTP_FUTURES_ENVS` excludes spot by construction, so #417's new cross-venue close tests are structurally blind to this — alpaca cannot be added to them without the format decision above.

BY571:
Closes #418.

## The defect

`AlpacaSLTPTradingEnvConfig.include_close_action=True` widened the `Categorical` by one and put `(None, None)` — HOLD's own tuple — in the new slot. The policy could emit it and nothing happened.

```python
>>> create_alpaca_sltp_action_map([-0.02], [0.03], include_close_action=True)
{0: (None, None), 1: (None, None), 2: (-0.02, 0.03)}
```

## The cause was a lossy wrapper, so the fix is a deletion

`create_alpaca_sltp_action_map` called the shared `create_sltp_action_map` and then **discarded the `side` field** to hand back 2-tuples. The CLOSE marker rides in that field. That discard *is* the bug.

Alpaca now calls the shared builder directly with `include_short_positions=False` — which is what long-only means, rather than dropping the side of every action to express it. Both offline SLTP envs already used the shared builder, so alpaca was the wrapper's only consumer: **`action_maps.py` loses 37 lines and the codebase loses an action-map format.**

Everything the close path needs was already inherited from `SLTPMixin` (`_close_action`, `_mark_flat`, `SIDE_DIRECTION`), and `AlpacaOrderClass.close_position()` already existed. The dispatch is three lines, ordered as on the four futures venues: below the lock guard (a locked position exits via SL/TP only) and **above** the in-position guard — below it, CLOSE is dead again.

## Issue #418 offered three options; this is option 1, and it came out net-negative

The issue proposed (1) widen alpaca to 3-tuples, (2) reject the flag at the boundary, (3) a cheaper encoding. (2) was the smallest patch but leaves alpaca the only SLTP venue that cannot flatten. (1) turned out to *remove* code rather than add it, so the consistent option was also the smaller one.

## Review

Two full rounds of the five mandatory agents. The findings that mattered were about my own tests:

- **A test of mine asserted the builder's arguments, not alpaca.** `test_alpaca_map_has_no_short_actions` called a helper that hardcoded `include_short_positions=False`. Mutating alpaca's *real* call site to `True` left it green. The map tests now construct an env and read `env.action_map`; that mutant fails 5 tests.
- **The CLOSE history row was unconstrained.** Recording a close as a long (`1.0`) left the account correctly flat, the action trace quietly wrong, and 182 tests passing. `history.actions` feeds the reward.
- **A short tuple would have submitted a BUY.** The entry branch never checked `side`; it inferred "long" from `sl is not None`. Now raises, as the mixin does.
- **alpaca was in no structural guard** for the five `SLTPMixin` methods it now inherits — the cross-venue close tests parametrize over `SLTP_FUTURES_ENVS`, which excludes spot. Added the rows it genuinely inherits (not `_step`, which it overrides on purpose). A `_close_action` override pasted onto alpaca now fails.

Also dropped an entry condition that the reorder made unreachable (de-indenting 49 lines), routed bracket pricing through `calculate_bracket_prices` to delete a third copy of `entry * (1 + pct)`, and fixed two stale docs that still described the 2-tuple map — including the class docstring, which stated the exact claim this issue was filed about.

Folding alpaca into the shared cross-venue close tests (`_SLTP_VENUES`). Those helpers call `_real_futures_env(...)` and pass `current_price=` to `_execute_trade_if_needed`; alpaca is spot and takes neither. Generalizing that helper to spot is the next #288 slice, and `tests/envs/test_live_env_base.py` carries a recorded decision that alpaca's `_step` is absent deliberately — "deleting alpaca's `_step` as redundant would hand a spot env the futures step."

## Verification

Suite **4554 passed, 16 skipped**. Mutation-checked: removing the close dispatch, moving it below the in-position guard, flipping `include_short_positions`, recording CLOSE as a long, removing the long-only raise, and pasting a `_close_action` override onto alpaca each kill tests.

Net diff **+226/−151**, production net-negative.

## Compatibility

**Breaking: Alpaca SL/TP checkpoints trained with `include_close_action=True` must be retrained.** Action index `1` changed from an inert duplicate of HOLD to a real CLOSE. Configurations with `include_close_action=False` — the default — are unaffected, and their action-space size and index layout are byte-identical to before.

**Removed export:** `torchtrade.envs.utils.action_maps.create_alpaca_sltp_action_map` (and its `torchtrade.envs.utils` re-export). Use `create_sltp_action_map(..., include_short_positions=False)`, which returns `(side, sl_pct, tp_pct)` 3-tuples instead of `(sl_pct, tp_pct)` 2-tuples.


BY571:
@CharlieHelps please review.

Short version: `include_close_action=True` on alpaca produced an action slot byte-identical to HOLD. The cause was a wrapper — `create_alpaca_sltp_action_map` called the shared builder and then threw away the `side` field, which is where the CLOSE marker lives. So the fix is a deletion: alpaca calls the shared builder directly, and `action_maps.py` loses 37 lines.

Two full agent rounds are done (a third is running). Since your four rounds on #420 were all the same defect one level deeper than the one before, I tried to front-run that here — the three things I would most want you to attack:

1. **Is the `side != "long"` raise right, or is it defensive code against an impossible state?** `include_short_positions=False` is hardcoded at the only call site, so nothing can produce a "short" tuple today. I added the raise because the entry branch previously inferred "long" from `sl is not None` and would have submitted a **BUY** for a short tuple. CLAUDE.md forbids "defensive programming against impossible scenarios" but its invariant 4 wants nonsense loud rather than absorbed. I went with loud, matching the mixin. Tell me if that is the wrong call.

2. **I deferred joining alpaca to the shared cross-venue close tests.** `_SLTP_VENUES` derives from `SLTP_FUTURES_ENVS`, and `_fire_close` passes `current_price=` to `_execute_trade_if_needed`, which alpaca's signature does not take. I judged generalizing `_real_futures_env` to spot to be the next #288 slice rather than this PR, and added ownership-table rows as the interim structural guard. That leaves alpaca as the one SLTP venue whose close is covered only by its own local tests — is that acceptable, or is it the gap that bites?

3. **The `closed_position` key.** `_close_action` writes it; alpaca's `trade_info` initializer did not declare it, so the dict's shape depended on which branch produced it. I added it to the initializer, but nothing on the alpaca path reads it, and `test_the_trade_info_defaults_are_the_whole_contract` is futures-only.

Worth naming: three of the round-1 findings were defects in **my own tests**, not the fix — one of them a test that asserted the shared builder honours its own argument and never touched alpaca at all. Mutating alpaca's real call site left it green.

Suite 4554 passed.

BY571:
@CharlieHelps round 2 landed in `3da8327` and it changes what you are reviewing, so re-pull before you look.

**Two of the five agents independently found the same thing, and it is the finding of this PR:** the CLOSE action I was adding **would have been a permanent no-op in production, with every test green.**

Alpaca is spot. A bracket entry leaves a working TP-limit-sell and SL-stop-sell, and on alpaca those legs *reserve the shares* — so `DELETE /v2/positions/{symbol}` submitted underneath them comes back insufficient-qty. The env then does the safe thing (`close_position()` → False → `success=False`, position kept, brackets kept), which is precisely why nothing caught it: the failure mode is silence.

Three things made it invisible:

1. The four futures venues `_close_action` came from are immune for a reason that does not transfer — they close `reduce_only`, which resting reduce-only legs don't block.
2. Every *other* alpaca close site already cancels first, deliberately, since #289 — init and reset both, with a comment saying why. The in-episode CLOSE was the one that skipped it, and it is the one this PR created.
3. **No test could see it.** `MockTradingClient` had no `cancel_order_by_id`, and `submit_order` discarded the bracket legs outright. That is issue #421's shape reproduced inside my own diff, so I applied #421's own prescription — fix the mocks first, then the branch.

Fixed at the executor, not the env: the venue fact belongs at the venue layer, it makes every caller safe, and it keeps alpaca honest in the ownership table. Cancelled **by order id**, not via `cancel_open_orders()` — that one falls through to the account-wide cancel-all and would kill an unrelated symbol's orders mid-episode.

**One thing I cannot verify and would value your view on:** I can prove the code path, not Alpaca's server-side semantics. If their liquidation endpoint already cancels child legs, my cancel is one redundant API call; if it doesn't, it's the fix. It's safe under both readings, which is why I shipped it — but if you know Alpaca's actual behaviour here, say so.

Second reachable bug the same round turned up: reverting my round-1 `!= 0` entry guard to `== 1` passed all 1346 alpaca+base+mixin tests. Alpaca equities permit shorts, so a synced `-1` is reachable, and under `== 1` a long action fell through to a **full-balance market BUY on top of the short**, brackets priced for a long. Now tested.

Also in this round: consolidated the refused/raised close pair into one parametrize (the refused case was missing the bracket-survives assertion), removed a dead conjunct in `SLTPMixin`, and narrowed a docstring of mine that overclaimed which tests actually go through `_step`.

Answering my own question 3 from above: I kept `closed_position` in the initializer for dict-shape consistency with the mixin, while noting it has 16 writers and 0 readers repo-wide. If you think that field should go, it's a repo-wide cleanup rather than this PR.

Suite **4557 passed**. Mutants killed this round: cancel-before-close removed, a stuck leg aborting the close, the entry guard reverted.

BY571:
@CharlieHelps round 3 is in (`7094216`) — three rounds done, all five agents returned MERGE. Re-pull; this round changed the cancel loop again.

Three findings I would not ship past, **two of them defects in my own tests**:

**1. A test of mine passed with the code under test deleted.** `test_a_leg_that_will_not_cancel_does_not_abort_the_close` asserted only that the close succeeded, so it stayed green with the entire cancel loop removed. It now counts the attempts and asserts the warning. That is the same shape as round 1's tautology — a test that pins the *outcome* but not the *mechanism*. Twice in one PR.

**2. The warning would have fired on every successful live close.** A bracket's TP and SL are an OCO pair: cancelling one terminates the other, so the second attempt in my loop always hits an already-closed order. Warning per attempt puts a line in the operator's log on every single successful close — which is how a log learns to be ignored. The loop is now silent per attempt and re-reads once afterwards, so it warns only when a leg genuinely **survived**, which is the case that actually gets the close rejected. A clean close now asserts an empty caplog.

**3. The symbol-scoping claim was unpinned.** My comment says "by id, NOT `cancel_open_orders()`, which would kill an unrelated symbol's orders" — and swapping the loop for the account-wide `cancel_orders()` passed 186/186, because the mock's `get_orders` discarded the symbol filter. That is #421's shape for the **third** time in this PR: the mock omits a field, so the branch asserts a default instead of a behaviour. Mock now honours `symbols`; that mutant fails 2 tests.

Deletions this round: `close_position(qty=...)` — my own cancel loop made that partial path newly wrong (clears every leg, then closes only part, leaving the residual unstopped), and it had zero production callers; plus two tests asserting grid arithmetic the shared builder owns, and a redundant README sentence.

**Final mutation sweep, 8 of 8 killed:** cancel loop deleted · per-id → account-wide · warn-per-attempt · close dispatch removed · entry guard reverted to `== 1` · long-only raise removed · CLOSE recorded as a long · alpaca asks for shorts.

**Production is net −11 lines** (`+128/−139`) while adding a CLOSE action that works. Tests are `+262/−88`, which is where the regression lives.

Two things I am deliberately leaving, and want your call on:

- **`close_all_positions()` does not cancel legs.** This PR establishes "never close underneath a working leg" and applies it to one of the executor's two public close paths. No production caller reaches the other, but it is documented in `live/README.md`. Fix here, or follow-up?
- **A behavioural note for the changelog:** `include_close_action` defaults to `False`, so nothing silently changes — but a config that *had* set it `True` gains a real flatten where index 1 was previously a dead duplicate of HOLD. A checkpoint trained against the inert slot will now trade on it. Retrain.

Suite **4557 passed**, CI green.

BY571:
@CharlieHelps you reviewed `3da8327`; that exact blocker is fixed in `7094216`, which landed while your review was in flight. Please re-pull.

Round 3's own sweep found the same thing independently, by the same method you'd use: I swapped the per-id loop for the account-wide `cancel_orders()` and it passed **186/186**, because the mock discarded the symbol filter. So the safety property my comment claimed was asserting a default, not a behaviour.

Both halves of your ask are in:

1. **The mock honours `symbols`** — `tests/mocks/alpaca.py::MockTradingClient.get_orders` now filters on `request.symbols` instead of returning every `NEW` order account-wide.
2. **An unrelated-symbol order is in the close test** — `test_the_cancel_is_scoped_to_this_symbol` plants a working `ETHUSD` order alongside the `BTCUSD` bracket and asserts it is still `NEW` after the close.

Verified against **your exact mutant**, not just a nearby one — dropping `symbols=[self.symbol]` from `GetOrdersRequest` in `get_open_orders`:

```
=== mutant: drop symbols=[self.symbol] ===
FAILED tests/envs/alpaca/test_order_executor.py::TestAlpacaOrderClass::test_the_cancel_is_scoped_to_this_symbol
1 failed, 184 passed
```

And the account-wide variant you were guarding against fails 2 tests.

Thanks for the confirmation on Alpaca's close endpoint — that the child-order cancellation guarantee lives on the account-wide close-all and *not* on the single-position close is precisely what I could not verify from here, and it is what makes the cancel necessary rather than merely defensive.

Round 3 also fixed two more of my own defects, both worth naming since they are the same class as the one you caught: a test that passed with the entire cancel loop deleted (it asserted the outcome, never the mechanism), and a warning that would have fired on every successful live close, because a bracket's OCO sibling is already terminated by the time the loop reaches it.

Also answering the one I flagged: `close_all_positions()` still does not cancel legs. No production caller reaches it, but it is documented in `live/README.md`, so it is inconsistent with the invariant this PR establishes. I left it out rather than widen the diff — say if you want it here instead of a follow-up.

Suite **4557 passed**, CI green on `7094216`.

BY571:
@CharlieHelps fixed in `cbf21f7`, and you found one of my own recorded recurring defects: a **fail-open guard**.

`get_open_orders()` swallows any lookup error and returns `[]` — the right shape for a caller that just wants a list, and exactly wrong for a caller that must not confuse *"there are no legs"* with *"I could not tell."* My cancel loop read `[]` as the former and went on to close underneath legs it could not see. The re-read afterwards had the identical hole, so the warning did not fire either — two fail-opens, same line of reasoning.

`_working_orders()` now does the lookup and **raises**; `get_open_orders()` keeps its swallow for every other caller. The close fails closed on a discovery failure and never reaches the venue.

```
=== mutant: revert to the swallowing lookup ===
FAILED tests/envs/alpaca/test_order_executor.py::TestAlpacaOrderClass::
       test_a_lookup_failure_refuses_the_close_rather_than_assuming_no_legs
1 failed, 185 passed
```

The new test asserts exactly what you asked: `client.close_position` is never called when order discovery fails.

**One deliberate asymmetry I want you to sign off on rather than discover.** The *re-read* after the cancels is **not** fail-closed — it warns and proceeds, treating "cannot confirm" as "something remains". By that point the cancels have already gone out, and refusing there would trade a *possible* stranded leg for a *certain* stuck position.

**And the cost of fail-closed, stated plainly:** under an order-API outage a policy can no longer flatten this symbol, and flatten is the one risk-off action in the space. I think you have the trade-off right — closing under unknown legs strands a live sell order against a position that no longer exists — but it is a real cost and I would rather it be on the record than implied.

Both of your other calls are in:

- **`close_all_positions()` → follow-up.** Agreed, and I will file it with your framing: decide whether to delegate to Alpaca's account-wide `close_all_positions(cancel_orders=True)` and pin working-leg cancellation.
- **Compatibility note → added to the PR description**, using your wording, with the removed `create_alpaca_sltp_action_map` export listed alongside it.

Suite **4558 passed**.

PR #423, opened 2026-09-01
fix(416): the SLTP balance verdict was one frame above the closure that needed it

### LINKED ISSUE #416: SLTP sizing: an unusable balance is cached as the confirmed read, and never counts as an outage
Found while folding the SLTP bracket sizing (#415). **Pre-existing on `main`** — the four pre-fold copies had the same shape — and deliberately not fixed in that PR, because changing halt semantics inside a fold makes the fold unreviewable.

Two consequences of one root cause. The SLTP sizing validates the balance **after** `_halting` returns; the plain path (`_calculate_fractional_position`) validates **inside** the closure `_halting` wraps.

### 1. The unusable balance is stored as the last confirmed read

Because the closure returns successfully, `_halting` reaches its store and caches the value. Only then does the guard reject it.

```
SLTP env, total_margin_balance = 0.0:
  _resolve_bracket_quantity      -> None          (correctly refuses)
  _last_confirmed_read['balance'] -> {'total_margin_balance': 0.0, ...}   <-- CACHED

plain env, same input:
  raises LiveObservationHalt
  _last_confirmed_read            -> {}           <-- nothing stored
```

So on an SLTP env with `max_unknown_status_steps > 0`, a `0.0`-equity read taken **while the venue is liquidating you** becomes the value the grace period serves for the rest of the outage. `_reset` reads `available_balance` from that same slot.

This is the "guard hoisted one frame up" defect that `test_an_unusable_balance_cannot_size_a_trade` exists to prevent on the plain path — #295's shape, with a quieter symptom.

### 2. It never counts as an outage

The read itself did not fail, so `_status_unknown_this_step` is never set and `consecutive_unknown_status` never increments. A persistently broken balance feed fails **every** SLTP open forever, with no `status_unknown` in the observation and no truncation — where the plain path would eventually truncate.

### The fix, and the tension

Moving the validation inside the closure fixes both at once. But `_halting` catches `ValueError` specifically so an impossible account state becomes a **halt**, and the SLTP path's current contract is to report a **failed trade** instead. So the fix has to decide which is right for sizing a *new* bracket from flat:

- **Refusing the trade** is arguably safer here: no position opens on garbage data, and the function returns before touching any existing position (the opposite-direction close happens after quantity resolution on all four venues).
- **Halting** is consistent with the plain path and gets the outage counters, the `status_unknown` flag, and eventual truncation.

Not obvious, which is exactly why it should not ride along in a refactor.

### A related divergence found alongside it

okx's min-qty short-circuit sits **before** the close-opposite-position block; binance/bitget/bybit close first and then fail the open. So on a direction switch where the new quantity is sub-minimum, **okx leaves the old opposite position open and the other three leave the account flat** — same input, two account states, by venue. Untested either way.

Also worth recording: okx's `_format_size` **clamps up** to `min_qty` (`0.0001 -> 0.001`, a 10× position), which is why okx alone needs that check — binance raises, bitget floors, bybit sends the raw value. Deleting okx's check passes the entire suite.

BY571:
Closes #416.

## The codebase already described this bug

Three functions above the defect, the plain path carries this comment:

```python
# The VERDICT is inside the closure, not just the read. `_halting` catches
# ValueError precisely so an impossible account state becomes a halt; raising one
# frame above it sent that straight out of `_step`.
```

`_resolve_bracket_quantity` **was** that hoisted guard. It read the balance under `_halting` and judged it *after* the call returned — and the read itself never fails, because `0.0` is a perfectly valid dict value. So the closure returned successfully, `_halting` reached its store, and the value the guard was about to reject became the **last confirmed read** — the state a grace period then serves. An equity of `0.0` is what a venue reports *while it is liquidating you*.

Confirmed by execution against `main` before the fix was written, not inferred.

## What the fix turned into

Round 1 showed the real shape: this is one rule written down three times, with two copies drifted. The three writers of the `cache_key="balance"` slot are now one `_read_sizing_balance()`:

| writer | before |
|---|---|
| `_calculate_fractional_position` | verdict inside the closure — correct |
| `_resolve_bracket_quantity` | verdict *after* `_halting` returned — the bug |
| `_reset` | bare callable, **no verdict at all** |

Folding them surfaced two more defects that only became visible with the copies side by side:

- **`_reset` seeded the slot unvalidated.** `_halting`'s grace path returns the cached value without re-running the read, so a value stored at reset was served, unvalidated, into sizing on a later failed read.
- **`float()` before `math.isfinite`.** The plain copy lacked it, and `math.isfinite("1000.0")` is a `TypeError`, not `False` — so any adapter reporting numbers as strings crashed the sizing path. `TypeError` is not in `_halting`'s catch tuple, so it left the halt policy entirely.

## The regression this PR introduced, and then fixed

Making the verdict a `ValueError` *inside* the closure fixed the caching — and made an **empty account indistinguishable from an unreadable venue**. `_halting` caught it, found a healthy balance from an earlier bar, and sized a live order against pre-liquidation equity. Measured, binance, `budget=2`, venue reporting `0.0`:

```
main        plain -> sized 97.96      sltp -> None (refused)
mid-PR      plain -> sized 97.96      sltp -> sized 97.96    <- propagated to a second path
final       plain -> HALT             sltp -> HALT
```

Grace exists for a venue that cannot be **read**. This one answered. The fix is one line — evict the cache slot before raising, so grace has nothing to serve. Verified across **16/16** combinations (4 venues × plain/SLTP × budget 0/2).

**Note the blast radius: the plain futures path had this on `main`**, and is fixed by the same line. Operators running `max_unknown_status_steps > 0` will see episodes halt where they previously continued — that is the point.

## Two claims from an earlier version of this description, withdrawn

Both were wrong and are corrected here rather than quietly dropped:

- **"`_reset` can read the poisoned slot."** It cannot. `_reset_outage_state()` clears `_last_confirmed_read` first, so an episode cannot start on cache. The real exposure was always mid-episode.
- **"This path advances the outage counter toward truncation."** It does not. Since the eviction an unusable balance is not grace-eligible, so the halt always leaves `_step` and `_finalize_step_flags` — which owns the counter — never runs. The bar is flagged and then halts; that is all. For a genuine **read failure** with a warm cache the chain does work end to end (measured: `bar1 unknown=1 counter=1`, `bar2 truncated=True`), and is already pinned by `test_a_real_env_flags_the_outage_and_truncates_on_the_budgeted_bar`.

The test named for that behaviour is `..._flags_the_bar_before_it_halts`, which is what it actually asserts.

## Tests

- `test_a_bracket_that_cannot_be_sized_is_a_failed_trade_on_every_venue` produced its refusal from a `0.0` balance, which is now account-level. It refuses at the **order** level instead, so it still pins the contract it was written for.
- `test_an_unusable_balance_refuses_to_size_a_bracket` asserted the asymmetry as *"pre-existing and deliberate"*. That is no longer true; replaced by a halt test.
- The two consequences are **two functions, not one with two asserts** — #416 broke both at once, which is exactly the case a merged test hides.
- The grace test drives `_execute_trade_if_needed` / `_execute_fractional_action`, not the sizing helpers: `assert not trader.trade.called` against a helper that cannot submit an order asserts nothing.

## Verification

Four review rounds (three plus a verification round on the commits none had seen). Mutants killed: verdict hoisted back out (**10**), `_reset` back to the bare callable, `float()` dropped at three separate sites, and the eviction removed (**2**, one per path).

Suite **4554 passed, 16 skipped**. Production is net **−3** lines.

### DIFF
--- a/docs/environments/online.md
+++ b/docs/environments/online.md
@@ -831,7 +831,7 @@ emergency flatten:
 | position read at the top of `_step()` (position sync, duplicate-action guard) | `PositionUnknownError` |
 | the initial read inside `_reset()` | `ValueError`, or `PositionUnknownError` if the position is the unreadable part |
 | `_current_mark_price()` at the **sizing** call sites in `_execute_fractional_action` / `_execute_trade_if_needed` | `ValueError` |
-| the sizing balance / portfolio checks | `ValueError` |
+| the sizing balance / portfolio checks | `LiveObservationHalt` on the four futures venues, whose verdict is inside the `_halting` closure ([#416](https://github.com/TorchTrade/torchtrade/issues/416)); bare `ValueError` on alpaca, which has no halt policy |
 | the candle close that prices SL/TP brackets | `ValueError` |
 
 So `except LiveObservationHalt` covers the two state reads and nothing else. Catch
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -4500,45 +4500,104 @@ def test_an_unusable_balance_halts_the_sltp_sizing_too(balance):
         env._resolve_bracket_quantity(100.0)
 
 
-def test_a_string_balance_sizes_rather_than_crashing_the_step():
+@pytest.mark.parametrize("sltp", [False, True], ids=["plain", "sltp"])
+def test_a_string_balance_sizes_rather_than_crashing_the_sizing_path(sltp):
     """`float()` before `math.isfinite`, because isfinite raises TypeError on a str.
 
-    An adapter that reports numbers as strings -- which is what several venue REST APIs
-    do -- crashed the plain sizing path out of `_step` on a perfectly good balance, since
-    `math.isfinite("1000.0")` is a TypeError, not False. The conversion has to come first.
-    A string that is not a number is a different case and still halts, below.
+    An adapter that reports numbers as strings -- which several venue REST APIs do --
+    crashed sizing on a perfectly good balance, since `math.isfinite("1000.0")` is a
+    TypeError, not False. TypeError is not in `_halting`'s catch tuple, so it left the
+    halt policy entirely.
+
+    BOTH entry points, because the helper returns the adapter's dict untouched and each
+    caller re-coerces: dropping the conversion in `_calculate_fractional_position` killed
+    no test at all while the SLTP twin was covered.
+
+    Scoped to the sizing path on purpose -- `_get_observation` still hands the raw value
+    to `math.isfinite`, so a string-reporting adapter cannot yet complete a step. That is
+    a separate read with a separate rule (it must accept <= 0 so `is_bankrupt` can see it)
+    and is deliberately not fixed here.
     """
-    env, trader = _real_futures_env(budget=0, venue="binance", sltp=True)
+    env, trader = _real_futures_env(budget=0, venue="binance", sltp=sltp)
     env.config.trade_mode = "fractional"
     trader.get_account_balance.return_value = {
         "available_balance": "1000.0", "total_margin_balance": "1000.0",
         "total_wallet_balance": "1000.0", "total_maintenance_margin": 0.0,
     }
 
-    assert env._resolve_bracket_quantity(100.0) > 0
+    if sltp:
+        assert env._resolve_bracket_quantity(100.0) > 0
+    else:
+        assert env._calculate_fractional_position(1.0, 100.0)[0] > 0
 
     trader.get_account_balance.return_value["total_margin_balance"] = "n/a"
     with pytest.raises(LiveObservationHalt):
-        env._resolve_bracket_quantity(100.0)
+        if sltp:
+            env._resolve_bracket_quantity(100.0)
+        else:
+            env._calculate_fractional_position(1.0, 100.0)
 
 
-def test_reset_cannot_seed_the_balance_cache_with_a_value_sizing_would_refuse():
+@pytest.mark.parametrize("balance", [0.0, -5.0, float("nan")],
+                         ids=["zero", "negative", "nan"])
+def test_reset_cannot_seed_the_balance_cache_with_a_value_sizing_would_refuse(balance):
     """`_reset` writes the same cache slot the sizing paths read (#416).
 
     It used to pass the bare callable, so the slot had one writer but two standards: a
-    NaN stored at reset was served, unvalidated, into sizing on the next failed read --
+    value stored at reset was served, unvalidated, into sizing on the next failed read --
     the poisoning this issue is about, arriving through the one door that had no guard.
+
+    0.0 and -5.0 are the values that matter. NaN already aborted the reset via
+    `_get_observation`'s own isfinite guard, so it never reached a trade; zero and
+    negative equity STARTED an episode and left an unusable entry in the slot.
     """
     env, trader = _real_futures_env(budget=0, venue="binance", sltp=True)
     trader.get_account_balance.return_value = {
-        "available_balance": float("nan"), "total_margin_balance": float("nan"),
-        "total_wallet_balance": float("nan"), "total_maintenance_margin": 0.0,
+        "available_balance": balance, "total_margin_balance": balance,
+        "total_wallet_balance": balance, "total_maintenance_margin": 0.0,
     }
 
     with pytest.raises(LiveObservationHalt):
         env.reset()
 
     assert "balance" not in env._last_confirmed_read
+
+
+@pytest.mark.parametrize("sltp", [False, True], ids=["plain", "sltp"])
+def test_a_grace_budget_does_not_ride_out_an_account_the_venue_says_is_empty(sltp):
+    """Grace is for a venue that cannot be READ, not for an answer we dislike.
+
+    An unusable balance raises inside the `_halting` closure so it is never cached -- but
+    that alone made it indistinguishable from a read FAILURE, so `_halting` caught it,
+    found a healthy balance from an earlier bar, and sized a live order against
+    pre-liquidation equity. Measured on main before this fix: a venue reporting 0.0
+    produced a 97.96-unit long on the plain path.
+
+    Both paths, because the fold made it one line and the plain path had the bug first.
+    """
+    env, trader = _real_futures_env(budget=2, venue="binance", sltp=sltp)
+    env.config.trade_mode = "fractional"
+    healthy = {"available_balance": 10000.0, "total_margin_balance": 10000.0,
+               "total_wallet_balance": 10000.0, "total_maintenance_margin": 0.0}
+    trader.get_account_balance.return_value = healthy
+    env.reset()
+    assert "balance" in env._last_confirmed_read, "setup: a healthy read must be cached"
+
+    # The venue ANSWERS, and the answer is that the account is gone.
+    trader.get_account_balance.return_value = dict(
+        healthy, total_margin_balance=0.0, available_balance=0.0
+    )
+
+    with pytest.raises(LiveObservationHalt):
+        if sltp:
+            env._resolve_bracket_quantity(100.0)
+        else:
+            env._calculate_fractional_position(1.0, 100.0)
+
+    assert not trader.trade.called, "sized an order against equity the venue disowned"
+    assert "balance" not in env._last_confirmed_read, (
+        "the stale balance is still there for the next grace bar to serve"
+    )
 
 
 def _refused_sizing_env():
@@ -4569,9 +4628,15 @@ def test_an_unusable_balance_is_not_stored_as_the_last_confirmed_read():
 
 
 def test_an_unusable_balance_counts_as_an_outage():
-    """#416(2). Nothing raised means the bar is never flagged, so the outage counter
-    never advances and a permanently broken feed refuses every open forever without
-    truncating. This is the only assertion in the file that pins the flag for this path.
+    """#416(2). The read produced nothing usable, so the bar is unconfirmed and must say
+    so -- the old code returned successfully and left the flag clear.
+
+    Deliberately asserts the FLAG and not truncation. I tried to pin the whole chain and
+    could not: the counter advances in `_finalize_step_flags`, which only runs if `_step`
+    completes, and a balance-feed outage does not get that far because
+    `_get_portfolio_value` reads the balance OUTSIDE the halt policy and dies with a raw
+    RuntimeError first. So the flag is what is true here; claiming truncation would be
+    claiming a chain that is currently broken elsewhere.
     """
     env = _refused_sizing_env()
 
--- a/tests/envs/test_live_env_base.py
+++ b/tests/envs/test_live_env_base.py
@@ -4446,11 +4446,12 @@ def test_a_bracket_that_cannot_be_sized_is_a_failed_trade_on_every_venue(venue):
     indistinguishable.
     """
     env, trader = _real_futures_env(budget=0, venue=venue, sltp=True)
-    env.config.trade_mode = "fractional"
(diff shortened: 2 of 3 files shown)


BY571:
@CharlieHelps please review when you have a moment — this is small, but it changes a live-money behaviour on all four futures SLTP venues, so I want it looked at.

**The codebase already described this bug.** Three functions above the defect, the plain path carries:

> `# The VERDICT is inside the closure, not just the read. _halting catches ValueError precisely so an impossible account state becomes a halt; raising one frame above it sent that straight out of _step.`

`_resolve_bracket_quantity` **was** that hoisted guard. It read the balance under `_halting` and judged it after the call returned — and the read itself never fails, because `0.0` is a perfectly valid dict value. So the closure returned successfully and two things went wrong at once, both silent:

1. `_halting` reached its store, so the value the guard was about to reject became the **last confirmed read** — and an equity of `0.0` is what a venue reports *while it is liquidating you*. On an env with a grace budget that is what gets served for the rest of the outage, and `_reset` reads `available_balance` from the same slot.
2. Nothing raised, so the bar was never flagged unknown, the outage counter never advanced, and a permanently broken balance feed refused every SLTP open **forever** — no `status_unknown`, no truncation. The plain path truncates.

Confirmed by execution against `main` before writing the fix, not inferred.

**The question I'd most like your view on.** #416 framed this as a genuine choice — halt, or report a failed trade — and said it "should not ride along in a refactor." I concluded it isn't really a choice: the plain path already halts for this exact input, the halt policy already owns the response (grace / FLATTEN / truncation), and two answers to one question by env type *is* the divergence rather than the safety. So the paths now agree, and `None` keeps the narrower meaning it should always have had — **the venue would reject the order** (nothing sized, or below a venue minimum), which the caller reports.

If you think the SLTP path should have kept refusing instead, this is the moment to say so — it is a one-line difference and I would rather argue it now than after it ships.

One consequence I want checked specifically: under `observation_failure_policy=FLATTEN`, a path that previously just refused a trade can now trigger an emergency close. I believe that is correct (it is what FLATTEN means, and the plain path already does it for the same input) but it is the sharpest edge of the change.

Hoisting the verdict back out of the closure fails **18 tests**. Suite **4558 passed**. Diff `+79/−31`, production `+21/−9`.

BY571:
@CharlieHelps **do not merge on that approval** — I changed the exact behaviour you signed off on, after you signed off on it. Flagging rather than letting it ride.

You approved `18f75a8` and wrote:

> **HALT grace also behaves consistently:** after a previously confirmed valid balance, an explicit `0.0` marks the bar unknown but can size the SLTP order from the last valid account state while the grace budget remains. That is already the documented grace contract for trade-path balance reads, including SLTP sizing.

Two of my review agents independently flagged that same behaviour as the most serious thing in the PR, and I reproduced it before touching anything. Binance, `budget=2`, healthy reset, then the venue reports `0.0`:

```
main        plain -> sized 97.96      sltp -> None (refused)
18f75a8     plain -> sized 97.96      sltp -> sized 97.96    <- my change propagated it
```

So my "make SLTP consistent with the plain path" made SLTP consistent with a path that already had this. A 97.96-unit market buy goes out against an account the venue has just said is empty.

**The argument I acted on, which is a disagreement with your reading, not an oversight:** grace exists for a venue that cannot be **read**. A truthful `0.0` is not an unconfirmed read — the venue *answered*. Raising the verdict inside `_halting` is what made the two indistinguishable, so "the account is gone" got ridden out as if it were "the API timed out", and sizing then used pre-liquidation equity. My own docstring says `equity == 0.0` is what a venue reports while liquidating you, and the code then traded into it.

Fix is one line — evict the cache slot before raising, so grace has nothing to serve. Both paths now halt, including the plain path, which had the bug first. Removing the eviction fails 2 tests.

**You may still think the documented grace contract should win here, and if so say it** — it is one line either way, and you have the better view of what that contract is meant to promise. What I am not willing to do is merge it silently under an approval that endorsed the opposite.

Your two follow-ups are both in:

1. **FLATTEN + invalid balance + SLTP open** is pinned in `8fec557`: exactly one close request, no bracket behind it, and the balance `ValueError` preserved as the halt's cause. FLATTEN bypasses grace by design so it reads the same before and after — worth recording now that the path beside it moved.
2. **You were right about the PR body.** `_reset_outage_state()` clears `_last_confirmed_read` first, so reset cannot read a poisoned slot; the real impact is mid-episode. Correcting it.

One more thing I withdrew rather than shipped: I tried to pin the flag → truncation chain and could not. The counter advances in `_finalize_step_flags`, which only runs if `_step` completes, and a balance-feed outage never gets there because `_get_portfolio_value` reads the balance **outside** the halt policy and dies with a raw `RuntimeError` first. The test is gone and the docstring that implied truncation now says what is actually true. That chain is broken elsewhere and is not this PR's to fix.

Suite **4554 passed**.

BY571:
@CharlieHelps head is now `4f8d955`, and your earlier approval has been dismissed by the new commits — correctly, since I changed the behaviour it endorsed. Ready for a fresh look.

**A verification round (all five agents) ran on the commits none of us had reviewed.** Verdict: MERGE from all five, and the grace regression is confirmed dead — one of them reproduced it across **16/16** combinations (4 venues × plain/SLTP × budget 0/2). Pre-fix, budget=2 sized 97.94–97.96 units against a venue reporting zero equity on every one; post-fix all sixteen halt with the cache evicted.

It found three things, all mine, all now fixed:

1. **The assertion guarding the money bug was dead.** The new grace test asserted `not trader.trade.called` while calling `_resolve_bracket_quantity` / `_calculate_fractional_position` — neither of which can submit an order. It now drives `_execute_trade_if_needed` / `_execute_fractional_action`, and both parametrizations fail when the eviction is removed. Fail-open guards and dead assertions are two of the four defects I keep repeating; this PR has now had one of each.

2. **I argued for the eviction on the wrong grounds, and the grounds mattered.** I framed it as "grace has nothing to serve on this bar." The reviewer showed the property is stronger: eviction is a *lasting* invalidation, whereas a dedicated non-grace-eligible exception type — which reads cleaner and which I would have taken on a later pass — is only a same-bar veto. On bar N+1 an honest venue timeout would still serve the stale balance and size the same order. Reframed as the invalidation `_handle_close_action` and `_mark_flat` already perform.

3. **I got the same docstring wrong three times.** Worth stating plainly rather than quietly fixing: I first implied truncation would otherwise happen; then blamed `_get_portfolio_value` for reading outside the halt policy (it does not, and on a `0.0` balance it does not raise at all); then concluded the chain was broken. It is not — for a genuine **read failure** with a warm cache it works end to end (measured: `bar1 unknown=1 counter=1`, `bar2 truncated=True`), and is already pinned in `test_live_observation_failsafe.py`. What is true is narrower: since the eviction, an unusable balance is not grace-eligible, so the halt always leaves `_step` and the counter never runs. The test is renamed `..._flags_the_bar_before_it_halts` and the docstring now records only what was driven and observed, and names the wrong turns.

**The one thing I still want your ruling on is unchanged**, and it is the substance rather than the polish: you approved the behaviour where a truthful `0.0` gets ridden out under grace, calling it the documented contract. I now halt instead, on the argument that grace is for a venue that cannot be *read*, and this one answered. If you still think the grace contract should win, say so — it is one line, and the measurement above is the whole case for my side.

Note the blast radius, since it is wider than the PR title: the plain futures path had this behaviour on `main` and is fixed by the same line. That is in the PR body.

Suite **4554 passed**.

BY571:
@CharlieHelps PR body corrected — that was your only remaining blocker.

Both stale claims are gone, and rather than quietly delete them I gave them their own section, since they were load-bearing in how I originally argued this PR:

- **"`_reset` can read the poisoned slot."** Wrong — `_reset_outage_state()` clears `_last_confirmed_read` first, so an episode cannot start on cache. The real exposure was always mid-episode, exactly as you said.
- **"This path advances truncation."** Wrong — since the eviction an unusable balance is not grace-eligible, so the halt always leaves `_step` and `_finalize_step_flags` never runs. The bar is flagged and then halts. I also state the thing I got wrong in the *other* direction: for a genuine read failure with a warm cache the chain does work end to end (`bar1 unknown=1 counter=1`, `bar2 truncated=True`), and is already pinned by `test_a_real_env_flags_the_outage_and_truncates_on_the_budgeted_bar`.

The body also now states the blast radius plainly, since it is wider than the title: **the plain futures path had this behaviour on `main`** and is fixed by the same line, so operators running `max_unknown_status_steps > 0` will see episodes halt where they previously continued.

Head is `4f8d955`, two commits past the `8fec557` you agreed with. Both are test-only:

- `1064628` — the grace test's `assert not trader.trade.called` was **vacuous**: it called `_resolve_bracket_quantity`, which cannot submit an order. It now drives `_execute_trade_if_needed` / `_execute_fractional_action`, and both parametrizations fail when the eviction is removed. Production in that commit is a comment reframe only.
- `4f8d955` — renamed `..._counts_as_an_outage` to `..._flags_the_bar_before_it_halts`, because after the eviction the old name promised a chain that does not run.

Thank you for ruling on the substance rather than standing on the earlier approval. Your framing is sharper than mine was — *"an authoritative account-level failure, not an unreadable venue response"* is the distinction the code needed and the one my first attempt collapsed.

Suite **4554 passed**.

BY571:
@CharlieHelps fixed in `d836a2b`. You were right, I reproduced it before touching anything, and it is the same defect class you have now caught four times across two PRs.

`float(info["total_margin_balance"])` runs **before** the eviction. So an unparseable balance raised out of the coercion, never reached the `pop`, and `_halting` caught the `ValueError`, found the warm cache, and sized a live order from stale equity. Binance, warm `10_000` cache, budget `2`, then `"n/a"`:

```
before   plain -> trade submitted, cache still populated
         sltp  -> trade submitted, cache still populated
after    both  -> no submission, cache evicted
```

The read now reaches **one verdict for every way a balance can be unusable** — unparseable, non-finite, nonpositive — with the coercion inside it. `KeyError` stays uncaught deliberately: a venue omitting the field is a config error, not an account state.

I parametrized the existing grace test over all six shapes rather than adding a seventh test: zero, negative, nan, inf, unparseable, and `None` — that last one because it fails through `TypeError` rather than `ValueError`, which the `except (TypeError, ValueError)` now has to cover and which a string alone would not have exercised. Your exact mutant, coercion back outside the verdict, fails 4 of the 12; removing the eviction fails all 12.

Worth naming, since you keep finding it: **three agent rounds plus a verification round did not catch this.** The pattern in your four findings is identical every time — I fix the instance reported and leave the class:

1. validated one quantity, sent another → I fixed the switch branch
2. unknown-vs-zero at fetch level → I left the field level
3. one `_get_min_notional` call site → I left the other
4. **the invalid number → I left the failed conversion**

Each time the general form was available and I took the specific one. The concrete change I am making: when a finding names a guard, I now enumerate every way control can reach past it before I call it fixed — here that would have been "what else can raise between the read and the eviction", which is one line of reading.

Suite **4564 passed**. Head `d836a2b`.

BY571:
@CharlieHelps thank you for re-running it across all 48 combinations rather than taking the diff at its word — that is the check I should have run myself before claiming the first eviction was complete.

One procedural thing before I merge: your formal review state on this PR is still **CHANGES_REQUESTED** at `4f8d955`, and your clearance above is scoped to "this defect class". On #422 you gave an explicit **MERGE**, so I would rather not read this one as equivalent and merge past a standing change-request on a live-money path.

Everything I know of is green at `d836a2b`:
- your 48/48 verification
- suite 4564 passed, 16 skipped
- both GitHub checks passing
- four review rounds (three plus a verification round on the commits none of us had seen)
- the PR body corrected, with both withdrawn claims kept visible rather than deleted

**Is that a MERGE?** If so I will squash it and close #416. If you would rather look at anything else first — the `None`/`TypeError` route is the newest and least-travelled part of the verdict — say so and I will hold.

PR #424, opened 2026-09-01
fix(421): seven branches no test could reach, two of them real bugs

### LINKED ISSUE #421: Seven live-venue branches that no test can reach, because the mock never models the field

BY571:
Closes #421.

Seven branches that no test could reach, because the production code read a field the mock never supplied. **Every one was verified dead by mutation before a test was written, and every one now dies.**

| # | branch | before | after |
|---|---|---|---|
| 1 | okx reads `markPx`, not `idxPx` | 0 | 1 |
| 2 | bybit's "never guess a direction" guard | **0** | 4 |
| 3 | okx signs a hedge LONG from `posSide` | **0** | 1 |
| 4 | alpaca's US-equities rounding leg | 0 | 3 |
| 5 | okx clamps a sub-minimum quantity UP | 0 | 1 |
| 6 | bitget's mark-price fallback | 0 | 1 |
| 7 | bitget hedge-mode `tradeSide` | 0 | 2 |

## Two were real bugs, not gaps

**bitget reported a mark price of `0.0` from a method documented to raise.** Two routes, both unexercised because the conftest supplied every field: `if mark_price:` passes the *string* `"0"` (a non-empty string is truthy), and `float(ticker.get('last', 0))` returns `0.0` when both fields are absent. A zero mark divides into fractional sizing and prices both bracket legs. binance, bybit and okx all raise; bitget's own docstring said it did too. The `last` fallback is correct and kept — what changed is that the resolved value is validated rather than trusted.

**Signing a hedge long as a short passed the entire suite** on okx, and **bybit's direction guard could not be reached at all**. Both are invariant 1 — *a direction is never guessed* — and both are #341's exact defect surviving on venues its fix didn't cover.

## How bybit was missed is the useful part

bitget and okx each had their **own** test. bybit had a sentence in okx's docstring:

> *"bitget and bybit at least degraded to POSITION_UNKNOWN."*

A claim about a third venue, in prose, inside a test of the second — which is worth **less** than no coverage, because it reads like coverage.

So the fix is one parametrized test over a `_DIRECTION_READERS` table, each entry driving the venue's real adapter over a stubbed client, plus an assertion that the table covers every futures venue except binance (exempt: it reports a signed quantity and never names a side). **Dropping a venue from the table is a collection error, not a silent shrink.** okx's dedicated test folds in — same values, same assertion — with its lesson kept, for a net −32 lines of test and strictly more coverage.

## One place the issue's reasoning was wrong

#421 said okx's min-qty clamp was unreachable because the conftest sets `minSz == lotSz`. That isn't why — the clamp fires whenever `qty < step`. The mutation is what settled it, not the reasoning. Worth noting because the clamp is the *premise* of `_refuse_below_venue_minimums`: okx turns a `0.0001` ask into a `0.001` order — a 10× position — where binance raises, bitget floors and bybit sends the raw value. That premise was itself untested. The new test also pins that it still **floors** above the minimum, so a future "just always clamp" cannot pass.

## Verification

Suite **4600 passed, 16 skipped**. Production change is confined to bitget's `get_mark_price` (+ the `math` import that file lacked — whose absence turned a `NameError` into `RuntimeError: Failed to get mark price: name 'math' is not defined`, a message that reads like a venue failure).


BY571:
@CharlieHelps please review. This one is mostly tests, but two of the seven were real bugs.

**Every branch was verified dead by mutation before a test was written, and every one now dies.** All seven were at 0 failures:

| # | branch | before | after |
|---|---|---|---|
| 1 | okx reads `markPx`, not `idxPx` | 0 | 1 |
| 2 | bybit's "never guess a direction" guard | **0** | 4 |
| 3 | okx signs a hedge LONG from `posSide` | **0** | 1 |
| 4 | alpaca's US-equities rounding leg | 0 | 3 |
| 5 | okx clamps a sub-minimum quantity UP | 0 | 1 |
| 6 | bitget's mark-price fallback | 0 | 1 |
| 7 | bitget hedge-mode `tradeSide` | 0 | 2 |

**The one production change**: bitget returned a mark price of `0.0` from a method whose docstring says it raises. Two routes, both unexercised because the conftest supplied every field — `if mark_price:` passes the *string* `"0"`, and `float(ticker.get('last', 0))` fires when both fields are absent. A zero mark divides into fractional sizing and prices both bracket legs; binance, bybit and okx all raise. The `last` fallback is right and kept; the resolved value is now validated rather than trusted.

**Two were direction bugs.** Signing an okx hedge long as a short passed the entire suite, and bybit's direction guard could not be reached at all. Both are invariant 1 and both are #341's defect surviving on venues its fix did not cover.

Three things I'd particularly like you to attack:

1. **Are these tests in the right file?** I put them all in `tests/envs/test_live_env_base.py`, which is for cross-venue contracts — but alpaca's equity rounding and bitget's hedge params are single-venue and arguably belong in the per-venue files. The cross-venue direction test clearly belongs there; the others I am less sure about.
2. **`_DIRECTION_READERS` asserts it covers every futures venue except binance** (exempt: it reports a signed quantity and never names a side). Is that exemption right, or does binance have an equivalent guess I have not found?
3. **My `MockAsset(fractionable=True, min_order_size=None, min_trade_increment=None)`** — I believe that is what the real Alpaca API returns for a US equity, since those fields are crypto-only. If that shape is wrong, the test is fiction.

One correction to the issue itself, since it was mine: #421 claimed okx's min-qty clamp was unreachable because the conftest sets `minSz == lotSz`. That is not why — the clamp fires whenever `qty < step`. The mutation settled it, not the reasoning.

Suite **4600 passed**. Three agent rounds are running.

BY571:
@CharlieHelps round 1 changed this materially — re-pull before reviewing. Head is `5d6a53c5`.

**The finding is the issue's own failure mode, reproduced inside its fix.** My bitget comment asserted *"every other venue raises rather than reporting one."* Two reviewers checked it independently against the real adapters:

```
bybit   markPrice="0" -> 0.0
okx     markPx="0"    -> 0.0
binance markPrice="0" -> 0.0
```

All three had the same bug — bybit and okx via `if mark_price:` (a non-empty string is truthy), binance with no guard at all. That is exactly what let bybit's dead direction guard survive in this same PR, via okx's docstring vouching for it in prose. I then wrote the identical shape myself, one commit after fixing it.

So the rule now has one home: `_validated_mark` on `ExecutorHelpersMixin`, which all four executors already mix in and whose docstring asks for precisely this. **The four fetches stay four** — ccxt `info`, pybit `result.list`, okx `data[]`, binance flat dict are different shapes with different error envelopes; only the verdict is shared. And a `_MARK_READERS` table makes the claim tested rather than asserted.

**A correction to what I told you in my first comment.** I said a zero mark "divides into fractional sizing and prices both bracket legs". That is not true today: `_current_mark_price` already rejects a non-positive mark (#347) and `_acquire_post_bar_state` guards its own read, so it cannot reach sizing. This is a contract fix and defence in depth across four venues, not a live money bug. Flagging because it was the strongest sentence in my pitch and it was wrong.

Also since round 1:
- Two tests of mine passed for narrower reasons than their docstrings claimed — inverting bitget's mark/`last` **priority** passed all seven mark tests (no row had both fields with different values), and removing alpaca's `ROUND_DOWN` passed the 9dp row because `approx`'s default `rel=1e-6` is coarser than the truncation it was meant to observe. Both mutants now fail.
- The `try` in bitget and binance is narrowed to the fetch, so an expected refusal is no longer logged at ERROR and re-wrapped into `"Failed to get mark price: <the refusal>"`. That broad `try` is what turned a missing `math` import into a message reading like a venue failure.
- Answering my own question 1 to you: the single-venue tests moved out of the cross-venue contract file into their venue files (−64 lines). The mark tests stayed, because they became cross-venue this round.

**Still open for you**, and both are places where I could be wrong rather than sloppy:
1. `_DIRECTION_READERS` exempts binance on the grounds that it reports a **signed** quantity and never names a side. Is that exemption right?
2. `MockAsset(fractionable=True, min_order_size=None, min_trade_increment=None)` — I believe that is what the real Alpaca API returns for a US equity, since those fields are crypto-only. If the shape is wrong, the test is fiction.

Suite **4609 passed**. Round 2 of the agents is running.

BY571:
@CharlieHelps thank you for the three answers — all three settled real uncertainty, and one of them I would have got wrong on my own.

**Please re-review: your approval is at `024c3ae`, four commits stale, and two of those are production changes on live-money paths.** Head is `b7f02032`. I am not merging on the stale approval.

**The one that matters most, and it is the biggest finding in the PR:** a branch-coverage sweep of binance — the venue #421 barely covered — found that **binance never received #341's fix.** Reproduced against the real adapter before changing anything:

```
healthy                  -> qty=0.001 leverage=10.0
liquidationPrice ''      -> POSITION_UNKNOWN
liquidationPrice None    -> POSITION_UNKNOWN
entryPrice ''            -> POSITION_UNKNOWN
unRealizedProfit ''      -> POSITION_UNKNOWN
```

bitget, bybit and okx all parse optional position fields as `float(x or 0)`. binance alone uses bare `float(pos["entryPrice"])`. POSITION_UNKNOWN is deliberately fail-closed, so the env freezes every bar **on a lie**, and at the default outage budget of 0 the episode truncates while still holding a live levered position. This is binance's **ordinary cross-margin shape** — it blanks `liquidationPrice` there — and `.get("liquidationPrice", 0)` never helped because the key is present-and-empty.

Two structural fixes alongside it, since "one venue missed the fix" is what both #341 and #421 are about: the null-field test covered **bitget only** (one venue's test for a four-venue rule), and the leverage guard was a **source-text scan** for the literal `"or self.leverage"` — the same swap spelled `if not pos.get("leverage")` walked straight past it, so it is now behavioural.

Worth your scrutiny, because it nearly shipped inert: my first version of that leverage test used the string `"0"` and **the mutant survived**. These venues send numbers as strings, `not "0"` is False, so a string zero cannot distinguish the correct guard from the falsy one. Only a numeric zero separates them. The test is now parametrized over both.

Also since your approval:
- My bitget mark-price comment claimed *"every other venue raises rather than reporting one."* False on all three — bybit, okx and binance all returned `0.0` for `markPrice="0"`. The verdict now lives once as `_validated_mark` on `ExecutorHelpersMixin` with all four routed through it; the four fetches stay four.
- Same class-vs-instance miss again: I fixed okx's `markPx`/`idxPx` confusion and never asked whether binance and bybit had it. They both ship `markPrice` alongside `indexPrice` in the same payload. Now a three-venue test; the swap dies on each.
- A **fail-open assertion** of mine — `pytest.raises((RuntimeError, ValueError))` where nothing raises ValueError.
- I corrected the strongest claim in my original pitch: a zero mark **cannot** reach sizing, since `_current_mark_price` already rejects it (#347). Defence in depth, not a live sizing bug.

On your two non-blocking notes: I took the bitget-hedge one as a follow-up rather than here, and left the okx direction fixture in the default mode for exactly the reason you gave.

Suite **4610 passed**, both checks green at `b7f02032`.

BY571:
@CharlieHelps re-review needed — head is `cb47c079`, four commits past your approval, and **one of them reverts a regression I introduced.**

**The finding, and it was mine.** In `b7f02032` I gave `positionAmt` the same `float(x or 0)` treatment as the fields beside it. That was wrong: it is not an attribute of a position, it is the field that decides whether one **exists**. Measured against the real adapter with a live 1.0 BTC position open:

```
positionAmt healthy -> qty=1.0
positionAmt ''      -> None == FLAT
positionAmt None    -> None == FLAT
```

Before my change both raised and reported `POSITION_UNKNOWN`. After it, a blank routes to `position_status = None`, `_sync_position_from_exchange` clears `current_action_level` and discards `hold_counter`, and the policy's next long opens a **second** position on top of the live one. POSITION_UNKNOWN freezes; that doubles exposure. Invariants 3 and 4 in one line.

It also bought nothing — binance always sends a numeric string there. `liquidationPrice` is the field genuinely blank on cross margin, and it keeps its `or 0`. And it was **invisible**: reverting the line alone left the entire suite green. My change had made binance the only venue whose existence field failed open; bybit, okx and bitget all keep theirs fail-closed.

`_EXISTENCE_FIELDS` now pins it on all four venues across both blank shapes, and `close_all_positions` — which carries the same bare subscript 150 lines down — agrees with `get_status` again instead of diverging.

Also in these four commits, all found by mutation and all "one venue short", which is this PR's recurring shape:

| mutant | before | after |
|---|---|---|
| okx leverage swap spelled without `or` | **0** | 1 |
| bybit leverage swap spelled without `or` | **0** | 1 |
| bybit mark chain simplified back to `or` | **0** | 1 |
| `positionAmt` fail-open restored | **0** | 2 |

The first two survived because `_NULLABLE_POSITION_FIELDS` stopped at the venues *known* to be broken — bybit and okx were correct-but-uncovered, which is the more dangerous state and precisely how binance went without #341's fix. The table now carries the completeness assert its siblings have, plus one on the field axis, because the venue assert cannot see a venue whose field list is emptied.

One test bug of mine that failed **loudly** rather than passing, which I'd rather you saw than didn't: okx names the leverage field `lever`, so my `leverage=0` override added a key it does not read and the base `"10"` stood. The key is a table column now.

Two things I am deliberately leaving out and want on the record rather than buried: `notional_value`'s truthiness rule is obeyed on binance and broken on the other three (`"0"` is truthy, so the `or abs(qty*mark)` fallback never fires and `account_state[0]` reads exposure 0 while a position is held), and binance's `close_position()` side is still untested. Both are pre-existing and separable; say if you would rather they rode along.

Suite **4660 passed**.

PR #426, opened 2026-09-02
fix(288): two venues re-traded on every price move, because the fold never happened there

BY571:
Closes the last production row of #288 — and it turned out not to be duplication.

## The finding

`_execute_trade_if_needed` reads as the tidiest remaining cluster in #288's table: 4 copies, 53 lines, "bybit/okx byte-identical". The similarity metric is right and the conclusion it invites is wrong. Those four copies implement **two different behaviours**, and pairwise similarity cannot see which half is correct:

| | binance / bitget / alpaca | bybit / okx |
|---|---|---|
| guard | `desired_action == current_action_level` → skip | none; recompute the target at the new mark, trade the delta |
| repeated action, price moved | **hold** | **rebalance** |

So a policy that repeats one action **resizes on every price move** on two of four venues. Measured end to end — full long held while the price walks 100 → 110, with the mocked venue reporting back exactly the qty the env submitted:

```
venue     |   100    101    103    106    110 | re-trades | extra notional
binance   | TRADE   .     .     .     .       |     0     |     $0.00
bitget    | TRADE   .     .     .     .       |     0     |     $0.00
bybit     | TRADE TRADE TRADE TRADE TRADE     |     4     |   $946.77
okx       | TRADE TRADE TRADE TRADE TRADE     |     4     |   $946.83
```

…against a ~$9,795 position. They are resizes, not round-trips, so ≈$0.52 in taker fees per 10% move — small each time, and it recurs on every drift.

**The tiebreaker is not tidiness.** `offline/sequential.py:697` holds:

```python
# If action hasn't changed and we already have a position, just hold
if action_value == self._prev_action_value and self.position.position_size != 0:
```

That is the env the policy **trains in**, and CLAUDE.md names offline correctness the highest priority. Live, the learned "hold" silently meant "rebalance" on bybit and okx, at fees the training reward never charged. bybit and okx are the outliers, not the majority.

## Why it drifted, and why the original reason no longer holds

This was not a decision taken twice. `80bdd892` (2026-02-14) removed the guard from bybit deliberately — *"rely on exchange state … robust to state drift from rejected orders, partial fills, or external intervention"* — and okx was written from bybit two months later, inheriting the absence.

That concern was real, and is now handled a layer up by machinery that did not exist in February: `_sync_position_from_exchange` compares the venue's qty against the target and NaNs `current_action_level` on a divergence, which **releases** this guard so the agent can correct (#243/#245/#276). Drift unblocks the trade; agreement skips it. I saw it fire during the investigation — my first probe printed *"venue holds 99.01 but the last action asked for 96.98; releasing the duplicate-action guard so the agent can correct it"* on all four venues, which is also **how not to measure this**: a harness that reports a stale position makes every venue re-trade and the difference vanishes.

## The change

One implementation on `TorchTradeFuturesLiveEnv`, four overrides deleted, and the `futures_live_base.py` caveat — *"Inert for bybit and okx"* — retired, because `_sync_action_level_after_reset` is now load-bearing for all four.

## Two consequences worth reviewing closely

**1. The abstractness proxy broke.** `_execute_trade_if_needed` was the *only* `@abstractmethod` on `TorchTradeFuturesLiveEnv`, so implementing it left the five intermediate bases with nothing unimplemented and `inspect.isabstract` stopped excluding them — the discovered-env lists grew **10 → 15**, caught by their own `assert len(...) == 10`. Two files used abstractness as a proxy for "is a real env"; both now also key on the venue module (`env.py`/`env_sltp.py`), which is stable under a refactor in the way "which method happens to be abstract" is not.

Declaring `_execute_fractional_action` abstract instead is **not** the fix, and I tried it: the four SLTP envs share this base and trade brackets, so it made them abstract and shrank the list to **6**. Flagging the tradeoff explicitly — the intermediate bases are no longer rejected by `EnvBase.__new__`. If a reviewer would rather pay for a `PlainFuturesLiveEnv` intermediate class to keep that property, say so.

**2. The per-venue tests covered one venue each and asserted the call, not the outcome.** binance and bitget had `test_fractional_resizing_executes`; bybit had `test_fractional_resizing_always_delegates`, which *pinned the bug* — `(1.0, 1.0)` "Same level: delegates to fractional" — written in the same commit that removed the guard. okx had none at all. All three mocked `_execute_fractional_action` and asserted it was called.

Replaced by three behavioural tests over **all four venues** — hold on repeat, trade on reverse, trade on a same-direction resize — plus a structural guard that no plain futures env re-forks the gate, which is what makes testing it once sufficient. Removing the guard fails **10** of them.

The shared harness mirrors both submission paths, because the venues reverse differently: binance calls `close_position()` then `trade(sell, ...)`, while bybit sends one `trade(sell, ...)` spanning close-and-open. Counting only `trade` leaves binance's mirrored position at +1.9 — still long — and the guard then releases on the divergence, exactly as designed.

## Numbers

Production **−29**, tests **−53**, net **−82**. Suite **4701 passed, 16 skipped**.

Reviewed pre-implementation by `torchrl-engineer` (verdict: proceed, home it on `TorchTradeFuturesLiveEnv` not `TorchTradeLiveEnv` — alpaca has a different signature and is not broken).


BY571:
@CharlieHelps ready for review at `864ca7a7`. Three rounds of five agents; the fold itself survived unchanged, and every round found something in **my own additions** instead.

## The change

`_execute_trade_if_needed` looked like the last row of #288's duplication table — 4 copies, 53 lines, "bybit/okx byte-identical". It is not duplication. Those four copies implement **two different behaviours**, and pairwise similarity cannot see which half is right:

| | binance / bitget / alpaca / **offline** | bybit / okx |
|---|---|---|
| repeated action, price moved | **hold** | **rebalance** |

```
venue     |   100    101    103    106    110 | re-trades | extra notional
binance   | TRADE   .     .     .     .       |     0     |     $0.00
bybit     | TRADE TRADE TRADE TRADE TRADE     |     4     |   $946.77
```

The tiebreaker is not tidiness: `offline/sequential.py:697` holds, and that is the env the policy **trains in**. Live, the learned "hold" silently meant "rebalance" on two of four venues, at fees the training reward never charged.

Not a decision taken twice, either. `80bdd892` (2026-02) removed the guard from bybit to be *"robust to state drift from rejected orders, partial fills, or external intervention"*; okx was written from bybit two months later. That concern is now handled a layer up by machinery that did not exist then — `_sync_position_from_exchange` NaNs `current_action_level` on a divergence, which **releases** this guard. I watched it fire while investigating: my first probe printed *"venue holds 99.01 but the last action asked for 96.98; releasing the duplicate-action guard so the agent can correct it"* on all four venues, which is also **how not to measure this** — a harness reporting a stale position makes every venue re-trade and the difference vanishes.

## What the rounds found, all in what I added

1. **A test I wrote to prove the guard works could not see it missing.** The resize row drove three bars at a constant price, so its final "then hold" passed with the guard **entirely deleted** — at constant price the delta is zero and sizing no-ops regardless. Moving one price took the guard-deletion mutant from failing 8 rows to failing all 20.
2. **The row's comment described coverage its assertion did not provide.** It claimed "a BUY with switching=False", but `_drive` only checked that `trade.call_count` rose. Reclassify a same-direction trim as a direction switch and binance closes the whole ~$9,796 position and reopens instead of sending one ~$4,898 order — **3x the taker notional, a flat window mid-step, full P&L realisation instead of partial, every bar the agent resizes.** That mutant passed all 4709 tests. `_drive` now returns the order **shape** per step.
3. **The gate asserted shape and never size.** Corrupting the forwarded price doubles the quantity on every venue and passed all 4718 tests — it survives because the corrupted target stays self-consistent with what the venue reports, so the sync never diverges. Pre-existing (it survives on `main`'s four forwarders too); the fold makes it one forwarder instead of four. Now pinned.
4. **A dead assertion round 1 itself created**, and one that needed *three* independent guards broken at once to fail. Both gone.

## Two things I declined, with evidence

- **Deleting `grow-long`/`grow-short` as redundant.** They coincide with the trim rows on the delta-sign branch — where the reviewer looked — but `switching` is a second axis, and there they are opposites: a mutant firing `switching` on a *grow* fails only the grow rows; one firing on a *trim* fails only the trim rows. Four rows, four cells of a 2x2. A second reviewer found the same mutant independently.
- **Deleting `TestRegressionIssue189` as superseded.** Running the mutant shows it is the **sole** killer of "the guard reports `executed=True` on a skip" across all 4722 tests — nothing trades either way, so no count-based assertion can see it. The reviewer that proposed the deletion reproduced this and withdrew it.

## One tradeoff to rule on

`_execute_trade_if_needed` was the **only** `@abstractmethod` on `TorchTradeFuturesLiveEnv`, so implementing it left the five intermediate bases instantiable. `BinanceBaseTorchTradingEnv()` now constructs into a half-built env and fails later on `action_spec` with `IndexError`, where it used to fail fast with `TypeError`. Declaring `_execute_fractional_action` abstract instead does **not** work — the four SLTP envs share this base and trade brackets, so it made them abstract and dropped the discovered-env list to 6. Two reviewers called a `PlainFuturesLiveEnv` intermediate class scope creep for a deletion PR, and no production code instantiates those bases; the cheaper fix is hoisting the 4x-duplicated `action_spec` assignment, which I have noted on #288 for the next slice. Say if you would rather pay for it here.

Production **−46**, tests **−170**, net **−216**. Suite **4722 passed, 16 skipped**.

PR #427, opened 2026-09-02
refactor(288): the venue test fixtures were one shape with four venues' data inlined

BY571:
The #288 thread flagged ~2,400 lines of non-test defs across the four venue test directories as *"the biggest remaining win and nobody has been looking at it"*, worth about −600. This is that slice, measured rather than estimated.

## What is foldable, and what is not

22 fixture names recur across ≥2 venues (1,754 lines). But foldability is governed by the **worst** pair, not the best — the thread's own lesson — and the three largest fail that test:

| fixture | lines | min-pair | verdict |
|---|---|---|---|
| `env_with_mocks` | 291 | **20%** | genuine venue differences — left alone |
| `env` | 247 | **17%** | left alone |
| `mock_trader` | 157 | **11%** | left alone |
| `mock_observer` + `mock_env_observer` | 194 | — | **folded** |
| nested `status` ×10 | 60 | 94% | **folded** |
| nested observation builders ×9 | ~50 | 54–100% | **folded** |
| `mock_env_trader` (bybit≡okx) | 32 | 100% | **folded** |
| `custom_preprocess` (3 venues ≡) | 33 | 100% | **folded** |

## The design: the venue's data stays at the call site

Every `mock_observer` builds the same MagicMock. What differs is **data** — binance and bitget name their timeframes `"1m_10"` where bybit and okx say `"1Minute_10"`, some tests need a fixed OHLC bar rather than noise, some set `intervals`/`window_sizes`, some `get_features` and others `mirror_features_on`. So the shape is shared and the data is passed in:

```python
    @pytest.fixture
    def mock_observer(self):
        return a_mock_observer(
            ["1Minute_10", "5Minute_10"], timestamps=True, features=True,
        )
```

That is the whole point. Folding the keys *into* the helper is how a shared fixture stops testing four venues and starts testing one — the trap this issue's thread documents. **Verified rather than asserted: giving bybit binance's timeframe naming fails 2 tests.**

The 10 nested `status(qty)` helpers each rebuilt the same `PositionStatus` dict, and the per-venue `from <venue>.order_executor import PositionStatus` they carried was itself duplication — it is **one class** from `core.common_types`, re-exported four times (verified: `is` the same object for all four).

## Numbers

**Tests −204** (175 insertions, 379 deletions). **Production untouched.** Suite **4722 passed, 16 skipped**.

Also removed 18 imports whose only users were the deleted bodies — 16 introduced here, 2 pre-existing in the same files (`main` reports 2 for this file set; HEAD reported 18).

## Recorded, not fixed

okx's conftest observer hardcodes an OHLC bar that nothing asserts: replacing it with noise passes all 175 okx tests. Pre-existing — `main`'s copy has the same bar and the same absence of a test — so it is noted rather than folded into this diff.

### DIFF
--- a/tests/envs/base_exchange_tests.py
+++ b/tests/envs/base_exchange_tests.py
@@ -12,7 +12,102 @@ import torch
 from tensordict import TensorDict
 from unittest.mock import MagicMock, patch
 from abc import ABC, abstractmethod
+from torchtrade.envs.core.common_types import PositionStatus
 from torchtrade.envs.utils.timeframe import TimeFrame, TimeFrameUnit
+
+
+# --- the mocks every venue builds, with the venue's own data at the CALL SITE (#288) -----
+#
+# These were 8 copies of `mock_observer`, 10 of a nested `status`, and a scatter of
+# `mock_observations`/`mock_get_observations`/`mock_obs_zero`. What differed between them
+# was data, not shape: binance and bitget name their timeframes "1m_10" where bybit and okx
+# say "1Minute_10", and a couple of tests need a fixed OHLC bar rather than noise. Passing
+# that in keeps the venue axis these tests exist to cover -- folding it into the helper is
+# how a shared fixture stops testing four venues and starts testing one.
+
+_OBS_FEATURES = {
+    "observation_features": ["feature_close", "feature_open", "feature_high", "feature_low"],
+    "original_features": ["open", "high", "low", "close", "volume"],
+}
+
+
+def some_observations(keys, base=None, timestamps=False):
+    """One window per key. `base` is the OHLC bar `return_base_ohlc=True` should yield --
+    a 4-tuple for a fixed bar, or None for noise."""
+    def observations(return_base_ohlc=False):
+        obs = {k: np.random.randn(10, 4).astype(np.float32) for k in keys}
+        if return_base_ohlc:
+            obs["base_features"] = (
+                np.random.randn(10, 4).astype(np.float32) if base is None
+                else np.array([list(base)] * 10, dtype=np.float32)
+            )
+            if timestamps:
+                obs["base_timestamps"] = np.arange(10)
+        return obs
+    return observations
+
+
+def a_mock_observer(keys, *, base=None, timestamps=False, features=False,
+                    intervals=None, window_sizes=None):
+    """The observer mock, parametrised by what actually differs between the venues."""
+    observer = MagicMock()
+    observer.get_keys = MagicMock(return_value=list(keys))
+    observer.get_observations = MagicMock(
+        side_effect=some_observations(keys, base, timestamps))
+    if features:
+        observer.get_features = MagicMock(return_value=dict(_OBS_FEATURES))
+    else:
+        mirror_features_on(observer)
+    if intervals is not None:
+        observer.intervals = intervals
+    if window_sizes is not None:
+        observer.window_sizes = window_sizes
+    return observer
+
+
+def a_mock_futures_trader():
+    """The flat, healthy trader every futures env test starts from."""
+    trader = MagicMock()
+    trader.cancel_open_orders = MagicMock(return_value=True)
+    trader.close_position = MagicMock(return_value=True)
+    trader.get_account_balance = MagicMock(return_value={
+        "total_wallet_balance": 1000.0, "available_balance": 900.0,
+        "total_unrealized_profit": 0.0, "total_margin_balance": 1000.0,
+    })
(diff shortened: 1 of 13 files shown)


BY571:
@CharlieHelps ready for review at `3e54d1ff`. Three rounds of five agents. **Tests −615, production untouched, suite 4724 passed.**

## What this is

The #288 thread flagged ~2,400 lines of non-test defs across the four venue test directories as *"the biggest remaining win and nobody has been looking at it"*. This is that slice.

## What I folded, and what I refused to

22 fixture names recur across ≥2 venues (1,754 lines) — but foldability is governed by the **worst** pair, and the three largest fail that test:

| fixture | lines | min-pair | verdict |
|---|---|---|---|
| `env_with_mocks` | 291 | **20%** | left alone |
| `env` | 247 | **17%** | left alone |
| `mock_trader` | 157 | **11%** | left alone |

Also left inline: ~19 `PositionStatus(...)` sites carrying deliberately hostile values (`entry_price=47500`, `leverage=20`, `unrealized_pnl=26.3`). Folding those needs three more knobs on a factory round 1 correctly shrank from six to four.

**The design point:** the venue's data stays at the call site.

```python
return a_mock_observer(["1Minute_10", "5Minute_10"])
```

binance and bitget say `"1m_10"`; bybit and okx say `"1Minute_10"`. Folding that in would have been an easy extra −20 lines and is exactly how a shared fixture stops testing four venues and starts testing one. Verified, not asserted: **giving bybit binance's naming fails 2 tests**, re-checked at this head.

## The evidence that matters for a harness refactor

A green suite proves nothing here. The reviewers ran **36 production mutations against `main` and against the branch and compared the exact failing test node IDs**: 35 byte-identical, and the 36th caught by **4 more** tests on the branch — because a mock that emits `base_features` only when asked kills a mutant that a permissive one let through. The fold made the suite stronger, not merely equal.

## The best thing in the PR is a bug the fold made fixable

Every mocked OHLC window was ten identical rows, so reading `base_features[0]` instead of `[-1]` for bracket pricing **passed the entire suite**. "Price off the most recent candle" was unpinned on every venue. On `main` that was twelve independently-wrong fixtures; after the fold it is one line, and a ramp fixes it everywhere at once — last row still exactly `base` (the bracket-price assertions read it), every earlier row faded 0.1% per bar. That mutation now **fails 9 tests**.

Round 3 then added the guard on that guard: widen the window or shrink the factor until float32 rounds the rows together and the blind spot returns, invisible to every existing assertion because they all read the last row.

## Three things I got wrong, corrected in the branch

1. **My round-1 fold moved the duplication into the helper rather than removing it** — I folded the class-level fixtures and then imported the two *ingredients* into files that still hand-built observers longhand. Both reviewers said so and both were right.
2. **The fold half-applied at six sites**, producing code that builds the factory's product and immediately overwrites it. Behaviourally identical, so the suite stayed green throughout — a partial refactor is invisible to tests by construction.
3. **My round-2 commit message had the `get_lot_size` story backwards.** I said two folded traders had been running on `float(MagicMock()) == 1.0`; every trader my predicate could fold already had that attribute, *because it is in the predicate*. The two genuinely on the accident were the ones it couldn't reach, and they are folded only now.

## Recorded on #288 rather than fixed here

The mutation sweep surfaced three refusal guards on real money paths where **deleting the guard fails zero tests** — `bybit/env.py:85`, `okx/env.py:81`, `okx/env_sltp.py:87`. Their siblings are well covered (`target_tol` mutations fail 9 tests each), so the omission looks accidental. Same shape as the #414 gaps. Also recorded: `a_position_status`'s five defaults are unobservable — but verified apples-to-apples, the eight inlined blocks this PR replaced carry byte-identical literals and mutating *those* on `main` also fails zero. The fold concentrated the deadness into one visible place rather than creating it.

Production diff is empty: `git diff main -- torchtrade/` returns nothing.

BY571:
Merged as `2d317cc7`. Two phrases in the squash commit message were eaten by shell backtick substitution when I composed it — correcting the record here, since the commit itself cannot be rewritten on `main`:

- Line 37 reads *"and 's reset, which nothing in 4722 tests kills"*. It should read: **"and `consecutive_unknown_status`'s reset, which nothing in 4722 tests kills"** — the outage counter's reset-to-zero on a healthy bar (`core/live.py:401`), recorded on #288.
- Line 39 reads *"Production untouched --  is empty."* It should read: **"Production untouched -- `git diff main -- torchtrade/` is empty."**

Nothing else was affected; the diff and the review are unchanged.

PR #428, opened 2026-09-02
refactor(288): the plain action space was built four times; #425 is what that costs

BY571:
The last production duplication in #288 — and I want to be upfront that **this diff is production +8, not negative**.

## What it folds

The tail of all four plain futures `__init__`s, byte-identical:

```python
self.reward_function = reward_function or log_return_reward
self.action_levels = config.action_levels
self.action_spec = Categorical(len(self.action_levels))
```

It cannot go on `TorchTradeFuturesLiveEnv`: the four SLTP envs inherit that same base, build their spec from `action_map`, and **their config has no `action_levels` field at all**. It cannot be a shared `__init__` either — the venues pass different credentials to `super()`, bitget and okx one more than the others.

So it is a **mixin**, which is what this repo already does for exactly this asymmetry:

```
plain: <venue>FuturesTorchTradingEnv     -> PlainFuturesLiveEnv -> <venue>Base -> ...
sltp : <venue>FuturesSLTPTorchTradingEnv -> SLTPMixin           -> <venue>Base -> ...
```

## Why, given it adds lines

Four copies × five lines is 20 removed; any new module costs ~15 in imports and class scaffolding. The line count cannot win here, so it is not the argument.

The argument is that **#425 was this exact drift**: these two lines in four copies, where binance derived its levels from a helper while three venues hard-coded a five-level list — so one unset config produced a **3-action space on one venue and 5 on the others**, and `action_spec.n` is what a checkpoint binds to. One copy cannot disagree with itself.

## Two guards, because the obvious one is not enough

`_SHARED_METHOD_OWNERSHIP` catches a venue that **re-forks** `_init_plain_trading` — verified, a pure `super()`-delegating override on bybit fails it.

It cannot catch a venue that stops **calling** it: an inlined copy still *resolves* the method, so the identity check passes. Same shape as the "sync call present in source but dead at runtime" hole from #426. I measured it — inlining a five-level list back into bybit failed exactly **one** test, and it was bybit's own `test_action_spec`. Which is precisely how #425 stayed invisible: the drift was only ever visible to whichever venue happened to assert on it.

`test_no_plain_venue_builds_its_own_action_space` reads the source instead. Both mutations now fail, on the right venue:

| mutation | caught by |
|---|---|
| inline the logic, never call the mixin | `test_no_plain_venue_builds_its_own_action_space[Bybit…]` |
| re-fork the method, even as a pure `super()` delegate | `test_no_futures_env_reforks_a_shared_step_or_reset[Bybit…-_init_plain_trading]` |

## Also

The function-local `log_return_reward` import guarded no cycle (`default_rewards` imports only numpy and torch) and is at module scope now; `Categorical` became dead in all four leaves.

**Docstrings deliberately not touched.** binance and bitget spend 11–12 lines documenting venue-specific parameters (`api_passphrase: Bitget API passphrase (required!)`). That is API documentation, not duplication — deleting it to make this diff look line-negative is not what the issue asks for.

## Numbers

Production **+8** (four venue files −20, new mixin +28), tests **+27**. Suite **4732 passed, 16 skipped**.

If reviewers judge that +8 production lines and a new MRO level are not worth foreclosing one bug class, saying so is a legitimate outcome and I will revert it — the measurement is in the open either way.


BY571:
@CharlieHelps ready for review at `a3718d52`. Three rounds of five agents — and round 1 told me to **revert**, which was correct. The PR you are looking at is not the PR I opened.

## What it is now

#288's last production duplication. Two shared initialisers, ten call sites folded:

- `TorchTradeLiveEnv._init_action_space(reward_function)` — the plain tail (`reward_function`, `action_levels`, `action_spec`), called by all **five** plain envs including alpaca.
- `SLTPMixin._init_bracket_action_space(reward_function, *, include_short_positions)` — the twelve-line bracket tail, including `create_sltp_action_map(...)`, called by all **five** SLTP envs.

No new file, no new MRO level. **Production −84, tests +102.**

## What I had wrong, and what fixed it

I opened this with a new `PlainFuturesLiveEnv` mixin, production **+8**, justified in its own docstring by "#425 was these two lines in four copies". A reviewer checked the pre-#425 tree: **that is false.** The drift lived in each venue's `Config.__post_init__`, and `b612fe8c` had already folded all four configs into one. The lines I was extracting were byte-identical before the drift, during it, and after — mechanical forwarding off a source of truth that was already singular. I had built an abstraction against a recurrence that could not occur, and written the false premise into the source as its rationale.

Three more, each verified rather than argued:

- `TorchTradeFuturesLiveEnv` **already** carries `_step`/`_reset` that the SLTP envs inherit and never run — so my "it would lie to the SLTP MRO" objection was contradicted by the status quo.
- A **fifth** copy sat in `alpaca/env.py`. My guard was futures-scoped, so it asserted "built once" while five call sites existed.
- My new guard contradicted an existing comment at `test_live_env_base.py:4798`.

Following the revert produced the version worth shipping — and the useful fold was in a throwaway line at the end of that same review: *the SLTP side was never folded*. Four envs each inlined twelve lines, `SLTPMixin` was already in every one of their MROs, so it cost nothing structural.

## The guards, and why five rather than one

| guard | the only thing that catches it |
|---|---|
| ownership table | a leaf **re-forking** the initialiser under the same name |
| AST call-site check | a leaf **inlining** a faithful copy and never calling it — verified: that passed all 4749 tests |
| `action_spec.n` derivation | `Categorical(3)` hard-coded — survived on `main` **and** the branch |
| reward-function identity (plain + SLTP) | the caller's function silently dropped — **0 of 4750** tests noticed |
| hold-action index | `include_hold_action` had no live-env test anywhere |

The last three were mutations that survived the whole suite before this PR. The first two are complementary and neither subsumes the other: identity cannot see a leaf that stops *calling*; AST cannot see one that *overrides*.

The hold-action one is the one I would look at hardest. With the flag off, HOLD leaves index 0 and **every index shifts down by one** — a policy trained without it reads live action `0` as "open a long bracket" rather than "do nothing". That is #418's shape at the single line this fold created.

## Two judgement calls worth your eye

1. **`include_short_positions` is a required keyword**, not read from the config. alpaca is spot: it passes `False` and its config has no such field. My first version used a `None` sentinel that had the shared method reach for a field alpaca lacks — which hides the one venue difference inside the helper. #427's whole lesson was that venue data belongs at the call site, and I violated it one PR later.
2. **alpaca is not in the three behavioural guards**, because `_real_futures_env` is scoped to the four futures venues. It gets structural coverage from the AST and ownership guards; the round-3 reviewer verified its rollout byte-identical to `main` by hand. Widening the harness is a follow-up, not this PR — say if you disagree.

Reviewers verified rollout equivalence by execution for all ten envs against `main` (byte-for-byte, including every entry of `vars(env)`), `check_env_specs` on all ten, and paired mutation sweeps finding no regression. Suite **4727 passed**.

One round-3 agent (a long paired mutation sweep) is still running; I will relay anything it surfaces.

BY571:
Correction to my review request, @CharlieHelps — the last round-3 agent landed and one of the two judgement calls I put to you was not a judgement call. It was a hole in my own test, and it is fixed at `fd668b9b`.

I wrote: *"alpaca is not in the three behavioural guards... widening the harness is a follow-up."* Measured: dropping `reward_function` at either of alpaca's call sites — `self._init_action_space()`, `self._init_bracket_action_space(include_short_positions=False)` — failed **0 of 3710 tests**, on `main` and on this branch.

`test_a_live_env_keeps_the_reward_function_it_was_handed` exists precisely to catch a silently discarded reward function. It closes that for four of five venues, because `PLAIN_VENUES` derives from `PLAIN_FUTURES_ENVS` and excludes spot. The AST guard forces alpaca to *call* the shared initialiser but says nothing about the arguments it passes — so the venue whose call sites differ most from the others was the one with no argument coverage. `test_alpaca_keeps_the_reward_function_it_was_handed` now covers both paths; each mutant fails on its own path.

Also corrected: the mixin docstring said *"four copies of `create_sltp_action_map(...)`"* four lines below *"the tail all five SLTP `__init__`s repeat"* — self-contradictory, and undercounting by exactly the venue this round folded. That is the **third** arity claim in that one docstring the fold falsified; a docstring enumerating a contract goes stale exactly when the contract moves, which is what a fold does by definition.

So one judgement call remains for you, not two: **`include_short_positions` as a required keyword** rather than a config read with an alpaca override.

Round 3's sweep, for the record: 25 paired mutations across both trees, **zero regressions**; a 102-configuration byte-identity snapshot of all ten envs against `main` — identical, alpaca's newly-folded SLTP env included; and mutations I had not asked for, all newly caught here (hold/close kwargs swapped 13→17, the short flag routed into the hold slot 4→16, single-call-site reward drops 0→1).

Head is `fd668b9b`. Production **−84**, tests **+131**, suite **4729 passed**.

PR #429, opened 2026-09-03
docs: three capabilities the docs promise that the code does not implement

BY571:
Three things the docs promise that the code does not do. Each verified against the source before changing anything.

## 1. Alpaca equities do not exist

`AlpacaObservationClass` builds a `CryptoBarsRequest` against a `CryptoHistoricalDataClient`. There is no `StockHistoricalDataClient` in the package. The README advertised `symbol="AAPL"`, `symbol="SPY"`, "Stocks: AAPL, GOOGL, MSFT, TSLA", "ETFs: SPY, QQQ, IWM".

The order side uses `TradingClient`, which accepts an equity order. So the env constructs, then returns no data. The failure looks like an outage rather than a config error.

The observer's own docstring said "Spot crypto bars from alpaca" the whole time. Only the README was wrong.

## 2. Rate limiting is not implemented

`grep -rn "ratelimit\|rate_limit\|throttl"` over `torchtrade/envs/live/` and `core/live.py` returns nothing. Two files promised it was automatic. They now say it is not, and when that starts to matter.

## 3. "Invalid orders, no crash" is backwards

The executors raise. #414 and #420 made them stricter on purpose: an order the env thinks it placed and the venue rejected is worse than a loud failure.

Also removed: `extended_hours=True` (in no source file), "WebSocket streaming" (the observer is REST), equity market hours on a 24/7 crypto venue, and `ALPACA_KEY`/`ALPACA_SECRET` where every working script uses `ALPACA_API_KEY`/`ALPACA_SECRET_KEY`.

## Two claims that cause the re-forking #288 removed

| file | said | actual |
|---|---|---|
| `live/README.md` | "there is no shared base to inherit from" | since #425 the futures venues inherit `BaseFuturesTradingConfig` |
| `utils/README.md` | "SLTPMixin is small: four members, does not own bracket pricing or the exit" | 13 members, owns `_step`, `_bracket_entry_price`, `_close_action`, `_mark_flat` |

The first is the costly one. Following it, a new venue re-declares shared fields, shadows the base defaults and stops tracking them. That is how the action space came to differ 5 vs 3 between venues.

## The #425 checkpoint break

`action_spec.n` went 5 to 3 on bitget, bybit and okx, so a policy head trained before the fold will not load. There was no note anywhere. The repo already had "Migrating from before #289"; #425 now has its sibling. The escape hatch is verified: passing the old five level list still constructs on all four venues.

## Notes

Docs only. Suite 4729 passed, including the 698 checks in `test_documented_imports.py`.

Those 698 passed while every claim above was wrong, and that is their nature rather than a gap. They check that symbols resolve and blocks parse. `symbol="AAPL"` is a valid kwarg with a valid value, and "automatic throttling" is prose. Capability claims need reading the docs against the code.

Second commit responds to review feedback on the first: I had written three passages explaining what the docs used to say and defending the change, which belongs here and not in a README. Sections went 6 lines to 2, 18 to 8, and so on, losing no facts. Added lines 79 to 56.

Follow-ups from the same audit: broken code examples in `docs/guides/custom-environment.md` (its spec assignment, step loop and base class name are all wrong), and structure docs describing directories that no longer exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,12 +190,12 @@ paper-only** — see its entry.
 - **⚠️ Paper trading only — not live:** `dry_run=False` raises. `py-clob-client` is archived/non-functional (Polymarket moved to CLOB V2), and the env holds bets to resolution while Polymarket only releases collateral on an on-chain redeem no client exposes — so a live bot would drain to zero *while winning*. Reviving live needs the V2 port **and** a redemption workflow.
 - **Get Started:** [Browse markets at Polymarket](https://polymarket.com/)
 
-### 📈 Stock & Crypto API
+### 📈 Crypto Spot API
 
 **[Alpaca](https://alpaca.markets/)** - Commission-free trading API
 - **Supported by:** `AlpacaTorchTradingEnv`, `AlpacaSLTPTorchTradingEnv`
 - **Features:** Commission-free crypto spot, paper trading, real-time data
-- **Best for:** US markets, algorithmic trading
+- **Best for:** crypto spot, algorithmic trading
 - **Get Started:** [Sign up for Alpaca](https://alpaca.markets/signup)
 
 ---
@@ -364,8 +364,11 @@ env = SequentialTradingEnv(df, config)
 ```python
 import os
 
+from dotenv import load_dotenv
 from torchtrade.envs.live.alpaca import AlpacaTorchTradingEnv, AlpacaTradingEnvConfig
 from alpaca.data.timeframe import TimeFrame, TimeFrameUnit
+
+load_dotenv()  # reads the .env written above; os.getenv does not read files
 
 config = AlpacaTradingEnvConfig(
     symbol="BTC/USD",
--- a/README.md
+++ b/README.md
@@ -311,8 +311,8 @@ source .venv/bin/activate  # Unix/macOS
 
 # 5. For live trading, create .env file
 cat > .env << EOF
-API_KEY=your_alpaca_api_key
-SECRET_KEY=your_alpaca_secret_key
+ALPACA_API_KEY=your_alpaca_api_key
+ALPACA_SECRET_KEY=your_alpaca_secret_key
 BINANCE_API_KEY=your_binance_api_key
 BINANCE_SECRET_KEY=your_binance_secret_key
 BYBIT_API_KEY=your_bybit_api_key
--- a/README.md
+++ b/README.md
@@ -133,8 +133,8 @@ TorchTrade supports live trading with major exchanges:
 
 | Environment | Exchange | Asset Type | Futures | Leverage | Bracket Orders |
 |-------------|----------|------------|---------|----------|----------------|
-| **AlpacaTorchTradingEnv** | Alpaca | Crypto/Stocks | ❌ | ❌ | ❌ |
-| **AlpacaSLTPTorchTradingEnv** | Alpaca | Crypto/Stocks | ❌ | ❌ | ✅ |
+| **AlpacaTorchTradingEnv** | Alpaca | Crypto | ❌ | ❌ | ❌ |
+| **AlpacaSLTPTorchTradingEnv** | Alpaca | Crypto | ❌ | ❌ | ✅ |
 | **BinanceFuturesTorchTradingEnv** | Binance | Crypto | ✅ | ✅ (1-125x) | ❌ |
 | **BinanceFuturesSLTPTorchTradingEnv** | Binance | Crypto | ✅ | ✅ (1-125x) | ✅ |
 | **BitgetFuturesTorchTradingEnv** | Bitget | Crypto | ✅ | ✅ (1-125x) | ❌ |
@@ -194,7 +194,7 @@ paper-only** — see its entry.
 
 **[Alpaca](https://alpaca.markets/)** - Commission-free trading API
 - **Supported by:** `AlpacaTorchTradingEnv`, `AlpacaSLTPTorchTradingEnv`
-- **Features:** Commission-free stocks & crypto, paper trading, real-time data
+- **Features:** Commission-free crypto spot, paper trading, real-time data
 - **Best for:** US markets, algorithmic trading
 - **Get Started:** [Sign up for Alpaca](https://alpaca.markets/signup)
 
@@ -362,6 +362,8 @@ env = SequentialTradingEnv(df, config)
 ### Live Trading with Alpaca
 
 ```python
+import os
+
 from torchtrade.envs.live.alpaca import AlpacaTorchTradingEnv, AlpacaTradingEnvConfig
 from alpaca.data.timeframe import TimeFrame, TimeFrameUnit
 
@@ -376,8 +378,12 @@ config = AlpacaTradingEnvConfig(
     paper=True  # Start with paper trading!
 )
 
-env = AlpacaTorchTradingEnv(config)
-# See examples/live/alpaca/collect_live.py
+env = AlpacaTorchTradingEnv(
+    config,
+    api_key=os.getenv("ALPACA_API_KEY"),
+    api_secret=os.getenv("ALPACA_SECRET_KEY"),
+)
+# See examples/online_rl/ppo/live.py
 ```
 
 ### LLM-Based Trading
--- a/docs/environments/offline-rl.md
+++ b/docs/environments/offline-rl.md
@@ -74,10 +74,19 @@ for batch in collector:
 ### From Real Online Environments
 
 ```python
+import os
+
+from dotenv import load_dotenv
 from torchtrade.envs.live.alpaca import AlpacaTorchTradingEnv, AlpacaTradingEnvConfig
 
+load_dotenv()
+
 # Collect real trading data (paper trading recommended)
-env = AlpacaTorchTradingEnv(config)
+env = AlpacaTorchTradingEnv(
+    config,
+    api_key=os.getenv("ALPACA_API_KEY"),
+    api_secret=os.getenv("ALPACA_SECRET_KEY"),
+)
 
 # Record interactions
 for episode in range(num_episodes):
--- a/docs/environments/offline.md
+++ b/docs/environments/offline.md
@@ -255,21 +255,21 @@ config = SequentialTradingEnvSLTPConfig(
     leverage=5,
     stoploss_levels=[-0.02, -0.05],
     takeprofit_levels=[0.03, 0.06],
-    ...
+    # ... the rest of the config
 )
 
 # Notional: always trade $500 USD worth
 config = SequentialTradingEnvSLTPConfig(
     trade_mode="notional",
     quantity_per_trade=500.0,    # $500 per trade
-    ...
+    # ... the rest of the config
 )
 
 # Quantity: always trade exactly 0.05 BTC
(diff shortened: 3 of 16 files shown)


BY571:
@CharlieHelps ready for review at `feb13c1b`. Docs only, no code.

Three capability claims that were false, each checked against the source first:

1. **Alpaca equities.** `observation.py` builds a `CryptoBarsRequest` against a `CryptoHistoricalDataClient`; there is no `StockHistoricalDataClient` in the package. The README advertised `symbol="AAPL"`, `symbol="SPY"`, "Stocks: AAPL, GOOGL, MSFT, TSLA". The order side accepts an equity order, so the env constructs and then returns no data, which looks like an outage rather than a config error.
2. **Rate limiting.** `grep -rn "ratelimit|rate_limit|throttl"` over the live stack returns nothing. Two files said it was automatic.
3. **"Invalid orders, no crash"** is backwards. The executors raise, and #414 and #420 made them stricter on purpose.

Two more that would cause the re-forking #288 removed: `live/README.md` said there is no shared config base (false since #425), and `utils/README.md` described `SLTPMixin` as four members that do not own bracket pricing or the exit. It has 13 and owns both.

And the #425 checkpoint break, which was documented nowhere. `action_spec.n` went 5 to 3 on bitget, bybit and okx. There is a migration note now, next to the existing #289 one, and I verified the escape hatch: passing the old five level list still constructs on all four venues.

Two things worth your judgement:

- **I did not run the five agent, three round cycle.** `test-justifier`, `pr-test-analyzer` and `torchrl-engineer` have nothing to say about a zero code diff, and the review that mattered was reading each claim against the source. Tell me if you would rather I run it anyway on docs PRs.
- **The 698 checks in `test_documented_imports.py` passed the whole time these claims were wrong.** That is their nature, not a gap: they verify symbols resolve and blocks parse, and `symbol="AAPL"` is a valid kwarg with a valid value. If you can see a cheap way to test a capability claim, I would like to hear it, because right now nothing catches this class automatically.

Suite 4729 passed. Second commit responds to review feedback on the first: I had written passages defending the change into the READMEs themselves, which belongs in the PR. Added lines went 79 to 56 with no fact lost.

BY571:
@CharlieHelps all seven applied in 0f4495c7. Each one was verified by execution first; all seven were correct.

1. `live/README.md`: tree comment, Markets, Features, and both config examples (`AAPL`, `SPY` -> `BTC/USD`). `grep -nE "AAPL|SPY|US equit|extended hours|fractional shares"` is now empty for that file.
2. Split the two paths. Pre-submission argument checks raise (missing limit price, sub-minimum quantity, non-finite LOT_SIZE step); a venue rejection is `except Exception -> logger.error -> return False` and the env treats the trade as not executed.
3. Scoped to the four futures venues. The ownership table, printed from `__qualname__`:

   ```
   method                        Alpaca   Binance  Bitget   Bybit    OKX
   _step                         own      MIXIN    MIXIN    MIXIN    MIXIN
   _execute_trade_if_needed      own      MIXIN    MIXIN    MIXIN    MIXIN
   _reset                        MIXIN    MIXIN    MIXIN    MIXIN    MIXIN
   _close_action                 MIXIN    MIXIN    MIXIN    MIXIN    MIXIN
   ```

4. Now "one date-range fetch per configured timeframe, each call" (`base_obs.py:71` loops `self.time_frames`).
5. The note now says it breaks checkpoints trained on the old *unset* default, and that a config which passed `action_levels` explicitly is unaffected.
6. Corrected. `order_executor.py` defaults to `ioc`; `env_sltp.py:253` passes `gtc`. Its docstring also claimed `default: "gtc"`, so I fixed the docstring and removed a stray `# gtc` comment beside the signature.
7. Replaced the figure with the provider's rate-limit page.

Chasing your third point past the README turned up four more, which is the part worth flagging:

- **`sltp_mixin.py`'s own class docstring carried the identical false claim** ("All five SLTP envs inherit them" about `_step`). I had corrected the README and left the source saying the same wrong thing.
- **"`_bracket_entry_price` differs by venue" reads as a leaf override.** It is one shared method on the mixin branching on `_PRICES_OFF_THREADED_MARK` (bybit/okx off the mark, binance/bitget off the candle close). okx is the only leaf that really overrides anything here (`_resolve_bracket_quantity`).
- **`getting-started.md`, root `README.md` and `online.md` told users to put `API_KEY` / `SECRET_KEY` in `.env`.** Every example reads `ALPACA_API_KEY` / `ALPACA_SECRET_KEY`, and every other venue already used the prefix. This one fails silently: `os.getenv` returns `None`, `TradingClient(None, secret_key=None)` constructs fine, and you find out when an order is rejected.
- **The equity claim also lived in `envs/README.md`, `getting-started.md` and `online.md`.**

I also cut "so the bracket survives the bar" from the time-in-force line. Nothing in the code states why `gtc` is passed there, so that was a rationale I invented rather than one I found.

4770 passed, 16 skipped; 698 documented-import checks pass. Ready for another look.

BY571:
@CharlieHelps all four applied in `90469d1d`. You were right on every one; here is what I measured.

**1. The rejection contract.** My "pre-submission checks raise" was wrong, and not just for Alpaca. I parsed each `trade()` for where the top-level `try` opens relative to its `raise ValueError` statements:

```
alpaca    try at 186, raises at 244/251/259   inside  -> returns False
binance   try at 319, raises at 344/350       inside  -> returns False
bitget    try at 405, no ValueError
bybit     try at 240, raise at 238            before  -> escapes
okx       try at 312, raise at 310            before  -> escapes
```

So the identical malformed order raises on bybit and okx and returns `False` on alpaca and binance. Documented as venue-specific, with the explicit advice not to rely on an exception to catch a bad order.

**2. Equities on the primary surfaces.** Confirmed and fixed: the capability table and provider list in the root `README.md`, plus the capability and comparison tables in `docs/environments/online.md`. My earlier pass corrected the prose in `online.md` and left the tables directly above and below it. That is the third time in this PR I fixed an instance and left the class.

**3. The Alpaca guide.** The market-hours and holiday items sat two lines below "Crypto trades 24/7"; removed. On the links, `curl` says:

```
docs.alpaca.markets/docs/rate-limit      301 -> /us/docs/rate-limit -> 404
alpaca.markets/docs/market-hours/        404
alpaca.markets/docs/trading/             301 -> elsewhere
paper-api.alpaca.markets/                401   (API endpoint, not the dashboard)
```

The rate-limit URL was one I added in the previous commit, so I shipped a dead link while fixing dead claims. I could not find a rate-limit page that resolves, so it now points at `docs.alpaca.markets` rather than a deeper URL that rots. The paper link now goes to the actual dashboard.

**4. The root example.** Correct: it documented `ALPACA_API_KEY` / `ALPACA_SECRET_KEY` and then called `AlpacaTorchTradingEnv(config)`, taking the constructor's empty-string defaults. It now passes both and imports `os` to do it. It also cited `examples/live/alpaca/collect_live.py`; there is no `examples/live/` directory at all. Repointed at `examples/online_rl/ppo/live.py`.

On your coverage advice: I took the Markdown link checker and left the capability-matrix idea alone. In a separate PR (structure docs) I added a guard asserting every repo path a doc names is in the git index, across three forms: backticked, github tree/blob URL, and relative link target. It found nine phantoms, including eight broken relative links, four of them pointing at a `torchtrade/envs/offline/README.md` that has never existed. I did not extend the import sweep to judge prose, per your note.

BY571:
@CharlieHelps both fixed in `bb754d63`.

**The header.** You are right that this was the part doing the damage. The section read `### 📈 Stock & Crypto API` with **Best for: US markets** underneath, and the corrected "Commission-free crypto spot" sat on the line between them. A reader skims the heading and the tagline, so the one accurate line was sandwiched between the two inaccurate ones. Now `### 📈 Crypto Spot API` and **Best for: crypto spot**.

**The `.env`.** Confirmed and worse than cosmetic: nothing bridged the file to the process. `os.getenv` does not read `.env`, so both keys resolve to `None` and the constructor takes its empty-string defaults. The failure surfaces as a rejected order, not as an error at construction.

Both live examples now call `load_dotenv()` first, matching what every live example under `examples/` already does (`examples/rule_based/live.py:30`, `examples/online_rl/ppo/live.py:43`) and what `python-dotenv`, already a declared dependency, is there for. `docs/getting-started.md` had that gap plus the one you caught in the root README last round: it constructed `AlpacaTorchTradingEnv(config)` with no credentials at all. It passes both now.

Verified `load_dotenv()` + `os.getenv` resolves keys from a real `.env`, and that all 15 python blocks across the two files still parse.

Worth noting what this sequence looked like from my side: across three passes you found the same equity claim in the prose, then the tables, then the header. Each time I fixed the instance I was shown. The grep that would have closed it in one go is the one I ran only after your second pass.

BY571:
@CharlieHelps fixed in `a1adc999`. I swept the whole tracked doc set for the shape this time instead of fixing the five you listed, and it found four more:

```
docs/environments/offline-rl.md          Alpaca
docs/environments/online.md              Alpaca, AlpacaSLTP, Binance, BinanceSLTP
docs/getting-started.md                  Binance
docs/guides/custom-features.md           Binance      <- not in your list
torchtrade/envs/live/binance/README.md   Binance      <- not in your list
```

All nine call `load_dotenv()` and pass `api_key`/`api_secret`. The two feature-preprocessing examples take them in their own file's style, `os.environ[...]`, since those blocks are about `feature_preprocessing_fn` and the same files already show credentials that way.

Two the sweep flagged that I checked and deliberately left alone:

- The replay example in `online.md` injects `observer=` and `trader=`. `_init_trading_clients` gives a supplied instance precedence, so credentials are never read there. Correct as written.
- The Polymarket examples take `private_key`, not `api_key`/`api_secret`, and the env is paper-only. Different problem.

I also took the two pre-existing fence failures you called non-blocking, plus eleven more of the same kind. The cause is `...` as an elision marker inside a config call: `f(kw=1, ...)` is a `SyntaxError` because `Ellipsis` is a positional argument after a keyword one, so every one of those blocks fails on copy. They are comments now. 182 python fences repo-wide parse, up from 178.

One process note, since it is relevant to trusting the verification: my first attempt at this commit generated the constructor edits with a regex, and it produced blank lines between every argument and a `)` sharing a line with the closing fence. The blocks still parsed, so my syntax check passed while the markdown was actually broken. I reverted the files and redid it with exact anchors per file, then checked that every fence is balanced rather than only that the Python parses.

And you do not need to apologise for the review sequence. Each pass was correct, and the incremental surfacing was mostly a consequence of my fixing exactly what I was shown. The sweep was mine to run.

BY571:
@CharlieHelps thanks, and you are right about the wording. My commit message and my comment both opened with "All nine now call `load_dotenv()`" and then corrected themselves one sentence later. The accurate split is:

- **seven** call `load_dotenv()` and read via `os.getenv` (the entry-point examples in `offline-rl.md`, `online.md` ×4, `getting-started.md`)
- **two** use `os.environ[...]` with no `load_dotenv` (the feature-preprocessing fragments in `custom-features.md` and `binance/README.md`), because those are mid-application fragments and their own files already show credentials that way

No code change, as you say. I am noting it here rather than silently, because "all nine call load_dotenv" is exactly the kind of claim this PR exists to remove from the docs, and I put it in the commit message for the fix.

PR #430, opened 2026-09-03
docs: the guide for writing an env had three examples that raise

BY571:
`docs/guides/custom-environment.md` is the designated guide for building an environment. Its spec assignment, its step loop and its base class name were all wrong. Every fix was run before it was written.

## Spec assignment

`self._observation_spec = Composite(...)` raises:

```
AttributeError: To set an environment spec, please use
`env.observation_spec = obs_spec` (without the leading underscore)
```

Three occurrences, plus a fourth in example 2 that mutated the spec in place (`self._observation_spec["sentiment"] = ...`). The spec is a property, so it is assigned whole. I ran the rebuild pattern end to end: reset and step both carry the extra key, and `observation_spec.is_in` accepts the result.

## Step loop

`obs = env.step(action)` passes a raw tensor and treats the return as the next observation. Four occurrences here, one more in `metrics.md`. `step` takes and returns a TensorDict, and results arrive under `td["next"]`.

```python
td = env.reset()
td["action"] = env.action_spec.rand()
td = env.step(td)
reward = td["next", "reward"]
if td["next", "done"]: break
td = td["next"]
```

## Sampler attributes

Example 2 read `self.sampler.reset_index` and `.current_index`. Neither exists. The env tracks position with `self.current_timestamp`, which is what it passes to `sampler.get_observation`.

## Class name

The architecture diagram said `BaseTorchTradeEnv`. The class is `TorchTradeBaseEnv`.

## Notes

All 9 python blocks in the guide and 4 in `metrics.md` parse. Suite 4729 passed, including the 698 checks in `test_documented_imports.py`, which passed before this too: they verify imports resolve and blocks parse, and every broken line here was syntactically valid.

While verifying, the sampler rejected my synthetic bars with `Malformed OHLC in 105 of 300 rows`. That validation is working as intended, and it is the reason the examples now use bars built around open and close rather than drawn off close alone.

Second of three docs PRs from the audit. First is #429. Last is the structure docs describing directories that no longer exist.

### DIFF
--- a/docs/guides/custom-environment.md
+++ b/docs/guides/custom-environment.md
@@ -142,6 +142,8 @@ Extend `SequentialTradingEnv` to add custom features:
 ```python
 from torchtrade.envs.offline import SequentialTradingEnv, SequentialTradingEnvConfig
 from tensordict import TensorDict
+import numpy as np
+import pandas as pd
 import torch
 
 class CustomLongOnlyEnv(SequentialTradingEnv):
@@ -149,7 +151,7 @@ class CustomLongOnlyEnv(SequentialTradingEnv):
     Extended SequentialTradingEnv with sentiment data.
     """
 
-    def __init__(self, df, config: SequentialTradingEnvConfig, sentiment_data: torch.Tensor):
+    def __init__(self, df, config: SequentialTradingEnvConfig, sentiment_data: pd.Series):
         super().__init__(df, config)
         self.sentiment_data = sentiment_data  # Timeseries sentiment scores
 
@@ -166,7 +168,7 @@ class CustomLongOnlyEnv(SequentialTradingEnv):
 
         # `current_timestamp` is where the env is in the data. The sampler has no
         # index attribute.
-        obs["sentiment"] = torch.tensor([self.sentiment_data[self.current_timestamp]])
+        obs["sentiment"] = torch.tensor([self.sentiment_data.loc[self.current_timestamp]])
 
         return obs
 
@@ -174,7 +176,7 @@ class CustomLongOnlyEnv(SequentialTradingEnv):
         """Add sentiment to step observations"""
         obs = super()._step(tensordict)
 
-        obs["sentiment"] = torch.tensor([self.sentiment_data[self.current_timestamp]])
+        obs["sentiment"] = torch.tensor([self.sentiment_data.loc[self.current_timestamp]])
 
         return obs
 
@@ -182,7 +184,7 @@ class CustomLongOnlyEnv(SequentialTradingEnv):
 import pandas as pd
 
 df = pd.read_csv("prices.csv")
-sentiment = torch.randn(len(df))  # Random sentiment scores
+sentiment = pd.Series(np.random.randn(len(df)), index=pd.to_datetime(df["timestamp"]))
 
 config = SequentialTradingEnvConfig(
     time_frames=["1min", "5min"],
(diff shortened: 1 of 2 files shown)


BY571:
@CharlieHelps please review. This corrects two guide files against the code:

- `docs/guides/custom-environment.md`: `self.observation_spec` (not `_observation_spec`), TensorDict step loops instead of gym-style tuples, `self.current_timestamp` (not `sampler.reset_index` / `current_index`), and `TorchTradeBaseEnv` (not `BaseTorchTradeEnv`, which does not exist).
- `docs/guides/metrics.md`: same spec and step-loop corrections.

Every example was run against the real sampler rather than read. CI green.

BY571:
@CharlieHelps all four applied in `63159ec9`. Every one was verified by executing the block rather than reading it, and all four were real.

**1. The sentiment lookup.** You are right, and it was worse than the type annotation. Running Example 2 as written:

```
IndexError: only integers, slices (`:`), ellipsis, None and long or byte Variables
are valid indices
```

It is a timeseries, so `sentiment_data` is a `pd.Series` now and both lookups use `.loc[self.current_timestamp]`. The usage block below it built the data with `torch.randn(len(df))`, which has the identical problem, so that is now a Series indexed by the frame's timestamps. The block also used `pd` and `np` without importing either.

**2. `_observation_spec` at 227 and 384.** Confirmed, and the private name was only half the defect. Both forms fail:

```
self._observation_spec["x"] = ...   AttributeError: no attribute '_observation_spec'
self.observation_spec["x"]  = ...   RuntimeError: Cannot modify a locked Composite
```

So your parenthetical about whole-spec reassignment is the operative part. The guidance is now clone, extend, reassign, which I verified works. The troubleshooting row says reassign the public property.

These two lines sat eleven lines below this PR's own note that assigning `_observation_spec` raises, so the file contradicted itself.

**3. `obs["reward"]`.** Correct, `obs` is undefined in that block after the step result moved to `td`. It reads `td["next", "reward"]` and I ran all three assertions in that snippet against a real env; they pass.

**4. `metrics.md:71`.** Correct, and `tests/test_examples.py:128` is the convention. Changed to `td = policy(td)` and ran the loop.

Example 2 now builds, resets and steps, with the sentiment value changing between reset and step. On your last point, that the doc checks do not execute these blocks: they do not, and I did not extend them to. The import sweep validates symbols and syntax only. Executing arbitrary guide blocks needs fixtures the suite does not have, and I would rather not add a half-executing harness that gives false confidence. I did add a path guard in a separate PR that catches the phantom-path class mechanically.

BY571:
@CharlieHelps fixed in `68383a42`. Reproduced your failure exactly before fixing it:

```
emitted sentiment dtype: torch.float64
declared spec dtype:     torch.float32
spec.is_in(value):       False
check_env_specs:         The dtypes of the real and fake tensordict don't match for
                         key sentiment. Got fake=torch.float32 and real=torch.float64.
```

This one is mine, and it is a good illustration of why the earlier check was not enough. Moving `sentiment_data` from a `torch.Tensor` to a `pd.Series` fixed the timestamp lookup and silently changed the dtype, because the pandas lookup returns a numpy `float64` while `torch.randn` had been giving `float32`. I verified the example builds, resets and steps, and stopped there. It does all three; the value just is not in its own spec.

Both tensors now take `dtype=self.observation_spec["sentiment"].dtype` rather than a literal `torch.float32`, so the cast follows the declared spec instead of restating it in a third place that can drift. `check_env_specs(env)` passes.

On index alignment, agreed, and thank you for saying so explicitly rather than leaving it ambiguous.

PR #431, opened 2026-09-03
docs: the structure docs named directories and files that do not exist

BY571:
Third of the docs PRs. This one is the structure documentation: directory trees, the class hierarchy, and the paths docs point readers at.

## Twelve phantom references

Every one verified absent by execution, not by reading:

| Doc | Named | Reality |
|---|---|---|
| `envs/README.md` | `offline/longonly/`, `offline/futures/` | neither directory exists |
| `visualization.md` | `offline/base.py` | `core/offline_base.py` |
| `sampler.md` | `offline/sampler.py` | `offline/infrastructure/sampler.py` |
| `offline-rl.md` ×2 | `examples/offline/iql/` | `examples/offline_rl/iql/` |
| 3 package READMEs ×4 | `offline/README.md` | never existed |
| 2 guides ×4 | `docs/examples.md` | `docs/examples/index.md` |
| `ppo_chronos/README.md` | `cd examples/online/ppo_chronos` | `examples/online_rl/` |

The `examples/offline/iql/` one is why the guard checks the git index and not the filesystem: that directory still exists here as an empty leftover, so it reads as fine locally while being absent for anyone who clones.

## The class hierarchy was wrong in a way that matters

`core/README.md` drew the offline envs as four siblings. They are a chain, so a change to `SequentialTradingEnv` reaches all three. It also placed `VectorizedSequentialTradingEnv(+SLTP)` under `TorchTradeOfflineEnv` when it subclasses `EnvBase` directly, as does `PolymarketBetEnv`. Those two families do **not** inherit the shared bases, so a change there must be applied to them by hand. Given CLAUDE.md's "apply to ALL environments" rule, that is the sort of thing the hierarchy doc exists to tell you. The live side also omitted `TorchTradeFuturesLiveEnv`, the base the four futures venues share.

Rebuilt from `__subclasses__` and `__mro__` rather than transcribed.

## Guards

`test_a_documented_repo_path_is_tracked` asserts every repo path a doc names is in the git index, across four forms: backticked, github tree/blob URL, relative link target, and shell-fence path. Each form has its own floor, because a single floor over the total let the whole github-URL form die unnoticed.

The hierarchy tests parse the diagram by indentation and assert each edge is a real **direct** base, that every intermediate class appears, and that the three envs documented as outside the tree still are. Direct bases rather than `issubclass` on purpose: `issubclass(OneStepTradingEnv, TorchTradeOfflineEnv)` is True, so an `issubclass` check passes the wrong diagram unchanged.

Run against `main`'s diagram they produce exactly the five errors this PR fixed by hand, and zero against this branch.

## Three rounds of review

The rounds mostly caught defects the fix itself introduced, which is the honest summary:

- Round 1: I annotated `TorchTradeFuturesLiveEnv` with `futures_live_base.py` as though it were in `core/`; called Polymarket "with a `dry_run` paper path" when it is paper-only and raises; and kept "each live provider directory has the same layout" in the same diff that added polymarket, which has neither `base.py` nor `env_sltp.py`. It also found 8 broken relative links already shipping.
- Round 2: my guard counted **test-defined** subclasses, so a harness in `test_live_env_base.py` made a production leaf look like an intermediate class and failed naming it. My first fix filtered the wrong site. Also: I generalized "they still share utils" to all three outside-the-tree classes without checking polymarket, which imports one utils module and no reward function.
- Round 3: deleted the venue list instead of maintaining it. It was the third copy in one file and the one that drifted, and my previous response had been to hand-update all three.

`4933 passed, 16 skipped`. ruff clean. 0 broken relative links repo-wide.

## Not covered

The ASCII trees. A correct checker is not the cheap one it looks like: the trees are display-rooted at each README's own directory and some entries are placeholders or generated output. The test docstring says so rather than implying coverage it lacks.

### DIFF
--- a/docs/environments/offline-rl.md
+++ b/docs/environments/offline-rl.md
@@ -110,10 +110,10 @@ TorchTrade's offline RL support is compatible with any offline RL algorithm from
 
 ## Next Steps
 
-- **[IQL Example](../examples.md#offline-training)** - Complete offline RL implementation
+- **[IQL Example](../examples/index.md#offline-rl)** - Complete offline RL implementation
 - **[Offline Environments](offline.md)** - Environments for dataset collection
 - **[Online Environments](online.md)** - Live trading for data collection
-- **[Examples](../examples.md)** - Browse all training examples
+- **[Examples](../examples/index.md)** - Browse all training examples
 
 ## References
 
--- a/docs/environments/offline-rl.md
+++ b/docs/environments/offline-rl.md
@@ -19,7 +19,7 @@ This approach is particularly valuable for:
 
 ## Example: IQL (Implicit Q-Learning)
 
-TorchTrade provides an example implementation of offline RL using **Implicit Q-Learning (IQL)** in `examples/offline/iql/`.
+TorchTrade provides an example implementation of offline RL using **Implicit Q-Learning (IQL)** in `examples/offline_rl/iql/`.
 
 ```python
 # Example: Training IQL on pre-collected dataset
@@ -43,7 +43,7 @@ for batch in replay_buffer:
     optimizer.step()
 ```
 
-For a complete implementation, see [examples/offline/iql/](https://github.com/TorchTrade/torchtrade/tree/main/examples/offline/iql).
+For a complete implementation, see [examples/offline_rl/iql/](https://github.com/TorchTrade/torchtrade/tree/main/examples/offline_rl/iql).
 
 ## Dataset Collection
 
--- a/docs/environments/visualization.md
+++ b/docs/environments/visualization.md
@@ -36,7 +36,7 @@ fig.savefig("episode_history.png")
 
 ## Visualization Types
 
-All environments render **3 subplots**:
+All three listed environments render **3 subplots**:
 
 1. **Price History with Actions**: Shows the asset price over time with long/short actions marked as green upward triangles (long/buy) and red downward triangles (short/sell). Position closes are shown as orange (long close) and cyan (short close) markers.
 2. **Portfolio Value vs Buy-and-Hold**: Compares your agent's portfolio value against a simple buy-and-hold baseline strategy.
--- a/docs/environments/visualization.md
+++ b/docs/environments/visualization.md
@@ -1,6 +1,6 @@
 # Visualizing Episode History
 
-All offline environments support **`render_history()`** to visualize episode performance after training or evaluation. The method automatically detects the environment type and renders appropriate plots.
+The three sequential offline environments support **`render_history()`** to visualize episode performance after training or evaluation. The method automatically detects the environment type and renders appropriate plots. The vectorized environments do not have it.
 
 ## Overview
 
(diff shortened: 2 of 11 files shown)


BY571:
@CharlieHelps please review.

This is the structure-docs PR I mentioned on #429. It takes your Markdown relative-link suggestion and generalises it: the guard asserts every repo path a doc names is in the git **index** (not the filesystem), across four forms — backticked, github tree/blob URL, relative link target, and shell-fence path. Each form carries its own floor, because a single floor over the total let the entire github-URL form die unnoticed in an earlier revision.

It found twelve phantom references, including the eight broken relative links you predicted that class would have. Four of them pointed at a `torchtrade/envs/offline/README.md` that has never existed.

I did not extend the import sweep to judge prose, per your note. The hierarchy test is the one addition beyond link checking: it parses the diagram in `core/README.md` and asserts each edge against the real **direct** base. `issubclass` would have passed the wrong diagram unchanged, since `issubclass(OneStepTradingEnv, TorchTradeOfflineEnv)` is True even when the diagram draws them as siblings. Run against main it reproduces exactly the five errors I had found by hand.

Worth your attention specifically: the claim that `VectorizedSequentialTradingEnv(+SLTP)` and `PolymarketBetEnv` sit outside the shared hierarchy, and the sentence about what they still share. I got that sentence wrong once already by generalising from the vectorized pair to polymarket without checking it.

BY571:
@CharlieHelps fixed in `3a37c387`. You were right, and the shape of the mistake is worth naming: I added an accurate exception paragraph and left every sentence above it asserting the opposite. All four:

```
core/README.md:3   "base classes that all TorchTrade environments inherit from"
core/README.md:8   "Root base class for all environments"
core/README.md:20  "Base class for all backtesting environments"
core/README.md:32  "Base class for all live trading environments"
```

Narrowed all four to the shared hierarchy, with one pointer at the top to the exceptions rather than repeating the carve-out in four places.

I then swept for the same claim elsewhere instead of stopping at the file you pointed at, which found three more in `envs/README.md`. One of them is false for a different reason and is the more interesting find:

> "**HistoryTracker**: Episode history used by every environment"

`HistoryTracker()` is constructed in exactly three places (`core/offline_base.py`, `live/shared/futures_live_base.py`, `live/alpaca/base.py`), so the vectorized envs and polymarket do not have one. Worth noting that a source grep is the wrong instrument here: `vectorized_sequential.py` mentions `HistoryTracker` only inside a comment explaining why it deliberately does **not** use one, so grepping for the name suggests the opposite of the truth.

`visualization.md` also said `render_history()` works "across all 3 offline environments", which this PR made worse by surfacing five as offline environments. The vectorized pair does not have the method, for exactly the reason the hierarchy paragraph gives.

Two that look like the same defect and are not, checked before touching them: "All environments support both spot and futures trading via config" in `README.md` and `docs/index.md` both sit under an **Offline Environments** heading, and all five offline configs do carry `leverage`. Correct as written, so left alone.

`4933 passed, 16 skipped`.

BY571:
@CharlieHelps taken in `2fc7dc49`. `visualization.md:39` now reads "All three listed environments render 3 subplots", which is the same broad-reading class as the four claims above it, so it belongs in this PR rather than deferred. Thanks for both passes on this one.

PR #433, opened 2026-09-04
chore: packaging metadata, before any publish

BY571:
Groundwork for pinning a version and publishing. A PyPI version can never be re-uploaded, so metadata mistakes are cheap now and permanent later. No release or upload here.

## Three imports were never declared

`torch`, `numpy` and `pandas` are imported directly by the package and arrived transitively, via `torchrl` and `alpaca-py`. That is not a contract, and this very commit moves dependencies around, which is when it would have bitten.

The floors are measured rather than picked:

| | floor | evidence |
|---|---|---|
| torch | `>=2.1.0` | torchrl 0.13's own floor. Everything used (`tensor`, `where`, `zeros`, `from_numpy`, `bfloat16`) predates it by years. |
| numpy | `>=1.26` | every numpy call in the package exercised on 1.26.4, 2.0.2, 2.3.5. No numpy-2-removed aliases, no numpy-2-only APIs. |
| pandas | `>=2.1` | `to_pandas_freq()` emits lowercase `'1min'`/`'1h'`/`'1D'`. Checked on 2.1.4/2.2.3/3.0.5 with `FutureWarning` promoted to an error: clean. 2.0.3 cannot be tested at all, it fails to import against numpy 2 with a dtype ABI error. |

Full suite at numpy 1.26.4 + pandas 2.1.4: **4927 passed**, zero failures under `tests/envs`.

## `datasets` had no floor, and it breaks

Declared as a bare `"datasets"`, so a clean resolve picked 2.14.4, which calls `pyarrow.PyExtensionType` — removed in pyarrow 25:

```
AttributeError: module 'pyarrow' has no attribute 'PyExtensionType'
```

The dev machine has 5.0.1, which is why it never showed locally. Floored at `>=3.0`.

## Four deps the package never imports

`python-dotenv`, `tqdm`, `hydra-core`, `scikit-learn` are used only by scripts under `examples/`. Moved to an `examples` extra. CI installs it, because `test_examples.py` runs those scripts, and the three places the docs list extras now mention it.

Deliberately **not** done: splitting the exchange SDKs into per-venue extras. `binance`, `okx`, `pybit` and `wandb` are already lazy-imported and could move today, but `alpaca` (6 module-scope imports) and `ccxt` (2) would need their imports made lazy first. That is a code change for marginal gain and does not belong in a metadata PR.

## One version source

`pyproject.toml` and `torchtrade/__init__.py` each held `0.1.0`, and a stale `torchtrade.egg-info/PKG-INFO` from an old editable install still said **`0.0.1`** and shadowed `importlib.metadata` in the repo. Hatchling now reads `__init__.py`; the egg-info is deleted (already gitignored).

`tests/test_version_pin.py` opened with *"The version lives in two files; keep them one number"*, so this PR falsifies its premise and the suite failed. Rewritten to guard the new invariant rather than dropped: `dynamic` is declared, no literal `version` has come back, and the hatch path points at a file that defines `__version__`. All three mutations verified to fail it.

`allow-direct-references` is removed. Nothing uses a direct reference (`trading-nets` appears only in three docstrings), and PyPI rejects them, so leaving it on hides that until upload day.

## A second CI job, because `pytest` structurally cannot catch any of this

The existing job runs with the repo on `sys.path`, so an undeclared dependency resolves from the checkout and the suite stays green. That is exactly why `numpy`, `pandas`, `torch` and the `datasets` floor all went unnoticed.

The new `wheel` job builds the wheel, installs it with only its declared runtime deps, then imports **and steps an env** from a directory that is not the checkout. I extracted the generated step out of the YAML and ran it against the real wheel before committing, because a step that parses is not a step that works.

## Review

I did not run the five-agent three-round cycle. The diff is packaging config plus one rewritten guard, and the verification that matters here is the clean-room install, which I ran: build, install with declared deps only, import from `/tmp`, construct and step an offline env, import the live and replay classes.

`4933 passed, 16 skipped`.

## Next, not in this PR

A release workflow using PyPI Trusted Publishing on tag push, dry-run to TestPyPI first. That needs a trusted publisher configured on the PyPI side, which is your account action.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv pip install --system torch --index-url https://download.pytorch.org/whl/cpu
-          uv pip install --system -e ".[dev,chronos]"
+          uv pip install --system -e ".[dev,chronos,examples]"
 
       - name: Login to HuggingFace
         env:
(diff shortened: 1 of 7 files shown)


BY571:
@CharlieHelps following up on this one, it has been sitting a few hours. No rush if you are backed up, but flagging it since it is the last open PR.

Short version of what it does: the package imports `torch`, `numpy` and `pandas` directly and declared none of them (they arrived transitively via torchrl and alpaca-py), and `datasets` was declared with no floor at all, which lets a clean resolve pick 2.14.4 and die on the removed `pyarrow.PyExtensionType`. Neither shows up locally, because `pytest` runs with the repo on `sys.path` and the dev machine already has new enough versions.

Three things I would value your eye on specifically:

1. **The floors.** `torch>=2.1.0` is inherited from torchrl 0.13's own floor. `numpy>=1.26` and `pandas>=2.1` are measured: I ran the full suite at numpy 1.26.4 + pandas 2.1.4 (4927 passed, zero failures under `tests/envs`) and checked the lowercase `'1min'`/`'1h'`/`'1D'` aliases `to_pandas_freq()` emits against 2.1.4/2.2.3/3.0.5 with `FutureWarning` promoted to an error. pandas 2.0.3 cannot be tested at all, it fails to import against numpy 2 with a dtype ABI error, so 2.1 is the floor I can defend rather than the one I would guess.

2. **Moving four deps to an `examples` extra.** `python-dotenv`, `tqdm`, `hydra-core` and `scikit-learn` are imported nowhere under `torchtrade/`. It is the one judgment call in the diff; everything else is fixing something broken.

3. **The new `wheel` CI job.** It builds the wheel, installs it with only its declared runtime deps, then imports and steps an env from a directory that is not the checkout. That is the check the existing suite structurally cannot perform.

One follow-up since I opened it, in `bf6f3c35`: `docs/examples/index.md` lists every runnable example and its `python examples/...` command and had no install step whatsoever. Fine while those four were runtime deps; broken the moment they became an extra. Verified the gap was real rather than theoretical: 14 example scripts import `tqdm`, 9 import `hydra`, 9 import `sklearn`, 5 import `dotenv`.

All three checks are green including the new job.

BY571:
@CharlieHelps both fixed in `5a15ed04`.

**The version comment.** You are right, and it is the same defect the docs PRs spent three rounds on. `torchtrade/__init__.py` still said the version was "kept in step with pyproject.toml" and that "two files holding one number is exactly the shape that drifts" — while this branch is what removed the second file. I changed the mechanism and left the prose describing the old one. It now says this file is the source, hatchling reads it, and the test guards that.

**`uv sync` does not activate.** Correct, and worse than the page being merely inconsistent: I added an install step for dependencies I had just moved, then left the commands below it resolving whichever interpreter happens to be on PATH, which is exactly the environment that would not have them.

The page was already split, 5 commands using `uv run` and 7 bare. I took your first option and made all 12 `uv run python`, since that works whether or not the reader activated, rather than adding an activation line that is easy to skip. The prose now states which interpreter the commands use and gives the activation alternative for anyone who prefers a bare `python`.

`4933 passed, 16 skipped`.

On what happens after this merges, in case it affects how you read the diff: the plan is a release workflow using Trusted Publishing on tag push with a TestPyPI dry-run first, then a `v0.1.0` tag. Nothing is published from this PR, and the trusted-publisher configuration is an account-side step for the maintainer.

PR #434, opened 2026-09-04
chore: release workflow (publishes nothing yet)

BY571:
Tag-driven release for when you decide to publish. **Nothing is uploaded by this PR.**

## The workflow

`git tag v0.1.0 && git push origin v0.1.0` runs `build → verify → publish`.

- **`publish` requires a pushed tag**, not merely a tag-shaped ref: `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')`. A ref-only check is not enough, because `workflow_dispatch` can target a tag.
- **The tagged commit must be an ancestor of `origin/main`.** `tests.yml` fires on pushes to `main` and PRs into it, so a tag push matches neither trigger; without this, a tag on any commit on any branch publishes code the suite never ran.
- **The tag must equal `__version__`.**
- **Trusted Publishing** (`id-token: write` on `publish` alone), and the publish action is SHA-pinned because `@release/v1` is a mutable *branch* on the only job holding the OIDC token.
- **`verify` installs both artifacts** into clean venvs and drives an env from outside the checkout, and runs `twine check` on both.

## What building an sdist for the first time exposed

**It shipped internal scaffolding.** hatchling ships everything not gitignored, and three untracked working directories are not gitignored, so the tarball carried `.superpowers/` (36 entries), `local_verify/`, `benchmarks/` and 1 MB of doc images. Now an allowlist: `2467 KB → 295 KB`. `tests/` is excluded too, because `test_documented_imports` shells out to `git ls-files` and `test_examples` resolves `REPO_ROOT/examples`, so the suite errors on collection from an unpacked sdist.

**`.env.example` contradicted the code**, in the file a user copies verbatim:

| it offered | the code reads |
|---|---|
| `API_KEY`, `SECRET_KEY` | `ALPACA_API_KEY`, `ALPACA_SECRET_KEY` |
| `BINANCE_SECRET` | `BINANCE_SECRET_KEY` |
| `BITGET_*` | `bitget/order_executor.py:110-130` warns those are deprecated |
| nothing for bybit/okx | the doc examples read `BYBIT_*` and `OKX_*` |

**The README is the PyPI page**, and it carried a relative `` plus two relative links while `docs/` is not in the sdist. PyPI does not resolve those, `twine check` cannot see it (it validates that the description parses, not that its assets exist), and a rendered page is not fixed by yanking. Absolute now.

## Review round

Five agents. Findings acted on, beyond the above: the publish gate and ancestry guard; `skip-existing` so a partial upload is retryable; a constant `concurrency` group, because a per-ref group puts two different tags in different groups and does not serialise them; a bounded build backend; top-level `permissions: contents: read`.

The two clean-install probes had already drifted in the commit that created them, the release copy having dropped the binance import, so the check guarding the release was the weaker one. They are now one shared `.github/scripts/smoke_installed.py`.

Two findings I checked and did not act on: `POLYGON_PRIVATE_KEY` is not read by the polymarket executor (both hits are comments explaining why it *refuses* a private key), and bitget's deprecated spellings stay out of `.env.example` because the executor reads them only to warn.

## Tests

`tests/test_packaging.py`: `.env.example` offers no credential the tree never reads, every credential an example reads is in `.env.example`, and the sdist config is an allowlist. Each of the four historical bugs was replayed individually and fails them.

`4936 passed, 16 skipped`. Both artifacts pass `twine check` and install into clean venvs.

## Correcting an earlier claim in this PR

An earlier version of this description said `.env.example` ships in the sdist. That was true of the old sdist and false after the include list this PR adds. It does not ship.

## After this, and outside the repo

1. Register a **pending** trusted publisher on PyPI: owner `TorchTrade`, repo `torchtrade`, workflow `release.yml`, environment `pypi`. The name is unclaimed, so this binds the OIDC identity and takes the name. Without it the first tag fails at upload.
2. Create the `pypi` environment with a required reviewer. It does not exist; GitHub auto-creates it empty, so today it is a label rather than a gate.
3. Protect `main` and `refs/tags/v*`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,15 +1,16 @@
-# Alpaca (US stocks & crypto) - https://alpaca.markets
-API_KEY=your_alpaca_api_key
-SECRET_KEY=your_alpaca_secret_key
+# Alpaca (crypto spot) - https://alpaca.markets
+ALPACA_API_KEY=your_alpaca_api_key
+ALPACA_SECRET_KEY=your_alpaca_secret_key
 
 # Binance (crypto futures) - https://www.binance.com
 BINANCE_API_KEY=your_binance_api_key
-BINANCE_SECRET=your_binance_secret_key
+BINANCE_SECRET_KEY=your_binance_secret_key
 
 # Bitget (crypto futures) - https://www.bitget.com
-BITGET_API_KEY=your_bitget_api_key
-BITGET_SECRET=your_bitget_secret_key
-BITGET_PASSPHRASE=your_bitget_passphrase
+# The BITGET_* spellings still work but the executor warns they are deprecated.
+BITGETACCESSAPIKEY=your_bitget_api_key
+BITGETSECRETKEY=your_bitget_secret_key
+BITGETPASSPHRASE=your_bitget_passphrase
 
 # Weights & Biases (optional, for training logs) - https://wandb.ai
 WANDB_API_KEY=your_wandb_api_key
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,107 @@
+name: Release
+
+# Tag-driven. `git tag v0.1.0 && git push origin v0.1.0` builds, verifies and
+# publishes that tag. workflow_dispatch runs everything except the PyPI upload, so the
+# build and the clean-install check can be exercised without spending a version number.
+#
+# A PyPI version can never be re-uploaded or replaced, only yanked, so `verify` runs
+# the same clean-room install the test workflow does before `publish` is allowed to
+# start.
+on:
+  push:
+    tags: ["v*"]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
+
+      - name: The tag must match the version the package declares
+        if: startsWith(github.ref, 'refs/tags/v')
+        run: |
+          tag="${GITHUB_REF_NAME#v}"
+          declared=$(grep -oP '^__version__ = "\K[^"]+' torchtrade/__init__.py)
+          echo "tag=$tag declared=$declared"
+          if [ "$tag" != "$declared" ]; then
+            echo "::error::tag $GITHUB_REF_NAME does not match __version__ $declared"
+            exit 1
+          fi
+
+      - name: Build sdist and wheel
+        run: uv build --out-dir dist .
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist
+          path: dist/
+
+  verify:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
+
+      - name: Install the wheel alone and drive an env from outside any checkout
+        run: |
+          uv venv --python 3.11 /tmp/relenv
+          uv pip install --python /tmp/relenv/bin/python dist/*.whl
+          mkdir -p /tmp/relprobe && cd /tmp/relprobe
+          /tmp/relenv/bin/python - <<'PY'
+          import numpy as np, pandas as pd, torchtrade
+          from importlib.metadata import version
+
+          assert "site-packages" in torchtrade.__file__, torchtrade.__file__
+          print("torchtrade", version("torchtrade"))
+
+          from torchtrade.envs.offline.sequential import (
+              SequentialTradingEnv, SequentialTradingEnvConfig,
+          )
+          n = 3000
+          close = 100 + np.cumsum(np.random.randn(n) * 0.1)
+          df = pd.DataFrame({
+              "timestamp": pd.date_range("2024-01-01", periods=n, freq="1min", tz="UTC"),
+              "open": close, "high": close + 0.5, "low": close - 0.5,
+              "close": close, "volume": np.random.rand(n) * 10,
+          })
+          env = SequentialTradingEnv(df=df, config=SequentialTradingEnvConfig(
+              time_frames=["1Min"], window_sizes=[10], execute_on="1Min", symbol="BTC/USD"))
+          td = env.reset()
+          td["action"] = env.action_spec.rand()
+          env.step(td)
+
+          from torchtrade.envs.live.alpaca import AlpacaTorchTradingEnv       # noqa: F401
+          from torchtrade.envs.replay import ReplayObserver                   # noqa: F401
+          print("offline env stepped; live and replay imports resolved")
+          PY
+
+  publish:
+    # Tags only. workflow_dispatch stops after verify, so a rehearsal cannot spend a
+    # version number.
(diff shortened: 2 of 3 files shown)


BY571:
@CharlieHelps re-pinging this one, it has been sitting since the 4th. It is the last open PR and it gates the release, so I would rather have your eyes on it than merge on green checks alone.

It adds a tag-driven release workflow and publishes nothing. Three things I would value scrutiny on specifically:

1. **The publish gate.** `publish` has `if: startsWith(github.ref, 'refs/tags/v')` and `needs: verify`. The intent is that `workflow_dispatch` rehearses build and verify but cannot reach an upload, so a rehearsal cannot spend a version number. Please check that gate actually holds, including for a tag pushed to a branch that is not `main`.

2. **The sdist include list** in `pyproject.toml`. Building an sdist for the first time showed hatchling was shipping `.superpowers/`, `local_verify/` and `benchmarks/`, which are untracked working directories that are not gitignored, plus 1MB of doc images: 2467 KB down to 773 KB. I kept `tests/` in the sdist deliberately. If you think tests do not belong there, say so.

3. **`.env.example`.** It contradicted the code in four ways and it ships in the sdist, so it is the file a user copies verbatim. The Alpaca row (`API_KEY` rather than `ALPACA_API_KEY`) is the same defect you found three times in #429; this was the fourth copy and the only non-prose one. The bitget rows named variables that `bitget/order_executor.py:110-130` actively warns are deprecated. Worth confirming I picked the right names rather than a fifth wrong set.

One gap I want to flag rather than have you find: I described a TestPyPI dry-run in earlier discussion and did not build one. `workflow_dispatch` rehearses build and verify but never exercises an upload, so Trusted Publishing auth is untested until the first real tag. My reasoning is that an auth failure costs a retry and no version number, while the irreversible failure mode is a successful upload of something wrong, which `verify` and the artifact inspection do cover. Tell me if you disagree; I would rather add it now than after the first tag.

BY571:
Update: I ran the five-agent review round I had skipped, and it was the right call. Pushed as `6d60d01a`. Two of the findings would have published something that cannot be undone.

**The rehearsal could publish.** My gate was `if: startsWith(github.ref, 'refs/tags/v')`. GitHub's `workflow_dispatch` ref picker offers tags, and `GITHUB_REF` is documented as "Branch or tag that received dispatch", so dispatching on `v0.1.0` passes a ref-only check and uploads. The comment directly above it told the reader that was impossible. Now `github.event_name == 'push' && startsWith(...)`.

**Nothing tied a tag to tested code.** `tests.yml` triggers on pushes to `main` and PRs into it; a tag push matches neither. So a tag on any commit on any branch published code the suite had never run, and unlike an auth failure that *succeeds*. `build` now refuses a tag whose commit is not an ancestor of `origin/main`.

**The README is the PyPI page.** It carried a relative `` and two relative links, while `docs/` is deliberately not in the sdist. PyPI does not resolve those, and `twine check` cannot catch it by construction: it validates that the description parses, not that its assets exist. A rendered page is not fixed by yanking.

Also: SHA-pinned `pypa/gh-action-pypi-publish` (`@release/v1` is a mutable *branch*, on the only job holding the OIDC token; `git ls-remote` confirms `dc37677` is both that branch and tag v1.14.2); the sdist is now actually installed in `verify` rather than published unexercised; `tests/` dropped from the sdist because it errors on collection from there anyway; `skip-existing` so a partial upload is retryable; a constant `concurrency` group, because a per-ref group does not serialise two different tags, which is what my comment claimed it did.

The probe scripts in the two workflows had **already drifted** in the commit that created them: the release copy silently dropped the binance import, so the check guarding the release was the weaker of the two, and my "same clean-room install the test workflow does" comment was false. One shared `.github/scripts/smoke_installed.py` now.

**Correcting something I told you earlier in this PR:** I said `.env.example` ships in the sdist. That was true of the old sdist and false after my own include list. It does not ship.

And my previous commit broke two examples. `.env.example` dropped `API_KEY`, `SECRET_KEY` and `BINANCE_SECRET`, which `examples/online_rl/ppo/live.py:100` and `dqn/live.py:108` read. Worth being precise about why I missed it: I scanned with `os\.getenv\("[A-Z_]+"\)`, which requires the closing paren and so skipped every call site with a default. `os.getenv("API_KEY", "")` was invisible to my check, and I reported the file consistent.

`tests/test_packaging.py` closes that class properly, and my first version of it was itself dead for its own motivating case: it filtered on a venue prefix when the names that shipped wrong were bare `API_KEY` and `SECRET_KEY`. All four historical bugs now fail the rebuilt guards, each replayed individually.

`4936 passed, 16 skipped`. Both artifacts pass `twine check`, and each installs into a clean venv and steps an env from outside the checkout.

Three things remain outside the repo and are yours, not mine, ordered by what blocks what:

1. **Register a *pending* trusted publisher on PyPI** (owner `TorchTrade`, repo `torchtrade`, workflow `release.yml`, environment `pypi`). The name is currently unclaimed, so this both binds the OIDC identity and takes the name. Without it the first tag fails at upload.
2. **Create the `pypi` environment with a required reviewer.** It does not exist; GitHub auto-creates it empty on first use, so as written it is a label rather than a gate. A reviewer is what makes a mistaken tag recoverable.
3. **Protect `main` and `refs/tags/v*`.** `main` is currently unprotected, which is what makes an arbitrary tag reachable in the first place.

BY571:
@CharlieHelps your review is against `3c7d85d`, and both blocking items were fixed in `6d60d01a`, which landed about 20 minutes before you posted. Bad timing rather than disagreement, and you and the review round independently found the same two things, which is a decent signal they were the right ones to find.

Verified at current HEAD:

**1. The publish gate.**

```
release.yml:113  if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
release.yml:38   if ! git merge-base --is-ancestor "$GITHUB_SHA" origin/main; then
```

Both halves, exactly as you asked: the event name, and provenance for the tagged commit. On the second I went with an ancestry check rather than running the suite in the release workflow, since the invariant that matters is "this commit was on main, where tests.yml ran it" rather than "tests pass right now". Say if you would rather see the suite run at tag time.

**2. The stale consumers.**

```
examples/online_rl/ppo/live.py:100-101   ALPACA_API_KEY / ALPACA_SECRET_KEY
examples/online_rl/dqn/live.py:108       BINANCE_SECRET_KEY
bitget/README.md                         3 blocks on BITGETACCESSAPIKEY, 0 on BITGET_API_KEY
```

I chose to update the consumers rather than preserve aliases: this is `0.1.0` with no prior release, so there is nothing to stay compatible with, and `ppo/live.py`'s own docstring already documented the names its code was not using.

**On `.env.example` not shipping: you are right, and thank you for catching it in the description rather than only in the diff.** I had already corrected it in a comment, but the PR body still carried the claim, which is the copy anyone reads first. The body is rewritten now. It does not ship, and that became true because of this PR's own include list.

One thing your review lets me correct: you noted the sdist contains 127 test files and that keeping them is reasonable. That was accurate at `3c7d85d`. They are out at HEAD, on the grounds that `test_documented_imports` shells out to `git ls-files` and `test_examples` resolves `REPO_ROOT/examples`, so the shipped suite errors on collection from an unpacked sdist. If you think tests should ship anyway, the counter-argument is real and I will take it.

Also agreed on TestPyPI, and for the same reason you give: it needs its own publisher config and would not exercise the production one. I have written the production setup up as an explicit pre-release checklist in the PR body instead.

`4936 passed, 16 skipped`. All three checks green on `6d60d01a`.

